PR #5, opened 2022-01-18
feat: set up edx-rbac

long74100:
https://openedx.atlassian.net/browse/ENT-5320

I put all the rbac related code in core since it seemed appropriate, open to moving some to the requests app once created.


christopappas (approved):
LGTM. normally i'd ask for tests so we don't have a bunch of uncovered code in rules.py, but i think once we add endpoints that use all this stuff we can test access then, so, 👍

iloveagent57 (approved):

PR #8, opened 2022-01-20
feat: ENT-5308 | Add request and access app, and configuration model

christopappas:


enterprise_access/apps/subsidy_requests/admin.py
```diff
@@ -0,0 +1,75 @@
...
+        Override save_model method to keep our change records up to date.
+        """
+        current_user = get_user_from_request_session(request)
+        jwt_data = get_data_from_jwt_payload(request, ['user_id'])
+        # Make sure we update the user object's lms_user_id if it's not set
+        # or if it has changed to keep our DB up to date, because we have
+        # no way to predict if the user has hit a rest endpoint and had
+        # their info prepopulated already.
+        lms_user_id = jwt_data['user_id']
+        if not current_user.lms_user_id or current_user.lms_user_id != lms_user_id:
```
iloveagent57, line 63: When would `current_user.lms_user_id != lms_user_id` be true? (Other than in cases where current_user.lms_user_id is null)

iloveagent57 (commented):
Looking good!  I'm going to check out this branch and manually test some stuff, one question for now.

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -160,3 +166,58 @@ def approve(self, reviewer_lms_user_id):
...
+        blank=True,
+        null=True,
+        choices=SubsidyTypeChoices.CHOICES,
+        help_text=("Which type of subsidy is used to grant access."),
+    )
+
+    pending_request_reminder_frequency = models.CharField(
+        max_length=32,
+        blank=False,
+        null=False,
```
iloveagent57, line 200: Can we add a default in here?  My vote would be for weekly?

christopappas: actually i wonder, give this feature isn't technically required in the MVP, should default be never? (until we're ready to release this feature)

enterprise_access/apps/subsidy_requests/admin.py
christopappas: if the user id changed on the lms side. it _shouldn't_, but was trying to code defensively to make sure we keep what is effectively the user's pk up to date

christopappas: Another weird edge case technically would be if someone changed a User's lms_user_id in the enterprise-access DB via the django admin. this would keep this up to date

iloveagent57: Ah, ok, so this only applies in the case where:
1. A user has an active session.
2. They're hitting the Django Admin site (which only does session auth, as you noted in another comment).
3. There's an existing `core.User` record based on the auth_user_id in the session cookie.
4. The `lms_user_id` between that `core.User` record and what's in the JWT Auth cookie differ.

Also, your second point is true, but I think you made `lms_user_id` readonly in the Django Admin form, right?

iloveagent57 (approved):

PR #9, opened 2022-01-21
feat: add subsidy models

long74100:
https://openedx.atlassian.net/browse/ENT-5305

- Create subsidy request models
- Added some validations since we have nullable fields
- Hook models up to django admin


enterprise_access/apps/subsidy_request/models.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    uuid = models.UUIDField(
+        primary_key=True,
+        default=uuid4,
+        editable=False,
+        unique=True,
+    )
+
+    lms_user_id = models.IntegerField()
```
christopappas, line 25: we're gonna store this on the User model i think too. i think it's fine to store it here, but for the sake of saying it outloud: would we want to have a foreign key to the User here instead?

although, the User object would not be available if a particular user never authenticated against enterprise-access though (via any REST api), so may not be appropriate 🤔 -- lemme know what you think

christopappas (approved):
just the one thought that i wanted to share, but i think this is good

enterprise_access/apps/subsidy_request/admin.py
```diff
@@ -0,0 +1,88 @@
...
+    """ Filter requests by approval status. """
+
+    title = _('approval status')
+
+    parameter_name = 'approval_status'
+
+    def lookups(self, request, model_admin):
+        return (
+            ('approved', _('Approved')),
+            ('pending', _('Pending')),
```
macdiesel, line 19: Chris, I want to [draw your attention here](https://openedx.atlassian.net/wiki/spaces/SOL/pages/3266543926?focusedCommentId=3313303563#comment-3313303563), we will need an additional state for approved pending assignment (The state when the admin as approved but the code/sub has not yet been assigned).

macdiesel: My other question on this would be by injecting this new state, should we add a state field to the request object to track said states?

long74100: Currently we don't have a state field and the state is computed from the approved_at/denied_at field. We could add a state but I think we could get away without it.

enterprise_access/apps/subsidy_request/models.py
long74100: Yeah I don't think we should link it to a user object for the reason you mentioned.

enterprise_access/apps/subsidy_request/admin.py
long74100: hmm i might want to change this a bit

long74100: going to consolidate approved_at/denied_at to reviewed_at and add a state field

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -0,0 +1,162 @@
...
+
+    course_id = models.CharField(
+        null=True,
+        blank=True,
+        max_length=128
+    )
+
+    customer_uuid = models.UUIDField()
```
iloveagent57, line 35: Nit: we call this `enterprise_customer_uuid` in license-manager and enterprise-customer, should we use the same name here?

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -0,0 +1,162 @@
...
+
+    reviewed_at = models.DateTimeField(
+        null=True,
+        blank=True
+    )
+
+    reviewer_id = models.IntegerField(
+        null=True,
+        blank=True
+    )
```
iloveagent57, line 53: Is this the lms user id of the reviewer, or a FK to `core.User`?

iloveagent57 (commented):
Looks great! two questions

enterprise_access/apps/subsidy_request/models.py
long74100: yep we should do that!

enterprise_access/apps/subsidy_request/models.py
long74100: this is the lms user id of the reviewer, I should maybe change lms_user_id -> requester_lms_user_id and this to reviewer_lms_user_id?

iloveagent57: yeah, or take the position of - everything related to the reviewer has `reviewer` as a prefix, otherwise the prefix is the empty string.  ex.
`lms_user_id` - The requesting learner's LMS user id
`reviewer_lms_user_id` - The reviewing admin's LMS user id

iloveagent57 (approved):

PR #12, opened 2022-01-28
feat: add subsidy request endpoints

long74100:
- set up get/post endpoints for license and coupon code requests
- set up edx-api-doc-tools

https://openedx.atlassian.net/browse/ENT-5306

### DIFF
--- a/enterprise_access/apps/api/exceptions.py
+++ b/enterprise_access/apps/api/exceptions.py
@@ -1,0 +1,17 @@
+"""
+Custom Exception classes for Enterprise Access API v1.
+"""
+
+class SubsidyRequestError(Exception):
+    """
+    A general exception dealing with subsidy requests.
+    """
+    def __init__(self, message, http_status_code=None):
+        super().__init__(message)
+        self.message = message
+        self.http_status_code = http_status_code
+
+class SubsidyRequestCreationError(SubsidyRequestError):
+    """
+    An exception indicating that a subsidy request cannot be created.
+    """
--- a/enterprise_access/apps/api/filters.py
+++ b/enterprise_access/apps/api/filters.py
@@ -1,0 +1,39 @@
+"""
+Filter backends for Enterprise Access API.
+"""
+
+from django.db.models import Q
+from edx_rbac import utils
+from rest_framework import filters
+
+from enterprise_access.apps.core import constants
+
+
+class SubsidyRequestFilterBackend(filters.BaseFilterBackend):
+    """
+    Filter that returns subsidy requests that were either created for the user or created under
+    enterprises that the user is an admin of.
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        """
+        Filter queryset for non staff/super users.
+        """
+
+        if request.user.is_staff or request.user.is_superuser:
+            return queryset
+
+        decoded_jwt = utils.get_decoded_jwt(request)
+        lms_user_id_from_jwt = decoded_jwt.get('user_id')
+        accessible_enterprises_as_admin = utils.contexts_accessible_from_request(
+            request, [constants.REQUESTS_ADMIN_ROLE]
+        )
+
+        # openedx operators will have access to all enterprises
+        if constants.ALL_ACCESS_CONTEXT in accessible_enterprises_as_admin:
+            return queryset
+
+        return queryset.filter(
+            Q(lms_user_id=lms_user_id_from_jwt) |
+            Q(enterprise_customer_uuid__in=accessible_enterprises_as_admin)
+        )
--- a/enterprise_access/apps/api/utils.py
+++ b/enterprise_access/apps/api/utils.py
@@ -1,0 +1,38 @@
+"""
+Utility functions for Enterprise Access API.
+"""
+
+from uuid import UUID
+
+from rest_framework.exceptions import ParseError
+
+
+def get_enterprise_uuid_from_query_params(request):
+    """
+    Extracts enterprise_customer_uuid from query params.
+    """
+
+    enterprise_customer_uuid = request.query_params.get('enterprise_customer_uuid')
+
+    if not enterprise_customer_uuid:
+        return None
+
+    try:
+        return UUID(enterprise_customer_uuid)
+    except ValueError as ex:
+        raise ParseError('{} is not a valid uuid.'.format(enterprise_customer_uuid)) from ex
+
+def get_enterprise_uuid_from_request_data(request):
+    """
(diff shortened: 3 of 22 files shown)


enterprise_access/apps/api/exceptions.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    def __init__(self, message, http_status_code=None):
+        super().__init__(message)
+        self.message = message
+        self.http_status_code = http_status_code
+
+class SubsidyRequestCreationError(SubsidyRequestError):
+    """
+    An exception indicating that a subsidy request cannot be created.
+    """
```
christopappas, line 17: Does this pass the linter? Should there be something different about these two classes? If they are the same, do we need both?

enterprise_access/apps/api/filters.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+from django.db.models import Q
+from edx_rbac import utils
+from rest_framework import filters
+
+from enterprise_access.apps.core import constants
+
+
+class IsSubsidyRequesterOrEnterpriseAdmin(filters.BaseFilterBackend):
```
christopappas, line 12: This name took me a bit to understand. Could this be named something like `FilterOnRole`, `RoleBasedFilter`, `FilterSubsidyRequestOnUserRole` (maybe too verbose)

enterprise_access/apps/api/filters.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Filter that returns subsidy requests that were either created for the user or created under
+    enterprises that the user is an admin of.
+    """
+
+    def get_accessible_enterprises_by_role(self, request):
+        return {
+            'admin': utils.contexts_accessible_from_request(request, [constants.REQUESTS_ADMIN_ROLE]),
+            'learner': utils.contexts_accessible_from_request(request, [constants.REQUESTS_LEARNER_ROLE]),
```
christopappas, line 21: This dict has a `learner` key but is not used at all later in the function. Where would this be used?

enterprise_access/apps/api/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+    enterprise_customer_uuid = request.query_params.get('enterprise_customer_uuid')
+
+    if not enterprise_customer_uuid:
+        return None
+
+    try:
+        return UUID(enterprise_customer_uuid)
+    except ValueError as ex:
+        raise ParseError('{} is not a valid uuid.'.format(enterprise_customer_uuid)) from ex
```
christopappas, line 23: TIL `raise from`

enterprise_access/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,369 @@
...
+        Test that a 422 response is returned when creating a request
+        if no customer configuration is set up.
+        """
+        payload = {
+            'enterprise_customer_uuid': self.enterprise_customer_uuid_1,
+            'course_id': 'edx-demo'
+        }
+        response = self.client.post(LICENSE_REQUESTS_LIST_ENDPOINT, payload)
+        assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
+        assert response.data == 'Customer configuration for enterprise: ' + \
```
christopappas, line 140: i personally don't love `\` in code in general, but especially for string concatenation. I usually go for ()

like,
```
expected_message = (
  'Customer configuration for enterprise: '
  f'{self.enterprise_customer_uuid_1} does not exist.'
)
assert assert response.data == expected_message
```

just me though 🤷

enterprise_access/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,369 @@
...
+        )
+        payload = {
+            'enterprise_customer_uuid': self.enterprise_customer_uuid_1,
+            'course_id': 'edx-demo'
+        }
+        response = self.client.post(LICENSE_REQUESTS_LIST_ENDPOINT, payload)
+        assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
+        assert response.data == f'Subsidy requests for enterprise: {self.enterprise_customer_uuid_1} are disabled.'
+
+    def test_create_subsidy_type_not_set_up(self):
```
christopappas, line 161: good test 🏆

enterprise_access/apps/api/v1/views.py
```diff
@@ -1 +1,163 @@
...
+
+        enterprise_customer_uuid = self.request.data.get('enterprise_customer_uuid')
+        course_id = self.request.data.get('course_id')
+
+        pending_request = CouponCodeRequest.objects.filter(
+            lms_user_id=self.lms_user_id,
+            enterprise_customer_uuid=enterprise_customer_uuid,
+            state__in=[SubsidyRequestStates.PENDING_REVIEW, SubsidyRequestStates.APPROVED_PENDING],
+            course_id=course_id
+
```
christopappas, line 156: nit: space

enterprise_access/apps/api/v1/views.py
```diff
@@ -1 +1,163 @@
...
+        ).first()
```
christopappas, line 157: whenever I see `filter().first()` i get worried, usually because we never know what that thing is that we're getting back because #ordering. if you just want to check something is returned, i would use `.exists()`

christopappas (commented):
looking good! some quick thoughts and questions

enterprise_access/apps/api/exceptions.py
long74100: It does. We don't need both, if we want to distinguish between different types of subsidy request errors this would be helpful.

enterprise_access/apps/api/filters.py
long74100: I might just name this SubsidyRequestFilterBackend and let the comments do the work here. Any semantic name might be too verbose here.

enterprise_access/apps/api/filters.py
long74100: It's not, I can get rid of it.

enterprise_access/apps/api/v1/tests/test_views.py
long74100: no preference here either :p will update

enterprise_access/apps/api/filters.py
christopappas: cool, `SubsidyRequestFilterBackend` sounds great

christopappas (approved):
Looks good!

PR #13, opened 2022-01-31
feat: add viewset for customer configurations

long74100:
https://openedx.atlassian.net/browse/ENT-5321


enterprise_access/apps/api/v1/tests/test_views.py
```diff
@@ -421,3 +423,234 @@ def test_create_happy_path(self):
...
+        customer_config.subsidy_requests_enabled = False
+        customer_config.subsidy_type = previous_subsidy_type
+        customer_config.save()
+
+        payload = {
+            'subsidy_requests_enabled': True,
+            'subsidy_type': new_subsidy_type
+        }
+        response = self.client.patch(
+            CUSTOMER_CONFIGURATIONS_LIST_ENDPOINT + f'{self.enterprise_customer_uuid_1}/',
```
adamstankiewicz, line 647: nit: Could `CUSTOMER_CONFIGURATIONS_LIST_ENDPOINT` go inside of the `f` string? :f:

enterprise_access/settings/devstack.py
```diff
@@ -59,3 +59,9 @@
 )
 # END CELERY
 
+# CORS CONFIG
+CORS_ORIGIN_WHITELIST = [
```
adamstankiewicz, line 63: 🙌

enterprise_access/apps/api/serializers.py
```diff
@@ -75,3 +80,21 @@ class Meta:
...
+    class Meta:
+        model = SubsidyRequestCustomerConfiguration
+        fields = [
+            'enterprise_customer_uuid',
+            'subsidy_requests_enabled',
+            'subsidy_type'
+        ]
+
+    def update(self, instance, validated_data):
+        validated_data.pop('enterprise_customer_uuid', None)
```
adamstankiewicz, line 99: [clarification] We remove `enterprise_customer_uuid` from `validated_data` here because we don't want the API user to be able to modify the `enterprise_customer_uuid`, right? I'm wondering if we could utilize `read_only_fields` here ([docs](https://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields))?

adamstankiewicz (commented):
This LGTM. I added one clarifying question and some nits, most of which are optional formatting type of nits.

enterprise_access/apps/api/v1/tests/test_views.py
```diff
@@ -421,3 +422,190 @@ def test_create_happy_path(self):
...
+        (False, True),
+        (True, True)
+    )
+    @ddt.unpack
+    def test_list_as_staff_or_superuser(self, is_staff, is_superuser):
+        if is_staff:
+            self.user.is_staff = True
+
+        if is_superuser:
+            self.user.is_superuser = True
```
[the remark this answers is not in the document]
long74100: lol big brain fart sorry

enterprise_access/apps/api/serializers.py
long74100: Yeah that's correct. If we want to put it in read_only_fields, we would need to have a separate serializer for write since enterprise_customer_uuid will be discarded.

adamstankiewicz: > Yeah that's correct. If we want to put it in read_only_fields, we would need to have a separate serializer for write since enterprise_customer_uuid will be discarded.

True. Personally, no strong feelings on your approach vs. separate serializer though we do have multiple serializers pattern elsewhere, e.g. `EnterpriseCourseEnrollmentWriteSerializer`, `EnterpriseCustomerUserWriteSerializer`, `EnterpriseCustomerInviteKeyWriteSerializer`.

If kept as is, maybe add a comment to explain why we're popping `enterprise_customer_uuid` as it might not be intuitive at first glance (at least it wasn't for me 😅 ).

long74100: No strong opinion here either, didn't want to write a separate serializer and add logic to the viewset since the difference is so small 😅 . I can add a comment to clarify.

adamstankiewicz: Awesome, works for me!

adamstankiewicz (approved):

PR #14, opened 2022-01-31
Add approve and decline api endpoints

christopappas:
- Adds approve and deny endpoints for licenses and coupons
- Hitting approval endpoint will hit license-manager/ecommerce to verify there are available licenses/coupons available

### DIFF
--- a/enterprise_access/apps/api/constants.py
+++ b/enterprise_access/apps/api/constants.py
@@ -1,0 +1,3 @@
+""" Constants for the api app. """
+
+LICENSE_UNASSIGNED_STATUS = 'unassigned'
--- a/enterprise_access/apps/api/v1/views.py
+++ b/enterprise_access/apps/api/v1/views.py
@@ -5,23 +5,31 @@ Views for Enterprise Access API v1.
 import logging
 
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import transaction
 from django.utils.functional import cached_property
 from django_filters.rest_framework import DjangoFilterBackend
 from edx_rbac import utils
 from edx_rbac.decorators import permission_required
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from requests.exceptions import ConnectionError as RequestConnectionError
+from requests.exceptions import HTTPError, Timeout
 from rest_framework import filters, permissions, status, viewsets
+from rest_framework.decorators import action
+from rest_framework.exceptions import ParseError
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 
 from enterprise_access.apps.api import serializers
-from enterprise_access.apps.api.exceptions import SubsidyRequestCreationError
+from enterprise_access.apps.api.constants import LICENSE_UNASSIGNED_STATUS
+from enterprise_access.apps.api.exceptions import SubsidyRequestCreationError, SubsidyRequestError
 from enterprise_access.apps.api.filters import (
     SubsidyRequestCustomerConfigurationFilterBackend,
     SubsidyRequestFilterBackend
 )
 from enterprise_access.apps.api.tasks import delete_enterprise_subsidy_requests_task
-from enterprise_access.apps.api.utils import get_enterprise_uuid_from_request_data
+from enterprise_access.apps.api.utils import get_enterprise_uuid_from_request_data, validate_uuid
+from enterprise_access.apps.api_client.ecommerce_client import EcommerceApiClient
+from enterprise_access.apps.api_client.license_manager_client import LicenseManagerApiClient
 from enterprise_access.apps.core import constants
 from enterprise_access.apps.subsidy_request.constants import SubsidyRequestStates, SubsidyTypeChoices
 from enterprise_access.apps.subsidy_request.models import (
@@ -48,7 +56,9 @@ class PaginationWithPageCount(PageNumberPagination):
         return response
 
 class SubsidyRequestViewSet(viewsets.ModelViewSet):
-    """ Base Viewset for subsidy requests. """
+    """
+    Base Viewset for subsidy requests.
+    """
 
     permission_classes = (permissions.IsAuthenticated,)
     serializer_class = serializers.SubsidyRequestSerializer
@@ -72,9 +82,30 @@ class SubsidyRequestViewSet(viewsets.ModelViewSet):
     def lms_user_id(self):
         return self.decoded_jwt.get('user_id')
 
+    def _validate_subsidy_request_uuids(self, subsidy_request_uuids):
+        """
+        Args:
+            subsidy_request_uuids: a list of one or more valid uuid strings
+        Raises:
+            SubsidyRequestError: if subsidy UUID(s) are not preset or invalid
+        """
+        if not subsidy_request_uuids:
+            error_msg = 'You must provide subsidy request UUID(s) to be approved.'
+            logger.exception(error_msg)
+            raise SubsidyRequestError(error_msg, status.HTTP_400_BAD_REQUEST)
+
+        for subsidy_request_uuid in subsidy_request_uuids:
+            try:
+                validate_uuid(subsidy_request_uuid)
+            except ParseError as exc:
+                error_msg = f'Subsidy Request UUID provided ({subsidy_request_uuid}) is not a valid UUID'
+                logger.exception(error_msg)
+                raise SubsidyRequestError(error_msg, status.HTTP_400_BAD_REQUEST) from exc
+
     def _validate_subsidy_request(self):
         """
-        Raises a SubsidyRequestCreationError if a subsidy request cannot be created.
+        Raises:
+            SubsidyRequestCreationError: if a subsidy request cannot be created
         """
 
         enterprise_customer_uuid = self.request.data.get('enterprise_customer_uuid')
@@ -99,6 +130,27 @@ class SubsidyRequestViewSet(viewsets.ModelViewSet):
             error_msg = f'Subsidy request type must be {customer_configuration.subsidy_type}'
             raise SubsidyRequestCreationError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
 
+    def _raise_error_if_any_requests_match_incorrect_states(self, subsidy_requests, incorrect_states, current_action):
+        """
+        Args:
+           subsidy_requests: SubsidyRequest Queryset
+           incorrect_states: SubsidyRequestStates states that should throw error
+           current_action: SubsidyRequestState that we are trying to apply where this is called
+        Raises:
+            SubsidyRequestError: if any request has SubsidyRequest state == incorrect_states
+        """
+
+        subsidies_in_wrong_state = subsidy_requests.filter(state__in=incorrect_states)
+        if subsidies_in_wrong_state:
+            uuids = [str(uuid) for uuid in list(subsidies_in_wrong_state.values_list('uuid', flat=True))]
+            pretty_uuids = ','.join(uuids)
+            pretty_verbs = '/'.join(incorrect_states)
+            error_msg = (
+                    f'{self.subsidy_type} Request(s) with UUID(s) {pretty_uuids} are already {pretty_verbs}. '
+                    f'Requests could not be {current_action}.'
+                )
+            raise SubsidyRequestError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
+
     @permission_required(
         constants.REQUESTS_ADMIN_LEARNER_ACCESS_PERMISSION,
         fn=get_enterprise_uuid_from_request_data,
@@ -114,8 +166,11 @@ class SubsidyRequestViewSet(viewsets.ModelViewSet):
         request.data['lms_user_id'] = self.lms_user_id
         return super().create(request, *args, **kwargs)
 
+
 class LicenseRequestViewSet(SubsidyRequestViewSet):
-    """ Viewset for license requests. """
+    """
+    Viewset for license requests
+    """
 
     queryset = LicenseRequest.objects.order_by('-created')
     serializer_class = serializers.LicenseRequestSerializer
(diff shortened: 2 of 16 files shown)


enterprise_access/apps/api/v1/views.py
```diff
@@ -68,6 +73,25 @@ def decoded_jwt(self):
...
+    def _validate_subsidy_request_uuids(self, subsidy_request_uuids):
+        """
+        Raises a SubsidyRequestError if subsidy UUIDs are not preset or invalid.
+
+        subsidy_request_uuids should be a list of valid uuid strings
+        """
+        if not subsidy_request_uuids:
+            error_msg = 'You must provide subsidy request UUIDs to be approved.'
+            logger.exception(error_msg)
+            raise SubsidyRequestError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
```
long74100, line 85: Should these be HTTP_400_BAD_REQUESTs? Alternatively if you wanted to you could create a serializer and put the validation logic in validate(), which should automatically raise 400s.

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +159,119 @@ def _validate_subsidy_request(self):
             logger.exception(error_msg)
             raise SubsidyRequestCreationError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
 
+    def _validate_subscription_plan_uuid(self, subscription_plan_uuid):
+        """
+        Raises SubsidyRequestError if subsidy_request_uuid not present or invalid
```
long74100, line 164: subsidy_request_uuid  -> subscription_plan_uuid in comments here

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +159,119 @@ def _validate_subsidy_request(self):
...
+        try:
+            validate_uuid(subscription_plan_uuid)
+        except ParseError as exc:
+            error_msg = f'Subscription Plan UUID provided ({subscription_plan_uuid}) is not a valid UUID'
+            logger.exception(error_msg)
+            raise SubsidyRequestError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY) from exc
+
+    def _verify_subsidies_remaining(self, subscription_plan_uuid, subsidy_request_uuids):
+        """
+        Check license manager to make sure there are subscriptions remaining.
```
long74100, line 185: nit: subscriptions remaining -> licenses remaining

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +159,119 @@ def _validate_subsidy_request(self):
...
+        """
+        client = LicenseManagerApiClient()
+        subscription_overview = client.get_subscription_overview(subscription_plan_uuid)
+        status_counts = subscription_overview.get('results')
+
+        remaining_unassigned_codes = 0
+        for status_count in status_counts:
+            if status_count['status'] == 'unassigned':
```
long74100, line 193: you could also use `next(status_count['count'] for status_count in status_counts if status_count['status'] == 'unassigned')`.

nit: should we move 'unassigned' to a variable in constants?

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +159,119 @@ def _validate_subsidy_request(self):
...
+
+        if remaining_unassigned_codes < len(subsidy_request_uuids):
+            error_msg = (
+                f'Not enough licenses available for subscription {subscription_plan_uuid} '
+                'to approve the requests'
+            )
+            raise SubsidyRequestError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
+
+    @permission_required(
+        constants.REQUESTS_ADMIN_LEARNER_ACCESS_PERMISSION,
```
long74100, line 205: I think this should be REQUESTS_ADMIN_ACCESS_PERMISSION since we only want to allow admins to approve.

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +159,119 @@ def _validate_subsidy_request(self):
...
+            self._verify_subsidies_remaining(subscription_plan_uuid, subsidy_request_uuids)
+        except SubsidyRequestError as exc:
+            logger.exception(exc)
+            return Response(exc.message, exc.http_status_code)
+
+        subsidy_requests = LicenseRequest.objects.filter(uuid__in=subsidy_request_uuids)
+        with transaction.atomic():
+            for subsidy_request in subsidy_requests:
+                if subsidy_request.state == SubsidyRequestStates.DENIED:
+                    error_msg = (
```
long74100, line 228: [clarification]: if the subsidy request has already been approved, will we just ignore them in the task?

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +159,119 @@ def _validate_subsidy_request(self):
...
+
+        serialized_subsidy_requests = serializers.LicenseRequestSerializer(subsidy_requests, many=True)
+
+        return Response(
+            serialized_subsidy_requests.data,
+            status=status.HTTP_200_OK,
+        )
+
+    @permission_required(
+        constants.REQUESTS_ADMIN_LEARNER_ACCESS_PERMISSION,
```
long74100, line 246: same permission as above

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +159,119 @@ def _validate_subsidy_request(self):
...
+        try:
+            self._validate_subsidy_request_uuids(subsidy_request_uuids)
+        except SubsidyRequestError as exc:
+            logger.exception(exc)
+            return Response(exc.message, exc.http_status_code)
+
+        subsidy_requests = LicenseRequest.objects.filter(uuid__in=subsidy_request_uuids)
+        with transaction.atomic():
+            for subsidy_request in subsidy_requests:
+                subsidy_request.deny(reviewer_lms_user_id)
```
long74100, line 265: we can't deny approved requests as well I believe so we need to handle that.

enterprise_access/apps/api/v1/views.py
```diff
@@ -160,3 +298,83 @@ def _validate_subsidy_request(self):
...
+            self._validate_subsidy_request_uuids(subsidy_request_uuids)
+            self._validate_redemptions_remaining(enterprise_customer_uuid, coupon_id, subsidy_request_uuids)
+        except SubsidyRequestError as exc:
+            logger.exception(exc)
+            return Response(exc.message, exc.http_status_code)
+
+        subsidy_requests = CouponCodeRequest.objects.filter(uuid__in=subsidy_request_uuids)
+        with transaction.atomic():
+            for subsidy_request in subsidy_requests:
+                if subsidy_request.state == SubsidyRequestStates.DENIED:
```
long74100, line 335: nit: could probably pull this part out into a helper?

enterprise_access/apps/api/v1/views.py
christopappas: ty

enterprise_access/apps/api/v1/views.py
christopappas: personally think is non-fancy way it's written now is more readable 🤷 

and sure, can move that to a constant

enterprise_access/apps/api/v1/views.py
christopappas: yeah i can change to HTTP_400_BAD_REQUESTs

enterprise_access/apps/api/v1/views.py
christopappas: gooood call

enterprise_access/apps/api/v1/views.py
christopappas: we can, though it's probably better if we don't spin off a task at all. i can update this to keep track of which approval requests have already been approved so that we have the right set of things to run celery tasks for

enterprise_access/apps/api/v1/views.py
christopappas: tis 1 if-statement that returns the response to the caller, if there were more bits i would consider this, but seems  too small imo

enterprise_access/apps/api/v1/views.py
christopappas: i guess i figured once something is approved, the UI wouldn't give the option to deny it (and even if someone hit the API directly, it wouldn't change the state of the license, but this is probably the right thing to do, so i'll add it

christopappas: added logic to prevent approved and pending approved requests from being denied 👍

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +160,128 @@ def _validate_subsidy_request(self):
...
+            status=status.HTTP_200_OK,
+        )
+
+    @permission_required(
+        constants.REQUESTS_ADMIN_ACCESS_PERMISSION,
+        fn=get_enterprise_uuid_from_request_data,
+    )
+    @action(detail=False, url_path='deny', methods=['post'])
+    def deny(self, *args, **kwargs):
+        """ Action of denying a License Subsidy Request """
```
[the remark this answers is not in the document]
macdiesel: I'm not going to mark any more of these, but please do take a look at refactoring on the docs.

enterprise_access/apps/api_client/ecommerce_client.py
```diff
@@ -0,0 +1,47 @@
+"""
+API client for calls to the ecommerce service.
+"""
+from django.conf import settings
+
+from enterprise_access.apps.api_client.base_oauth import BaseOAuthClient
+
+
+class EcommerceApiClient(BaseOAuthClient):
```
macdiesel, line 9: Same question for the `LicenseManagerAPIClient`: Can we set a client side timeout of these requests to ~45 seconds? This way this code base can catch the timeout and do something sane without instead of leaving the users browser spinning or returning a generic 500 error?

enterprise_access/apps/api_client/ecommerce_client.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    API client for calls to the ecommerce service.
+    """
+    api_base_url = settings.ECOMMERCE_URL + '/api/v2/'
+    enterprise_coupons_endpoint = api_base_url + 'enterprise/coupons/'
+
+    def get_coupon_overview(self, enterprise_uuid, coupon_id):
+        """
+        Call ecommerce API overview endpoint for data about a coupon.
```
macdiesel, line 18: I assume this is copy pasta, that said, this is what we should shoot for docstring wise.

enterprise_access/apps/api_client/license_manager_client.py
```diff
@@ -0,0 +1,26 @@
...
+
+from enterprise_access.apps.api_client.base_oauth import BaseOAuthClient
+
+
+class LicenseManagerApiClient(BaseOAuthClient):
+    """
+    API client for calls to the license-manager service.
+    """
+    api_base_url = settings.LICENSE_MANAGER_URL + '/api/v1/'
+    subscriptions_endpoint = api_base_url + 'subscriptions/'
```
macdiesel, line 14: Should we set these as config options instead of inline?

enterprise_access/apps/api_client/tests/test_ecommerce_client.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+
+import mock
+from django.test import TestCase
+from requests import Response
+
+from enterprise_access.apps.api_client.ecommerce_client import EcommerceApiClient
+
+
+class TestEcommerceClient(TestCase):
```
macdiesel, line 12: Would suggest adding some error case tests.

enterprise_access/settings/devstack.py
```diff
@@ -59,3 +59,5 @@
 )
 # END CELERY
 
+ECOMMERCE_URL = 'http://ecommerce.example.com'
```
macdiesel, line 62: Are there devstack URLs that would actually work we could put in here?

macdiesel (commented):
Left a few thoughts and ideas. I did not review the tests for the views.

enterprise_access/settings/devstack.py
christopappas: actually kinda curious, if you run tests locally, does it say it's using devstack.py as the env? i found locally that test env was being ignored (and using devstack instead). so i put those there to get tests to work. but yeah, i can change these once i get the tests running against the right env

macdiesel: I did not try running, only read the code.

enterprise_access/apps/api/v1/views.py
christopappas: different approach: on deny/approve endpoints, we filter the uuids handed to us, and only actually act on things that are appropriate to act on

so if you had 5 UUIDs for approval, but 1 is already approved, we just approve 4 and do not error. and same for denying, if you go to deny a thing that is already denied, we simply do nothing instead of error

thoughts?

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +160,128 @@ def _validate_subsidy_request(self):
             logger.exception(error_msg)
             raise SubsidyRequestCreationError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
 
+    def _validate_subscription_plan_uuid(self, subscription_plan_uuid):
+        """
+        Raises SubsidyRequestError if subscription_plan_uuid not present or invalid
```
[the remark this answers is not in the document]
christopappas: weird that pydocstyle didn't complain, i will (begrudgingly) change this

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +160,128 @@ def _validate_subsidy_request(self):
...
+            error_msg = f'Subscription Plan UUID provided ({subscription_plan_uuid}) is not a valid UUID'
+            logger.exception(error_msg)
+            raise SubsidyRequestError(error_msg, status.HTTP_400_BAD_REQUEST) from exc
+
+    def _verify_subsidies_remaining(self, subscription_plan_uuid, subsidy_request_uuids):
+        """
+        Check license manager to make sure there are licenses remaining.
+        """
+        client = LicenseManagerApiClient()
+        subscription_overview = client.get_subscription_overview(subscription_plan_uuid)
```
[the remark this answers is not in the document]
christopappas: added in tries here and with the other client. please take a look at the language i use in the error message. happy to change it to whatever

enterprise_access/apps/api/v1/views.py
```diff
@@ -68,6 +74,25 @@ def decoded_jwt(self):
...
         return self.decoded_jwt.get('user_id')
 
+    def _validate_subsidy_request_uuids(self, subsidy_request_uuids):
+        """
+        Raises a SubsidyRequestError if subsidy UUIDs are not preset or invalid.
+
+        subsidy_request_uuids should be a list of valid uuid strings
+        """
+        if not subsidy_request_uuids:
+            error_msg = 'You must provide subsidy request UUIDs to be approved.'
```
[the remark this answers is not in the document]
christopappas: updated

enterprise_access/apps/api/v1/views.py
```diff
@@ -68,6 +74,25 @@ def decoded_jwt(self):
     def lms_user_id(self):
         return self.decoded_jwt.get('user_id')
 
+    def _validate_subsidy_request_uuids(self, subsidy_request_uuids):
+        """
+        Raises a SubsidyRequestError if subsidy UUIDs are not preset or invalid.
```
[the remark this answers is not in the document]
christopappas: updated

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +160,128 @@ def _validate_subsidy_request(self):
...
+            )
+            raise SubsidyRequestError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
+
+    @permission_required(
+        constants.REQUESTS_ADMIN_ACCESS_PERMISSION,
+        fn=get_enterprise_uuid_from_request_data,
+    )
+    @action(detail=False, url_path='approve', methods=['post'])
+    def approve(self, *args, **kwargs):
+        """ Action of approving a License Subsidy Request """
```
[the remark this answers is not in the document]
christopappas: updated

enterprise_access/apps/api/v1/views.py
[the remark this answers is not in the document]
christopappas: updated

enterprise_access/apps/api_client/license_manager_client.py
christopappas: probably not. LICENSE_MANAGER_URL makes sense  to be in settings, but the expected paths from that service should always be the same. if we ever update LM or ecommerce to use api/v2, then we can make a change here then

christopappas: (in particular cause these paths are used nowhere but here. and if it's any consolation, that's how it is in LM)

enterprise_access/settings/devstack.py
christopappas: alrighty, i'm inclined to file a follow up ticket/make a follow up PR to change this value in devstack.py. this PR has enough stuff in it ><'

agreed though that devstack.py and test.py should have different values

enterprise_access/apps/api_client/license_manager_client.py
macdiesel: 👍

enterprise_access/settings/devstack.py
macdiesel: Follow up ticket on this is fine.

enterprise_access/apps/api/v1/views.py
macdiesel: hrm.... I wonder what the industry at large thinks about stuff like this? IMO to be semantically correct it should either be all or nothing. That said I'm not sure that provides the best experience either. It does seem OK to let the API return a 200 in the case when you have n requests but 1+ of those are already in the state you are attempting to set. That seems reasonable. 

What about partial applies? IF only one state is able to be changed, but the others not, what should we do there? Return a partial result? A partial error? for simplicity up front it would likely be better for us to make the failure case all or nothing. But I think we are already set for this as an exception raised would cause the atomic block to rollback.

I think that's my long winded way of saying I think this is ok....

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,8 +167,153 @@ def _validate_subsidy_request(self):
...
+    def _verify_subsidies_remaining(self, subscription_plan_uuid, subsidy_request_uuids):
+        """
+        Check license manager to make sure there are licenses remaining.
+        """
+        try:
+            client = LicenseManagerApiClient()
+            subscription_overview = client.get_subscription_overview(subscription_plan_uuid)
+        except (ConnectionError, HTTPError, Timeout) as exc:
+            error_msg = (
+                'We were unable to process your request at this time. '
```
macdiesel, line 201: Maybe something more specific to leave ourselves a better breadcrumb when/if this happens?  Perhaps unable to assign (code/licenses) at this time. Please try again later or reach out ..... blah blah blah? Thoughts?

enterprise_access/settings/devstack.py
christopappas: ticket created

enterprise_access/apps/api/v1/views.py
christopappas: yeah that sounds reasonable

christopappas:
```                
'We were unable to approve/deny the selected coupon requests at this time. '
'Please try again, or reach out to customer support if you '
'continue to experience issues. '
```

and then replace coupon with license for the other 👍

enterprise_access/apps/api/v1/views.py
```diff
@@ -160,3 +308,92 @@ def _validate_subsidy_request(self):
                 f'under enterprise: {enterprise_customer_uuid}.'
             logger.exception(error_msg)
             raise SubsidyRequestCreationError(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
+
+    def _validate_redemptions_remaining(self, enterprise_customer_uuid, coupon_id, subsidy_request_uuids):
+        """
+        Make sure there are coupon redemptions remaining.
+        """
+        ecommerce_client = EcommerceApiClient()
```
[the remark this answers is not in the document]
christopappas: updated

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,6 +160,128 @@ def _validate_subsidy_request(self):
...
+        try:
+            validate_uuid(subscription_plan_uuid)
+        except ParseError as exc:
+            error_msg = f'Subscription Plan UUID provided ({subscription_plan_uuid}) is not a valid UUID'
+            logger.exception(error_msg)
+            raise SubsidyRequestError(error_msg, status.HTTP_400_BAD_REQUEST) from exc
+
+    def _verify_subsidies_remaining(self, subscription_plan_uuid, subsidy_request_uuids):
+        """
+        Check license manager to make sure there are licenses remaining.
```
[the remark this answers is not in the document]
christopappas: updated

enterprise_access/apps/api_client/tests/test_ecommerce_client.py
christopappas: so this (and the license manager) client is importing from `BaseOAuthClient` which, imports from: `from edx_rest_api_client.client import OAuthAPIClient`, because the custom logic we add is really just making the URL and then doing a GET, I personally don't think tests here are necessary (I'm making the assumption that the package we're importing from does testing of its own)

enterprise_access/apps/api/v1/views.py
christopappas: cool. i did partially mistype:

an admin tries to deny 4 requests, 1 of which has already been approved. error? or simply ignore the 1 approved and  deny the 3-not-yet-denied?

enterprise_access/apps/api_client/ecommerce_client.py
christopappas: Added in 45 second timeouts for both clients

enterprise_access/apps/api/v1/views.py
macdiesel: > an admin tries to deny 4 requests, 1 of which has already been approved. error?

Yes I do think this is an error case. That said I don't believe/am hopeful the UI wouldn't allow this state to be submitted.

enterprise_access/apps/api_client/tests/test_ecommerce_client.py
macdiesel: I was thinking about the default error cases: timeout, host not found, stuff like that. If this stuff is already covered please proceed as completed.

enterprise_access/apps/api/v1/views.py
macdiesel: Perhaps "Please try again in a little while, or ...

Alena, ideas on language for this error?

enterprise_access/apps/api/v1/views.py
```diff
@@ -134,8 +168,157 @@ def _validate_subsidy_request(self):
...
+            enterprise_customer_uuid=enterprise_customer_uuid,
+        )
+        with transaction.atomic():
+            for subsidy_request in subsidy_requests:
+                if subsidy_request.state == SubsidyRequestStates.DENIED:
+                    error_msg = (
+                        f'License Request with UUID {subsidy_request.uuid} has already been denied. '
+                        'Cannot approve the request.'
+                    )
+                    return Response(error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY)
```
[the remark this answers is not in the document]
christopappas: good question. i'm actually going to refactor this a bit to make it a little more straight forward for hooking up celery task stuff

enterprise_access/apps/api/constants.py
```diff
@@ -0,0 +1,3 @@
+""" Constants for the api app. """
+
+LICENSE_MANAGER_UNASSIGNED_STATUS = 'unassigned'
```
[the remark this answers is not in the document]
christopappas: updated

enterprise_access/settings/test.py
```diff
@@ -21,3 +21,6 @@
 results_dir = tempfile.TemporaryDirectory()
 CELERY_RESULT_BACKEND = f'file://{results_dir.name}'
 # END CELERY
+
+ECOMMERCE_URL = 'http://ecommerce.example.com'
+LICENSE_MANAGER_URL = 'http://license-manager.example.com'
```
long74100, line 26: nit: new line

long74100 (approved):
🐳

macdiesel (approved):

PR #15, opened 2022-02-07
feat: update subsidy request states

long74100:
https://openedx.atlassian.net/browse/ENT-5383


enterprise_access/apps/subsidy_request/migrations/0004_update_subsidy_states.py
```diff
@@ -0,0 +1,73 @@
...
+        'denied': SubsidyRequestStates.DECLINED
+    }
+
+    for request in LicenseRequest.objects.all().iterator():
+        request.state = new_states.get(request.state)
+        request.save()
+
+    for request in CouponCodeRequest.objects.all().iterator():
+        request.state = new_states.get(request.state)
+        request.save()
```
adamstankiewicz, line 24: optional nit/consideration: Do you think it's worth being a bit more defensive around setting the `request.state`? For example, what is `new_states.get(request.state)` doesn't return anything because the current state isn't one of the previous options? I don't think this should happen but it might be worth considering _juuust_ in case; that way we won't unintentionally update existing requests if we can deduce its updated state 🤔

adamstankiewicz (approved):
Looks good to me! Just one nit on the management command. I'll leave it as your discretion whether it's worth addressing :)

enterprise_access/apps/subsidy_request/migrations/0004_update_subsidy_states.py
adamstankiewicz: Also, one other thought that just occurred to me... while the number of objects for requests are limited right now, do you think it'd be worth doing these operations as a `bulk_update` instead of calling `.save()` on each record?

long74100: it should not happen, but it doesn't hurt to add request.state as a default so it doesn't get change.

long74100: refactored!

adamstankiewicz: Nice! 🚀

PR #17, opened 2022-02-11
feat: ENT-5401 | add auto-decline of requests when sub type changes o…

christopappas:
…n configuration model


enterprise_access/apps/api/tasks.py
```diff
@@ -1,26 +1,94 @@
...
+    SUBSIDY_TYPE_CHANGE_DECLINATION,
+    SubsidyRequestStates,
+    SubsidyTypeChoices
+)
 from enterprise_access.apps.subsidy_request.models import CouponCodeRequest, LicenseRequest
 
+logger = logging.getLogger(__name__)
+
+
+def _aliased_recipient_object_from_email(user_email):
```
macdiesel, line 23: Should this (or something similar to it) be in the braze client instead of here?

enterprise_access/apps/api/tasks.py
```diff
@@ -1,26 +1,94 @@
...
+    """
+    Returns a dictionary with a braze recipient object, including
+    a braze alias object.
```
macdiesel, line 26: Docs need a return & args statement.

enterprise_access/apps/api/tasks.py
```diff
@@ -1,26 +1,94 @@
...
+            ],
+        )
+
+    # Why I don't used subsidy_requests.update() #
+    # When you run .update() on a queryset, you "lose" the objects, because by
+    # nature of them being updated in the DB (update runs raw SQL),
+    # they no longer are returned by the original
+    # queryset. To make sure we send out notifications for the exact objects we are
+    # declining here, I've opted to use a save() in a for-loop (which the django
+    # docs even recommend in some cases).
```
macdiesel, line 68: Thank you for the explanation.

enterprise_access/apps/api_client/braze_client.py
```diff
@@ -0,0 +1,30 @@
+"""
```
macdiesel, line 1: I thought there was a library we could use for this?

macdiesel (commented):
Thank you for all the language changes. FWIW the license manager client API comment is non-blocking for this PR.

enterprise_access/apps/api/tasks.py
```diff
@@ -1,26 +1,94 @@
...
+    for subsidy_request in subsidy_requests:
+        logger.info(f'Declining subsidy {subsidy_request} because subsidy type changed on Configuration.')
+        subsidy_request.state = SubsidyRequestStates.DECLINED
+        subsidy_request.decline_reason = SUBSIDY_TYPE_CHANGE_DECLINATION
+        subsidy_request.save()
+
+    if send_notification:
```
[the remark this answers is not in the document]
christopappas: Sure, I can pull in that ticket and consider it part of this, though question:

I currently have this notification logic within the decline task (as you see), because it felt correct to keep the Config from pulling lots of DB objects from within the synchronous HTTP lifecycle, while still being able to only query the DB once for the SubsidyRequest objects. The AC of 5403 does specify the notifications being sent via their own task... how much do we care about that AC? As in, is the current set up (1 decline task with optional notifications being sent) reasonable?

enterprise_access/apps/api/tasks.py
christopappas: probably. this is copy pasted from license-manager. shall i make a ticket to have this code put in braze-client, and then enterprise-access and LM refactored to use the updated braze-client code?

enterprise_access/apps/api/tasks.py
christopappas: i need to find copy pasta with better doc strings ;)

enterprise_access/apps/api/tasks.py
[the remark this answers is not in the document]
macdiesel: My initial thought is I care a lot about that AC. I would prefer for all these notifications to be sent asynchronously via the same code paths using a celery task. A bit much for a single request? Possibly... However I would like for us to reuse as much code as possible here and for the sake of troubleshooting I would prefer all mechanisms for sending these alerts follow the same process/code paths. 

I also like the idea of the asynchronous nature of this. The assignment should fail if braze is down. Emails should be kept in queue until braze comes back online. I think there is some additional work to get us to that last state, that said separating this out in to it's own celery task takes us the first step on this journey.

enterprise_access/apps/api_client/braze_client.py
christopappas: we do. this is the [edx-braze-client](https://github.com/edx/braze-client) but this is the expected way to set up its keys, etc. tis taken from license-manager where we do the same

enterprise_access/apps/api_client/tests/test_lms_client.py
```diff
@@ -0,0 +1,39 @@
+"""
```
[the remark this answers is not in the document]
christopappas: If we are going to be more microservice oriented, I think it would be nice to have an abstract api client that can be instantiated with a handful of values depending on the service we're going to talk to (LM, enterprise-catalog, LMS, ecom, and so on)

i think someone was talking about a slumber library or something that exists 🤔 -- i'll ask around, but yeah, i think longer term we're going to want something more general

enterprise_access/apps/api_client/braze_client.py
macdiesel: 👍 Ignore this one.

enterprise_access/apps/api_client/tests/test_lms_client.py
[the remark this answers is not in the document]
macdiesel: Could you add to the tech debt tracker please?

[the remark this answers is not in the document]
christopappas: added

enterprise_access/apps/api/tasks.py
christopappas: added args. return was already in docstring

enterprise_access/apps/api/tasks.py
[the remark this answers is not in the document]
christopappas: So, okay, I will split the notifications into its own task

PR #18, opened 2022-02-16
fix: have devstack use edx.devstack.lms for lms hostname

christopappas:


enterprise_access/settings/devstack.py
```diff
@@ -69,4 +69,4 @@
 
 ECOMMERCE_URL = 'http://localhost:18130'
 LICENSE_MANAGER_URL = 'http://localhost:18170'
```
christopappas, line 71: this probably will also need to change (with ecommerce) to be `license_manager.app` and edx.devstack.ecommerce` but not 100% sure

long74100 (approved):

PR #19, opened 2022-02-17
ENT-5043 | add notification email task

christopappas:


enterprise_access/apps/api/tasks.py
```diff
@@ -133,10 +145,25 @@ def send_notification_emails_for_requests(
...
-        # Todo: add things to this dictionary once the campaign template exists
-        braze_trigger_properties = {}
+
+        # Use the contact_email of the enterprise customer user's
+        # enterprise customer object in the LMS (which we have from the API)
+        # Note this means that if you hand the task contact_email explicitly,
+        # this will overwrite it. I didn't want to overengineer something like
+        # setting a flag that detects whether or not the task got called with
+        # a particular var set, but we can do that in the future if need be
+        contact_email = enterprise_learner_data[subsidy_request.lms_user_id]['enterprise_customer']['contact_email']
```
christopappas, line 157:
```
# Use the contact_email of the enterprise customer user's
        # enterprise customer object in the LMS (which we have from the API)
        # Note this means that if you hand the task contact_email explicitly,
        # this will overwrite it. I didn't want to overengineer something like
        # setting a flag that detects whether or not the task got called with
        # a particular var set, but we can do that in the future if need be
```

enterprise_access/apps/api_client/tests/test_lms_client.py
```diff
@@ -15,17 +15,58 @@ class TestLmsApiClient(TestCase):
     Test LMS Api client.
     """
 
+    @mock.patch('requests.Response.json')
     @mock.patch('enterprise_access.apps.api_client.base_oauth.OAuthAPIClient')
-    def test_get_enterprise_customer_user_data(self, mock_oauth_client):
+    def test_get_enterprise_customer_user_data(self, mock_oauth_client, mock_json):
```
christopappas, line 20: wanted to beef this test up with something a little more concrete now that we modify the response json in the lms_client a bit before returning

enterprise_access/apps/api/tests/test_tasks.py
```diff
@@ -129,12 +134,19 @@ def test_send_notification_emails_for_requests(self, mock_braze_client, mock_lms
                 'alias_name': user_email,
             },
         }
-        mock_braze_client().send_campaign_message.assert_called_with(
+        expected_course_about_page_url = (
+            'http://enterprise-learner-portal.example.com/test-org-for-learning/course/' +
+            self.license_requests[0].course_id
+        )
```
adamstankiewicz, line 140: nit: Would this make sense as an `f` string, e.g. `f'http://enterprise-learner-portal.example.com/test-org-for-learning/course/{self.license_requests[0].course_id}'`?

enterprise_access/apps/api/v1/tests/test_views.py
```diff
@@ -459,6 +462,46 @@ def test_approve_subsidy_request_success(self, mock_get_sub, _):
...
+            'subsidy_request_uuids': [self.user_license_request_1.uuid],
+            'subscription_plan_uuid': self.user_license_request_1.subscription_plan_uuid,
+            'send_notification': True,
+        }
+        response = self.client.post(LICENSE_REQUESTS_APPROVE_ENDPOINT, payload)
+        assert response.status_code == status.HTTP_200_OK
+
+
+
+
```
adamstankiewicz, line 504: nit: extra newlines

enterprise_access/apps/api/tasks.py
```diff
@@ -113,14 +114,25 @@ def decline_enterprise_subsidy_requests_task(subsidy_request_uuids, subsidy_type
...
         subsidy_request.save()
 
-@shared_task(base=LoggedTask)
+
+@shared_task(base=LoggedTaskWithRetry)
 def send_notification_emails_for_requests(
         subsidy_request_uuids,
         braze_campaign_id,
         subsidy_type,
+        braze_trigger_properties,
```
adamstankiewicz, line 123: nit: It seems like we're passing `{}` for this argument in many places. I'm wondering if it might make sense to add `{}` as a default/fallback value when it's not explicitly provided?

adamstankiewicz (approved):
Looks good! Just 3 nits/questions.

enterprise_access/apps/api/tasks.py
christopappas: fun python thing: If you add `{}` as a default, there will only ever be 1 dictionary for that task in the runtime (so each time you call that function, it would refer to the _same_ dictionary)

A common pattern to avoid this is to set it to None and then have an if-statement to make the var a dict if it is None. I can do that

enterprise_access/apps/api/v1/tests/test_views.py
christopappas: forgot 1 commit! pushed this up

enterprise_access/apps/api/tests/test_tasks.py
christopappas: I split this up because the line was too long, I think either way is fine? I can combine 2 f strings together though

christopappas: Twas a long line, which is why I split it up this way

enterprise_access/apps/api/v1/tests/test_views.py
```diff
@@ -934,6 +974,26 @@ def test_decline_request_success(self):
             state=SubsidyRequestStates.DECLINED
         ).count() == 1
 
+    @mock.patch('enterprise_access.apps.api.v1.views.send_notification_emails_for_requests.apply_async')
+    def test_decline_send_notifcation(self, mock_notify):
```
long74100, line 978: nit: typo here

long74100 (approved):

PR #20, opened 2022-02-18
feat: add assign license and coupon code tasks

long74100:
https://openedx.atlassian.net/browse/ENT-5402

- Create tasks for assigning licenses and coupon codes
- Chain them so that the retries are separate (We don't want to retry license/coupon code assignment codes if our db commit fails at the end)


manny-m (commented):
Had one clarifying question, otherwise looks good 👍

enterprise_access/apps/api/v1/views.py
```diff
@@ -423,47 +433,56 @@ def approve(self, *args, **kwargs):
...
             return Response(exc.message, exc.http_status_code)
 
-        subsidies_to_approve = subsidy_requests.filter(
+        coupon_code_requests_to_approve = coupon_code_requests.filter(
             state__in=[SubsidyRequestStates.REQUESTED, SubsidyRequestStates.ERROR]
         )
         with transaction.atomic():
-            for subsidy_request in subsidies_to_approve:
-                subsidy_request.approve(reviewer_lms_user_id)
+            for coupon_code_request in coupon_code_requests_to_approve:
```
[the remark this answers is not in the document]
long74100: We only want to mark individual requests as approved

enterprise_access/apps/api/tasks.py
```diff
@@ -37,6 +64,33 @@ def _aliased_recipient_object_from_email(user_email):
         },
     }
 
+def _get_enterprise_learner_data(lms_user_ids):
```
christopappas, line 67: cool

enterprise_access/apps/api/tasks.py
```diff
@@ -37,6 +64,33 @@ def _aliased_recipient_object_from_email(user_email):
...
+    """
+
+    lms_client = LmsApiClient()
+    unique_lms_user_ids = set(lms_user_ids)
+    learner_data = lms_client.get_enterprise_learner_data(unique_lms_user_ids)
+
+    # Sanity check to make sure that we have user data for each lms_user_id
+    missing_lms_user_ids = [
+        lms_user_id for lms_user_id in unique_lms_user_ids if lms_user_id
+        not in learner_data
```
christopappas, line 85: is learner_data here a dict?

enterprise_access/apps/api/tasks.py
```diff
@@ -89,3 +144,177 @@ def send_notification_emails_for_requests(
...
+    learner_data = license_assignment_results['learner_data']
+    assigned_licenses = license_assignment_results['assigned_licenses']
+    subscription_uuid = license_assignment_results['subscription_uuid']
+
+    license_requests = LicenseRequest.objects.filter(
+        uuid__in=license_request_uuids
+    )
+
+    for license_request in license_requests:
+        user_email = learner_data[str(license_request.lms_user_id)]['email']
```
christopappas, line 226: can this potentially throw a key error?

enterprise_access/apps/api/tasks.py
```diff
@@ -89,3 +144,177 @@ def send_notification_emails_for_requests(
...
+    return {
+        'coupon_code_request_uuids': coupon_code_request_uuids,
+        'learner_data': learner_data,
+        'assigned_codes': assigned_codes,
+        'coupon_id': coupon_id
+    }
+
+
+@shared_task(base=LoggedTaskWithRetry)
+def update_coupon_code_requests_after_assignments_task(coupon_code_assignment_results):
```
christopappas, line 284: for this and the other "_after_assignments_task" -- the functionality is fine, but i personally don't love this function's name "knowing" where it's going to get called, because there's technically nothing special about being called after the other task

christopappas (approved):
overall 👍 -- i don't want to block you with the naming things

enterprise_access/apps/api/tasks.py
long74100: leaving for now since removing _after_assignments_task makes it too general.

PR #24, opened 2022-03-01
feat: add overview endpoint

long74100:
https://openedx.atlassian.net/browse/ENT-5450

- basically license manager overview endpoint ported over, kept it the same for consistency
- add user_email field because the ui needs this information (not backfilling in the migration since our db isn't populated yet)  @christopappas pls advise if this is ok


enterprise_access/settings/base.py
```diff
@@ -369,3 +369,8 @@ def root(*path_fragments):
 BRAZE_API_URL = ''
 BRAZE_API_KEY = os.environ.get('BRAZE_API_KEY', '')
 BRAZE_APP_ID = os.environ.get('BRAZE_APP_ID', '')
+
+# shell_plus
+SHELL_PLUS_IMPORTS = [
```
christopappas, line 374: what's this?

long74100: This automatically loads additional imports when using shell_plus. I found myself needing to import localized_utcnow in the shell so this makes life easier.

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -34,6 +34,8 @@ class SubsidyRequest(TimeStampedModel, SoftDeletableModel):
 
     lms_user_id = models.IntegerField()
 
+    user_email = models.EmailField()
```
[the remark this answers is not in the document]
long74100: I am not opposed to this approach, we don't really care about snapshotting. As long as we're guaranteed to have an email (aka the user model) for each request.

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -32,7 +32,11 @@ class SubsidyRequest(TimeStampedModel, SoftDeletableModel):
         unique=True,
     )
 
-    lms_user_id = models.IntegerField()
+    user = models.ForeignKey(
```
christopappas, line 35: noice

christopappas (approved):

enterprise_access/apps/api/serializers.py
```diff
@@ -35,9 +38,9 @@ class Meta:
         read_only_fields = [
             'uuid',
             'state',
+            'email',
```
adamstankiewicz, line 41: 🎉

adamstankiewicz (approved):

PR #34, opened 2022-03-10
ENT-5215 | Command to send email for newly created requests

christopappas:
This PR creates the management command that is run. There is a companion PR for the cron functionality that will be shared internally


enterprise_access/apps/api/tasks.py
```diff
@@ -11,37 +11,14 @@
...
+from enterprise_access.apps.subsidy_request.constants import SUBSIDY_TYPE_CHANGE_DECLINATION, SubsidyRequestStates
 from enterprise_access.apps.subsidy_request.models import CouponCodeRequest, LicenseRequest
 from enterprise_access.tasks import LoggedTaskWithRetry
-from enterprise_access.utils import get_subsidy_model
+from enterprise_access.utils import get_aliased_recipient_object_from_email, get_subsidy_model
 
 logger = logging.getLogger(__name__)
 
 
-def _get_aliased_recipient_object_from_email(user_email):
```
christopappas, line 26: moved this to a place accessible to our different django apps in this project

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -27,3 +34,72 @@ def update_course_title_for_subsidy_request_task(subsidy_type, subsidy_request_u
...
+    """
+    braze_client = BrazeApiClient()
+    lms_client = LmsApiClient()
+    enterprise_customer_data = lms_client.get_enterprise_customer_data(enterprise_customer_uuid)
+
+    contact_email = enterprise_customer_data.get('contact_email')
+    if not contact_email:
+        logger.warning(
+            f'No contact email set for {enterprise_customer_uuid}. '
+            'Not sending new subsidy request email to admin.'
```
long74100, line 55: galactical nit: request(s)?

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -27,3 +34,72 @@ def update_course_title_for_subsidy_request_task(subsidy_type, subsidy_request_u
...
+    coupon_model = get_subsidy_model(SubsidyTypeChoices.COUPON)
+    coupon_subsidies = coupon_model.objects.filter(
+        enterprise_customer_uuid=enterprise_customer_uuid,
+        state=SubsidyRequestStates.REQUESTED,
+    )
+
+    # Go through all subsidy_requests, licenses and coupons, and gather
+    # their data into braze_trigger_properties, so we can send 1 email
+    # with all the bits we need.
+    braze_trigger_properties = OrderedDict()
```
long74100, line 77: q: does this need to be an ordered dict?

enterprise_access/apps/subsidy_request/migrations/0008_auto_20220310_2004.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.2.11 on 2022-03-10 20:04
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
long74100, line 6: Let's name the migration file.

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -27,3 +34,72 @@ def update_course_title_for_subsidy_request_task(subsidy_type, subsidy_request_u
 
     # Use bulk_update so we don't trigger save() again
     subsidy_model.bulk_update([subsidy_request], ['course_title'])
+
+
+@shared_task(base=LoggedTaskWithRetry)
+def send_admins_email_with_new_requests_task(enterprise_customer_uuid):
```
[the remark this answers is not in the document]
macdiesel: Long what do you mean by
> pull both license & coupon codes in request agnostically
?

enterprise_access/apps/subsidy_request/tasks.py
christopappas: yeah. the request is to have results listed newest to oldest in the email, and while the query set is ordered, regular dicts are just hashes and order is not deterministic

long74100: resolved: only the request need to be ordered, not the trigger properties dict.

enterprise_access/apps/subsidy_request/tasks.py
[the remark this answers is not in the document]
long74100: this task currently sends both new license & coupon code requests in the same email, but we need a distinction between them since there is a button in the email that leads the admin to the correct page.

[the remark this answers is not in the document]
christopappas: fixed in latest PR (dynamically creates the url based on the subsidy type associated with the subsidy config)

thanks for this callout btw

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -17,6 +17,26 @@ class LmsApiClient(BaseOAuthClient):
...
+        Gets the data for an EnterpriseCustomer for the given uuid.
+
+        Arguments:
+            enterprise_customer_uuid (string): id of the enterprise customer
+        Returns:
+            dictionary containing enterprise customer metadata
+        """
+
+        try:
+            endpoint = f'{self.enterprise_learner_endpoint}/{enterprise_customer_uuid}'
```
adamstankiewicz, line 33: Should this be `self.enterprise_customer_endpoint`?

adamstankiewicz: Related, I think you have a double `/` situation going on here as well?

christopappas: good catch. seems i forgot a test on this too. fixed this and added a test

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -27,3 +33,87 @@ def update_course_title_for_subsidy_request_task(subsidy_type, subsidy_request_u
 
     # Use bulk_update so we don't trigger save() again
     subsidy_model.bulk_update([subsidy_request], ['course_title'])
+
+
+def _get_manage_requests_url(subsidy_model, enterprise_slug):
+    """
+    Get a manage_learners url based on the type of subsidy.
```
long74100, line 40: nit: manage_requests url*

christopappas: fixed

enterprise_access/apps/subsidy_request/management/commands/send_admins_email_with_new_requests.py
```diff
@@ -0,0 +1,44 @@
...
+    last_updated_date field on the SubsidyRequestCustomerConfiguration model. This gives us 100%
+    coverage of requests to email about, and also
+    gives us fault tolerance if this ever fails. We can simply rerun this
+    and have it include all requests that haven't been included in an email yet).
+
+    This DOESN'T give us fine grained control of different enterprises wanting different
+    email frequencies, but that's not part of the MVP. If and when we would want that, we
+    can simply give config models a field that holds how frequently admins want to be notified
+    and then update the cronjob to run more frequently (a "higher resolution") to let us
+    include all use cases.
```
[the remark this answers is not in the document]
christopappas: 🚀

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -27,3 +33,87 @@ def update_course_title_for_subsidy_request_task(subsidy_type, subsidy_request_u
...
+    Task to send new-request emails to admins.
+
+    Args:
+        enterprise_customer_uuid (str): enterprise customer uuid identifier
+    """
+    braze_client = BrazeApiClient()
+    lms_client = LmsApiClient()
+    enterprise_customer_data = lms_client.get_enterprise_customer_data(enterprise_customer_uuid)
+
+    contact_email = enterprise_customer_data.get('contact_email')
```
long74100, line 69: clarification: Is this the admin email/the email we want to use? Most of the contact email is null for the enterprises (only 23 have it set). Or should we fetch enterprise admins and send emails to those?

christopappas: I honestly thought this was the field we were using in other places :/ 

what do you mean by "fetch enterprise admins" in this case?

long74100: we can query the lms for the enterprise admins, there is an example in license-manager.

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -17,6 +17,73 @@ class LmsApiClient(BaseOAuthClient):
...
+        """
+
+        try:
+            endpoint = f'{self.enterprise_customer_endpoint}{enterprise_customer_uuid}'
+            response = self.client.get(endpoint, timeout=settings.LMS_CLIENT_TIMEOUT)
+        except requests.exceptions.HTTPError as exc:
+            logger.exception(exc)
+            raise
+
+        return response.json()
```
macdiesel, line 39: Why not move the json parsing up and log that as well? Also we know that the method can raise at least one exception, this should be covered in the docs.

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -17,6 +17,73 @@ class LmsApiClient(BaseOAuthClient):
...
+                response.raise_for_status()
+                resp_json = response.json()
+                url = resp_json['next']
+
+                for result in resp_json['results']:
+                    user_data = result['user']
+                    user_data.update(ecu_id=result['id'], created=result['created'])
+                    results.append(user_data)
+
+            return results
```
macdiesel, line 79: Wondering if we have guidelines on where these return statements should be? I would have expected this to be after the except block as it is on the other method above.

enterprise_access/apps/subsidy_request/management/commands/send_admins_email_with_new_requests.py
```diff
@@ -0,0 +1,44 @@
...
+from enterprise_access.apps.subsidy_request.models import SubsidyRequestCustomerConfiguration
+from enterprise_access.apps.subsidy_request.tasks import send_admins_email_with_new_requests_task
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    This command is intended to run as frequently as specified in a crontab. The window of time used
+    to grab requests that were made will be from the time this is last run (stored in the
```
macdiesel, line 18: Opening ( without a )

macdiesel: Could you restate this for clarity: 

> The window of time used  to grab requests that were made will be from the time this is last run 

Do you mean that the command usees the last _updated_field on SubsidyRequestCustomerConfiguration model to determine the window for the job run?

enterprise_access/apps/subsidy_request/management/commands/send_admins_email_with_new_requests.py
```diff
@@ -0,0 +1,44 @@
...
+    and have it include all requests that haven't been included in an email yet).
+
+    This DOESN'T give us fine grained control of different enterprises wanting different
+    email frequencies, but that's not part of the MVP. If and when we would want that, we
+    can simply give config models a field that holds how frequently admins want to be notified
+    and then update the cronjob to run more frequently (a "higher resolution") to let us
+    include all use cases.
+    """
+    help = (
+        'Spin of celery tasks to send enterprise admins an email that lists requests that '
```
macdiesel, line 31: "Spin **up** celery tasks"?

enterprise_access/apps/subsidy_request/management/commands/send_admins_email_with_new_requests.py
```diff
@@ -0,0 +1,44 @@
...
+        'came in since this was last run.'
```
macdiesel, line 32: Came in = made? Requested? since last run.

enterprise_access/apps/subsidy_request/management/commands/send_admins_email_with_new_requests.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def handle(self, *args, **options):
+        enterprise_customer_uuids = SubsidyRequestCustomerConfiguration.objects.filter(
+            subsidy_requests_enabled=True
+        ).values_list(
+            'enterprise_customer_uuid',
+            flat=True,
+        )
+
+        for enterprise_customer_uuid in enterprise_customer_uuids:
```
macdiesel, line 43: Perhaps this is accounted for in the logging elsewhere but if not could you please log the enterprises, their UUIDs and if not too hard the count of requests in each email (if harder than adding a one line logging statement this final bit is not necessary).

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -27,3 +34,95 @@ def update_course_title_for_subsidy_request_task(subsidy_type, subsidy_request_u
...
+
+
+@shared_task(base=LoggedTaskWithRetry)
+def send_admins_email_with_new_requests_task(enterprise_customer_uuid):
+    """
+    Task to send new-request emails to admins.
+
+    Args:
+        enterprise_customer_uuid (str): enterprise customer uuid identifier
+    """
```
macdiesel, line 65: You know it raises at least one exception. It should be part of the docs.

enterprise_access/apps/subsidy_request/tasks.py
christopappas: updated in latest commit

macdiesel (commented):
A few thoughts.

enterprise_access/apps/api_client/lms_client.py
christopappas: Hmm, do you mean like adding in a `logger.info(response.json())` type line to the except block? or having one outside of the accept block? both?

In the case of "log it each time" and not just the except block, I think eventually there would be cost implications, and also just really noisy. If you just meant the except block, I think am okay with adding that there.

And in any case, I can update the doc string

enterprise_access/apps/subsidy_request/management/commands/send_admins_email_with_new_requests.py
christopappas: I think yes, but lemme say again:

We plan to run the job every 12 hours, which is going to be defined in configuration/crontab elsewhere.

The email that gets sent out will only include subsidy requests created later than or equal to the last time this job ran. And the timestamp for when it was last run is on the `last_remind_date` on the SubsidyRequestCustomerConfiguration model

enterprise_access/apps/subsidy_request/management/commands/send_admins_email_with_new_requests.py
christopappas: spin off i think. will change

enterprise_access/apps/subsidy_request/tasks.py
christopappas: This is a recurring theme where you ask me to put things in the docstrings. Do we have linting for that?

long74100: I believe this is definitely a tech debt item, we do not have consistent documentation, and I think I actually logged it at one point.

enterprise_access/apps/subsidy_request/management/commands/tests/test_send_admins_email_with_new_requests.py
```diff
@@ -0,0 +1,248 @@
...
+        for _ in range(2):
+            factories.LicenseRequestFactory(
+                enterprise_customer_uuid=self.enterprise_customer_uuid,
+                state=SubsidyRequestStates.REQUESTED,
+            )
+
+        factories.SubsidyRequestCustomerConfigurationFactory(
+            enterprise_customer_uuid=self.enterprise_customer_uuid,
+            subsidy_requests_enabled=True,
+            last_remind_date=datetime.now()
```
long74100, line 243: should last_remind_date be None here since we want the licenses to be created after the last_remind_date?

long74100 (approved):
this lgtm, thanks for making the changes :)

enterprise_access/apps/subsidy_request/management/commands/tests/test_send_admins_email_with_new_requests.py
christopappas: i think in this test we want no sub requests to be applicable 

```
past | ---> request1, r2, r3 -----> last_remind_date ------> no sub requests after ---> | future
```
^ if we run the command then there will be no requests between `last_remind_date` and the future, which is what we want here i think

enterprise_access/apps/subsidy_request/tasks.py
christopappas: techdebt item added and put in a line for raises in latest commit

enterprise_access/apps/subsidy_request/management/commands/send_admins_email_with_new_requests.py
christopappas: reworded!

PR #42, opened 2022-03-23
fix: send correct enterprise slug to braze, make notification task retry safe

long74100:
- fix sending the wrong slug to braze, removed _get_enterprise_learner_data since we don't need this anymore
- refactor task for sending emails, it's not safe to retry a task that sends emails in a loop
- rename variables for clarity

### DIFF
--- a/enterprise_access/apps/api/tasks.py
+++ b/enterprise_access/apps/api/tasks.py
@@ -6,7 +6,6 @@ import logging
 from celery import shared_task
 from django.conf import settings
 
-from enterprise_access.apps.api.exceptions import MissingEnterpriseLearnerDataError
 from enterprise_access.apps.api.serializers import CouponCodeRequestSerializer, LicenseRequestSerializer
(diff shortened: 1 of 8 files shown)


enterprise_access/apps/api/tasks.py
```diff
@@ -94,20 +65,20 @@ def decline_enterprise_subsidy_requests_task(subsidy_request_uuids, subsidy_type
 
 
 @shared_task(base=LoggedTaskWithRetry)
-def send_notification_emails_for_requests(
-        subsidy_request_uuids,
+def send_notification_email_for_request(
```
macdiesel, line 68: I like this single use version of the function much more than the multi use version as it puts the onus on the caller to properly handle any exceptions that may occur during execution. My theory being that the caller knows better how to handle a single failure within the workflow than this generic function does. 👍

macdiesel (approved):

adamstankiewicz (approved):
LGTM

PR #48, opened 2022-03-30
feat: Add database indexes

macdiesel:


enterprise_access/apps/subsidy_request/models.py
```diff
@@ -112,6 +116,10 @@ def bulk_update(cls, subsidy_requests, field_names, batch_size=SUBSIDY_REQUEST_B
 
     class Meta:
         abstract = True
+        index_together = [
+            ['uuid', 'state'],
```
iloveagent57, line 120: `uuid` is the PK of this table, so does make sense to include it in a composite index?

macdiesel: My thought here is that this would be in support of this query pattern: https://github.com/edx/enterprise-access/blob/8573d80cb21f761699652953f1c624703ed93ebb/enterprise_access/apps/api/tasks.py#L141

It's been a while but my recollection is that by grouping these together it would be faster than adding an index just on state. That said this particular query is part of a celery task so perhaps not as important to optimize here. On the other side its basically free so why not?

iloveagent57 (approved):

PR #51, opened 2022-04-11
feat: send events to kafka

long74100:


enterprise_access/apps/events/data.py
```diff
@@ -0,0 +1,123 @@
...
+    Wrapper class used to ensure a single instance of the CouponCodeRequestEventSerializer.
+    This avoids errors on startup.
+    """
+
+    SERIALIZER = None
+
+    @classmethod
+    def get_serializer(cls):
+        """
+        Get or create a single instance of the TrackingEvent serializer to be used throughout the life of the app
```
rgraber, line 91: update comment (still says TrackingEvent)

enterprise_access/apps/events/utils.py
```diff
@@ -0,0 +1,147 @@
...
+from confluent_kafka.error import ValueSerializationError
+from confluent_kafka.serialization import StringSerializer
+from django.conf import settings
+
+from enterprise_access.apps.events.data import CouponCodeRequestEvent, CouponCodeRequestEventSerializer
+
+logger = logging.getLogger(__name__)
+
+
+# def generate_coupon_code_request_data(coupon_code_request):
```
rgraber, line 19: Can we either remove this block comment or add a meta-comment explaining why it has been left in?

enterprise_access/apps/events/utils.py
```diff
@@ -0,0 +1,147 @@
...
+            'sasl.mechanism': 'PLAIN',
+            'security.protocol': 'SASL_SSL',
+            'sasl.username': settings.KAFKA_API_KEY,
+            'sasl.password': settings.KAFKA_API_SECRET,
+        })
+
+    admin_client = AdminClient(KAFKA_ACCESS_CONF_BASE)
+
+    topic = NewTopic(topic_name,
+                                   num_partitions=settings.KAFKA_PARTITIONS_PER_TOPIC,
```
rgraber, line 94: spacing

enterprise_access/apps/events/utils.py
```diff
@@ -0,0 +1,147 @@
...
+def verify_event(err, evt):
+    """
+    Simple callback method for debugging event production
+    :param err: Error if event production failed
+    :param evt: Event that was delivered
+    """
+    if err is not None:
+        logger.warning(f"Event delivery failed: {err}")
+    else:
+        # Don't log msg.value() because it may contain userids and/or emails
```
rgraber, line 145: This comment is now inaccurate since there are no emails in the new event and we don't actually care about userids since they are not PII

enterprise_access/apps/events/utils.py
```diff
@@ -0,0 +1,147 @@
...
+                'sasl.username': getattr(settings, 'KAFKA_API_KEY', ''),
+                'sasl.password': getattr(settings, 'KAFKA_API_SECRET', ''),
+            })
+
+        new_producer = SerializingProducer(producer_settings)
+        cls._type_to_producer[event_type] = new_producer
+        return new_producer
+
+
+def create_topic_if_not_exists(topic_name):
```
rgraber, line 76: now that we have two topics it'd be more efficient to update this to take a list of topics instead of a single one since create_topics expects a list anyway

rgraber (changes requested):
Couple things:
Wherever there are 'we need to do this later' comments, please add a ## TODO (EventBus) note to the top. we'll be using it to clean up prototyping code later
What is the LicenseRequest used for?

long74100:
@rgraber the LicenseRequestData (renamed) is for the other event that will be produced in this service. Not used yet but figured that could stay.

enterprise_access/apps/events/data.py
```diff
@@ -0,0 +1,126 @@
...
+            KAFKA_SCHEMA_REGISTRY_CONFIG = {
+                'url': getattr(settings, 'SCHEMA_REGISTRY_URL', ''),
+                'basic.auth.user.info': f"{getattr(settings,'SCHEMA_REGISTRY_API_KEY','')}"
+                f":{getattr(settings,'SCHEMA_REGISTRY_API_SECRET','')}",
+            }
+            schema_registry_client = SchemaRegistryClient(KAFKA_SCHEMA_REGISTRY_CONFIG)
+            cls.TRACKING_EVENT_SERIALIZER = AvroSerializer(schema_str=CouponCodeRequestEvent.AVRO_SCHEMA,
+                                                           schema_registry_client=schema_registry_client,
+                                                           to_dict=CouponCodeRequestEvent.to_dict)
+            return cls.TRACKING_EVENT_SERIALIZER
```
dianakhuang, line 107: nit: This return statement is redundant.

dianakhuang (approved):
Looks good to me! I do think in the future we won't want to create topics in our Python code as that can and should be handled by terraform, but for these early use cases, we don't have to care about that.

enterprise_access/apps/events/apps.py
```diff
@@ -0,0 +1,31 @@
...
+
+from django.apps import AppConfig
+from django.conf import settings
+from edx_toggles.toggles import SettingToggle
+
+from enterprise_access.apps.events.utils import create_topics
+
+logger = logging.getLogger(__name__)
+
+KAFKA_ENABLED_TOGGLE = SettingToggle("KAFKA_ENABLED", default=False)
```
christopappas, line 15: do we still want this or are we just going to use settings.KAFKA_ENABLED?

enterprise_access/settings/devstack.py
```diff
@@ -78,3 +78,17 @@
...
+################### Kafka Related Settings ##############################
+KAFKA_ENABLED=False
+
+KAFKA_BOOTSTRAP_SERVER = 'edx.devstack.kafka:29092'
+SCHEMA_REGISTRY_URL = 'http://edx.devstack.schema-registry:8081'
+KAFKA_REPLICATION_FACTOR_PER_TOPIC=1
+
+COUPON_CODE_REQUEST_TOPIC_NAME="coupon-code-request-dev"
+LICENSE_REQUEST_TOPIC_NAME="license-request-dev"
+KAFKA_TOPICS = [COUPON_CODE_REQUEST_TOPIC_NAME, LICENSE_REQUEST_TOPIC_NAME]
```
christopappas, line 92: do we want to add a blank entry for the topics in base.py settings?

enterprise_access/apps/events/utils.py
```diff
@@ -0,0 +1,133 @@
...
+        if existing_producer is not None:
+            return existing_producer
+
+        producer_settings = {
+            'bootstrap.servers': getattr(settings, 'KAFKA_BOOTSTRAP_SERVER', ''),
+            'key.serializer': event_key_serializer,
+            'value.serializer': event_value_serializer,
+        }
+
+        if settings.KAFKA_API_KEY and settings.KAFKA_API_SECRET:
```
christopappas, line 46: what should happen if we don't have the key and secret set? basically just expect us not be able to communicate with kafka? given that we have the getattr with default of `''`, not sure this if is necessary

enterprise_access/apps/events/utils.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def create_topics(topic_names):
+    """
+    Create topics in the event bus
+    :param topic_names: topics to create
+    """
+    KAFKA_ACCESS_CONF_BASE = {'bootstrap.servers': getattr(settings, 'KAFKA_BOOTSTRAP_SERVER', '')}
+
+    if settings.KAFKA_API_KEY and settings.KAFKA_API_SECRET:
```
christopappas, line 66: and here, i guess i don't super care how we handle whether or not the settings are there, but should these look the same for consistency's sake? 🤷

enterprise_access/apps/events/data.py
```diff
@@ -0,0 +1,129 @@
...
+            schema_registry_client = SchemaRegistryClient(KAFKA_SCHEMA_REGISTRY_CONFIG)
+            cls.TRACKING_EVENT_SERIALIZER = AvroSerializer(schema_str=CouponCodeRequestEvent.AVRO_SCHEMA,
+                                                           schema_registry_client=schema_registry_client,
+                                                           to_dict=CouponCodeRequestEvent.to_dict)
+
+        return cls.TRACKING_EVENT_SERIALIZER
+
+
+@attr.s(frozen=True)
+class LicenseRequestData:
```
christopappas, line 115: is it worth adding licenseRequestEvent and licenseRequestEventSerializer now? or not bother cause LM is not part of this first use case yet?

christopappas (approved):
basically LGTM, just have a handful of nitty questions (that don't necessarily require action), but nothing major

enterprise_access/apps/events/apps.py
long74100: yeah we don't need this, will remove

enterprise_access/apps/events/utils.py
long74100: That would be what we expect, this is necessary because we don't want to set 'sasl.mechanism' and 'security.protocol' if there is no key/secret.

enterprise_access/apps/events/data.py
long74100: Yeah not worth, we're going to end up moving these definitions to openedx-events anyway so we can do it then.

PR #56, opened 2022-04-21
docs: add adr for auto enrollment post request approval

long74100:


docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,113 @@
...
+
+Context
+=======
+
+Learners are currently able to request a subsidy (license or coupon code) through the enterprise-access service.
+After a request is approved, the learner still has to redeem the subsidy and enroll in a course manually.
+Since the goal of a learner is to access a course, we want to automatically redeem the subsidy and enroll
+them after a request approval.
+
+For coupon code requests, requiring manual redemption is not ideal. We cannot that the coupon code is only used
```
christopappas, line 18: typo: We cannot confirm*?

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,113 @@
...
+for the course it was requested for, and have no good way of tracking the usage of the coupon code.
+
```
christopappas, line 20: Should we explicitly add a line here about license subsidy_request approval for thoroughness? (Even if this event-bus stuff doesn't directly impact the flow)

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,113 @@
...
+i.e. when a coupon code offers 50% discount, a license request without course id, etc.. We will add a field `attempt_auto_enrollment`
+to the `SubsidyRequestCustomerConfiguration` model to allow for the auto-enrollment to be turned on/off.
+
+As for handling auto enrollments, there are a few options:
+
+Proposed Solution 1
+===================
+The subsidy granting services will consume the approval events, redeem the granted subsidy,
+and enroll a learner in the course specified in the request.
+
```
christopappas, line 47: test

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,113 @@
...
+Changes required
+----------------
+* We will no longer send an email to the learner at the time of approval. Instead, we will send an email to
+  the learner when they have been enrolled.
+* enterprise-access will consume enrollment events from the LMS, check if the enrollment corresponds to an approved
+  request by matching on `lms_user_id`` and `course_id`, and send an approval email to the learner if a matching
+  request is found.
+* We migth not need the above changes if we make the messaging in the approval email generic enough,
```
christopappas, line 55: migth -> might

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,113 @@
...
+that a subsidy has been granted.
+
+
+Changes required
+----------------
+* We will introduce a new subsidy request state `fulfilled` that signifies a subsidy has been granted.
+* enterprise-access will no longer call the services directly to grant a subsidy.
+* A subsidy request is fulfilled when a subsidy is granted, so the subsidy granting services will have to send an
+  event after granting a subsidy to be consumed by enterprise-access. The state of a subsidy request will be changed
+  from `approved` to `fulfilled` upon consumption of such an event event.
```
christopappas, line 85: event x2

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,113 @@
...
+* We could add a new state `fulfilled` on the subsidy request models, but unless we need to store that information,
+  I'd propose not changing any of the current models.
+
+Consequences
+------------
+* The learner could see a granted subsidy and enroll manually in course before auto-enrollment occurs.
+* Auto-enrollment could fail and enterprise-access won't receive an enrollment event. The learner will still have their
+  subsidy and can enroll manually but they might not receive an request approved email.
+* Matching an enrollment to a subsidy request based on lms_user_id and course_id is not guaranteed to be reliable.
+  We might want an identifier on the events produced/consumed to tie them to a single interaction.
```
christopappas, line 67: My concern with passing an event-specific identifier across service boundaries is that it starts to connect events to one another when, to be truly decoupled, events should be as dumb as possible, and just snapshots of the state of data/thing happening at a given point in time. I also have additional questions of "How would a service know when an event started with itself? Do we store that some place in the service?"

I think the antidote to adding event-tracking complexity comes in the form of generating events with rich enough data that a consuming service can make good decisions around what to do when it sees the events. Something that came up in our chats before was maybe using the source of an enrollment as a way for enterprise-access to understand the nature of the enrollment's creation (e.g. https://github.com/openedx/edx-enterprise/blob/master/enterprise/models.py#L1432)

As a side note, I don't think an event-specific UUID is quite the same as duplicating data across services like we do with enterprise_customer_uuids for example, because `enterprise_customer_uuid` itself is a piece of metadata that has meaning outside of the event lifecycle.

long74100: I do not believe the identifier has to be an event-specific UUID, but perhaps an identifier that allows us to trace event propagation. I don't think the event becomes less dumb in the presence of something like this. The additional questions are something I have to think about a little more. 

Or perhaps is the problem here that we have a loop in our architecture and perhaps that should be avoided?

christopappas: Agreed that technically adding a UUID doesn't make the event itself less-dumb. And I don't think the looping itself is bad. To me, the crux is more about how/why the looping occurs. 

In the case where we feel the need to add an identifier to trace events across services, I think that means our events are not descriptive enough with the data we have to understand the flow, or we need to categorize our event streams differently. In theory, with just an lms_user_id, enterprise_customer_uuid, and the timestamp for when something is approved in enterprise-access, we should be able to examine logs to understand the flow of what events were produced/consumed across systems (provided we're logging enough)

I worry that an additional identifier bleeds info specific to enterprise-access downstream beyond one consuming service. For example, ecommerce consumes from enterprise-access, so any UUIDs, etc. enterprise-access has is perfectly fine to be expected when ecommerce consumes that event. However, when LMS consumes data from ecommerce, why should the LMS know that the event emitted from ecommerce was triggered by a change in enterprise-access? To truly decouple the services, the LMS can't know what happened before ecommerce emitted its event. Additionally, there is the question again of "what should it do with that tracking UUID in its own data model?"

To me, a tidier solution might be to have a unique enterprise-fulfillment stream (or something like that) that makes the context clear to the LMS the type of event it is consuming without needing to pass extra data between services

Sort of thinking this through as I go. What do you think?

long74100: We took the conversation offline, updating adr to reflect the conversation.

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+auto-enrollment would just save learners clicks for this case.
+
+Questions to ponder (this is a draft so I'm putting it here for visibility for now):
+* Does enrolling a learner after a request approval provide a good/better user experience?
+  The current flow for request approval gives learners the flexibility to choose when to enroll in a course.
+  A user might request access to multiple courses with the intention of taking them one at a time or sometime later.
+  Automatically enrolling them in multiple courses might force learners to bookmark courses or even unenroll.
+* How valuable is the guarantee that a code is used to enroll a learner in the intended course?
+  A learner can use a code intended for one course to enroll in another, but it's an edge case since the email we send
+  after a request approval links them to the request course.
```
christopappas, line 34: nit: to the requested* coursed

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+We're attempting to adopt event-based communication between the enterprise-access service, the LMS (possibly), and the services
+responsible for granting/redeeming a subsidy.
+
+The following events will be sent to an event bus when the state of a subsidy request is approved:
+* `org.openedx.enterprise.access.coupon-code-request.approved.v1`
+* `org.openedx.enterprise.access.license-request.approved.v1``
+These events will be produced via Django signals that live in the `openedx-events <https://github.com/openedx/openedx-events>`__ repo.
+
+Auto-enrollment will be a feature that could be toggled off as it does not always make sense to auto-enroll a learner,
+e.g. when a coupon code offers 50% discount, a license request without course id, etc.. We will add a field `attempt_auto_enrollment`
```
christopappas, line 50: Question/thoughts for the team:

Case: customer has auto-enroll on. customer has 1 coupon. is 50% off. user requests course in coupon catalog... can admin even approve the request? also, does admin control toggle for whether or not they're on auto approval?

Case: customer has auto-enroll on. customer has 3 coupons. 2 coupons are 100%, one is 50%. a course exists in catalogs for 50% and 100% coupons. what's the UX on approval for the admin?

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+to the `SubsidyRequestCustomerConfiguration` model to allow for the auto-enrollment to be turned on/off.
+We could also explore having the auto-enrollment be a request specific rather than an enterprise specific toggle.
+
+The request approval process will involve these steps:
+1. An admin approves a subsidy request.
+2. The state on the subsidy request changes to `approved`.
+3. An subsidy request approved event is sent to the event bus.
+4. license-manager or ecommerce will consume this event, assigns a subsidy, and send a subsidy assigned event.
+   It then enrolls the learner (either by calling the LMS or sending an event for the LMS to consume)
```
christopappas, line 59: 5. ???

profit

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+
+Changes required
+----------------
+* enterprise-access will no longer call the services directly to grant a subsidy.
+* We will no longer send an email to the learner at the time of approval.
+* We will introduce a new subsidy request state `fulfilled` that signifies a subsidy has been granted.
+* We will modify the approval email to be a generic email saying that the learner has been approved for the course.
+  The generic email will contain a link to the course run that the learner requested, without mentioning enrollment.
+  When the learner clicks the link, they will be lead to the course about page where they can view their course.
+  It's important for this email to be generic because it allows enterprise-access to be decoupled from the enrollment process
```
christopappas, line 71: 🏆

christopappas (approved):
this is looking good to me, few nitty things and then just adding a few Qs for the group. overall though ✅

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+
+Context
+=======
+
+Learners are currently able to request a subsidy (license or coupon code) through the enterprise-access service.
+After a request is approved, the learner has to redeem the subsidy and enroll in a course manually.
+Since the goal of a learner is to access a course, we want to automatically enroll them after a request approval.
+
+For coupon code requests, requiring manual redemption is not ideal because we cannot enforce that a code is only used
+for the course it was requested for.
```
macdiesel, line 18: nit: for the course it was requested for. to: "for the course requested"

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+Admins have some visibility into the usage of a coupon code via the admin portal. It shows only if a coupon code has been redeemed,
```
macdiesel, line 19: "it shows only if" What is the it? The request? The code? Be more specific. Perhaps "The code only shows if it has been redeemed, but not the course for which it was redeemed." 

(FWIW I'm not a fan of dangling participles)

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+not the course it's was redeemed for. We have to query ecommerce data to check the order created using the coupon code
+for that information. Automatically enrolling a learner after a request approval would guarantee that a code is used to enroll in
+the intended course.
+
+For license requests, it's less of a problem because a license can be used for all courses in an enterprise catalog and
+auto-enrollment would just save learners clicks for this case.
+
+Questions to ponder (this is a draft so I'm putting it here for visibility for now):
+* Does enrolling a learner after a request approval provide a good/better user experience?
```
macdiesel, line 28: I had not thought of this but we I would like to draw Joe's attention to this one.

docs/decisions/0002-auto-enrollment-post-request-approval.rst
macdiesel: Clever solution

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+Auto-enrollment will be a feature that could be toggled off as it does not always make sense to auto-enroll a learner,
+e.g. when a coupon code offers 50% discount, a license request without course id, etc.. We will add a field `attempt_auto_enrollment`
+to the `SubsidyRequestCustomerConfiguration` model to allow for the auto-enrollment to be turned on/off.
+We could also explore having the auto-enrollment be a request specific rather than an enterprise specific toggle.
+
+The request approval process will involve these steps:
+1. An admin approves a subsidy request.
+2. The state on the subsidy request changes to `approved`.
+3. An subsidy request approved event is sent to the event bus.
+4. license-manager or ecommerce will consume this event, assigns a subsidy, and send a subsidy assigned event.
```
macdiesel, line 58: Would it be better to use fulfilled here instead of subsidy assignment? Help us align the language better.

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+
+enterprise-access still calls ecommerce and license-manager to assign a subsidy, then sends a request approved event
+to the event bus. The subsidy granting services will consume the events, redeem the granted subsidy,
+and enroll a learner in the course specified in the request.
+
+Changes required
+----------------
+* We will no longer send an email to the learner at the time of approval. Instead, we will send an email to
+  the learner when they have been enrolled.
+* enterprise-access will consume enrollment events from the LMS, check if the enrollment corresponds to an approved
```
macdiesel, line 135: Line 60 above states:
>  enterprise-access consumes the subsidy assigned event and changes the state on the subsidy request to `fulfilled`.
   It then sends an request approved email to the learner of the request. As far as enterprise-access knows, it's work is done at this point.

Will these events be consumed or is that out of scope?

docs/decisions/0002-auto-enrollment-post-request-approval.rst
```diff
@@ -0,0 +1,158 @@
...
+  request by matching on `lms_user_id`` and `course_id`, and send an approval email to the learner if a matching
+  request is found.
+* We might not need the above changes if we make the messaging in the approval email generic enough,
```
macdiesel, line 138: Need to shore these up with the decisions we make above?

macdiesel (approved):

docs/decisions/0002-auto-enrollment-post-request-approval.rst
long74100: $$$$$

PR #57, opened 2022-04-26
feat: retrieve course partner info upon save of subsidy request model

adamstankiewicz:
## Description

ENT-5601

* Adds `course_partners` JSON field to hold a list of partners for the course associated with the subsidy request.
* If `course_partners` is not set on `LicenseRequest` or `CouponCodeRequest` object, request it from course-discovery and store it in `course_partners`.
* Display `course_partners` as formatted list in Django Admin, if set.
* Include `course_partners` in the API response (see below).

## Example response
```json
{
	"count": 1,
	"next": null,
	"previous": null,
	"results": [{
		"uuid": "0229c579-f0aa-4df9-88fc-2f56c1fbff1b",
		"lms_user_id": 3,
		"email": "edx@example.com",
		"course_id": "edX+DemoX",
		"course_title": "Demonstration Course",
		"course_partners": [{
			"uuid": "61f2d9b4-5fdf-41df-ba87-0ae59e6eef99",
			"name": "edX"
		}],
		"enterprise_customer_uuid": "b6fd6057-4e66-4c6f-b747-05474acdf6a6",
		"state": "requested",
		"reviewed_at": null,
		"reviewer_lms_user_id": null,
		"decline_reason": null,
		"created": "2022-04-26T12:07:52.803041Z",
		"modified": "2022-04-26T12:07:52.803041Z",
		"subscription_plan_uuid": null,
		"license_uuid": null
	}],
	"num_pages": 1
}
```


enterprise_access/apps/subsidy_request/models.py
```diff
@@ -277,19 +290,20 @@ def _history_user(self, value):
...
 @receiver(models.signals.post_save, sender=CouponCodeRequest)
 @receiver(models.signals.post_save, sender=LicenseRequest)
-def update_course_title_for_subsidy_request(sender, **kwargs):  # pylint: disable=unused-argument
-    """ Post save hook to grab course_title from discovery service """
+def update_course_info_for_subsidy_request(sender, **kwargs):  # pylint: disable=unused-argument
+    """ Post save hook to grab course info from discovery service """
 
     subsidy_request = kwargs['instance']
-    if subsidy_request.course_title:
+    print('!!!', subsidy_request.course_title, subsidy_request.course_partners)
```
iloveagent57, line 297: nit: delete

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -53,6 +56,16 @@ class SubsidyRequest(TimeStampedModel, SoftDeletableModel):
...
+    course_partners = JSONField(
+        blank=True,
+        null=True,
+        load_kwargs={'object_pairs_hook': collections.OrderedDict},
+        dump_kwargs={'indent': 4, 'cls': JSONEncoder, 'separators': (',', ':')},
+        help_text=_(
+            "List of course partner dictionaries."
+        )
+    )
+
```
iloveagent57, line 68: The fact that we now have 1 field identifying a course and 2 fields holding metadata about a course, all on a model called `SubsidyRequest`, smells fishy.  It's outside the scope of this PR, but we should consider replicating the content metadata we need into a new model.

iloveagent57 (commented):
Curious to hear your thoughts on creating a new model to hold this metadata.

enterprise_access/apps/subsidy_request/models.py
adamstankiewicz: D'oh. Thanks :)

enterprise_access/apps/subsidy_request/models.py
long74100: I agree, those 2 fields don't really belong in this model :/. It does feel like a decent size scope increase though, we'd probably want to update the serializer to reflect the nested structure, update consumers, etc. Should this be a quick tech debt follow-up?.

adamstankiewicz: @iloveagent57 @long74100 I agree it feels out of scope for this particular PR at this point. Let's tackle it as a follow-up? Also, curious - do we have a similar situation like this in `CustomerAgreement` in license-manager since it's storing 1 field to identify a enterprise UUID and 2 fields holding metadata about the customer?

Is the difference that there's a 1:1 mapping between CustomerAgreement <-> EnterpriseCustomer vs. many SubsidyRequest <-> 1 course?

adamstankiewicz: Filed ticket to capture these thoughts: https://openedx.atlassian.net/browse/ENT-5776

long74100: I think we do have a similar situation in CustomerAgreement. The 1:1, 1:n mapping difference doesn't matter too much imo.

long74100 (approved):

PR #58, opened 2022-04-26
build: explicitly install librdkafka in Dockerfile

adamstankiewicz:
Necessary for Apple M1 Silicon chip support.

## Testing Instructions

```bash
make dev.down
make docker_build
make dev.up
make app-shell

# in the shell
make requirements
which pytest # this should return its installed path
```


johnnagro:
fyi similar to https://github.com/openedx/license-manager/pull/416

johnnagro (approved):

PR #64, opened 2022-05-20
fix: send campaign messages using recipients

long74100:
https://2u-internal.atlassian.net/browse/ENT-5854

Merge after https://github.com/openedx/edx-enterprise/pull/1566.

There were two scenarios we ran into that prevented emails from being sent:
1. Users without a braze profile, we failed to create one or import the user into the Braze ecosystem
2. Profiles without email, a profile with lms_user_id but not an email. Bizzare.

These changes guarantee that our emails will be sent regardless.


enterprise_access/apps/api_client/braze_client.py
```diff
@@ -28,3 +29,48 @@ def __init__(self):
             api_url=settings.BRAZE_API_URL,
             app_id=settings.BRAZE_APP_ID
         )
+
+    def generate_mailto_link(self, emails):
```
johnnagro, line 33: why does this need to be in the braze client?

johnnagro: [long] need CSV list, easier than template. braze lib better than random task method hanging out.

johnnagro (approved):
lgtm -

some assumptions we want to better understand, opportunities to make this stuff more awesome/easier to use:
- where/how can we use aliases before they're "created"? do they always need to exist?
- where/how can we add more calls `identify` to more thoroughly stamp out the duplicate profile issue?
- what kind of content template functions exist? to do things like create a CSV from an array

PR #70, opened 2022-07-14
feat: Add some additional logging to help sort out an email sending issue

macdiesel:


christopappas (approved):

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -136,8 +136,8 @@ def send_admins_email_with_new_requests_task(enterprise_customer_uuid):
             recipients=recipients,
             trigger_properties=braze_trigger_properties,
         )
-    except HTTPError as exc:
-        logger.exception(exc)
+    except:
```
christopappas, line 139: actually can you change this to `except HTTPError:`? or bare `except`s are not preferable

iloveagent57 (approved):

christopappas (changes requested):
just the bare except bit needs to be addressed, otherwise looks fine

enterprise_access/apps/subsidy_request/tasks.py
iloveagent57: I feel like it's permissible in a case where you're trying to debug and casting a wide net.  Maybe leave a comment with a super-duper-pinky-promise to un-bare-except when the issue is resolved?

christopappas (approved):

PR #72, opened 2022-07-19
add psg test file

christopappas:


enterprise_access/apps/subsidy_request/management/commands/python_study_group.py
```diff
@@ -0,0 +1,43 @@
...
+
+from enterprise_access.apps.subsidy_request.models import SubsidyRequestCustomerConfiguration
+from enterprise_access.apps.subsidy_request.tasks import psg_send_admins_email_with_new_requests
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    This command is intended to run as frequently as specified in a crontab. The window of time used
```
rgraber, line 17: A one-line summary at the top would be useful

johnnagro:
missing in the description a plain english explanation of what you're trying to accomplish and why

enterprise_access/apps/subsidy_request/management/commands/python_study_group.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    def handle(self, *args, **options):
+        configs = SubsidyRequestCustomerConfiguration.objects.filter(
+            subsidy_requests_enabled=True
+        )
+
+        for config in configs:
+            enterprise_customer_uuid = config.enterprise_customer_uuid
+            send_admins_email_with_new_requests_task.delay(enterprise_customer_uuid)
```
ashultz0, line 43: this actually isn't the imported method, which means this code has not been run, which would probably make me stop right there

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+    # "future" is greater than "past"
+    # so if created is greater than last remind date, it means
+    # it was created after cron was last run
+    if customer_config.last_remind_date is not None:
+        subsidy_requests = subsidy_requests.filter(created__gte=customer_config.last_remind_date)
+
+    subsidy_requests = subsidy_requests.order_by("-created")
+
+    if not subsidy_requests:
+        logger.info(
```
schenedx, line 55: I would log this a different format. Like `warning` or `error`?

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+def psg_send_admins_email_with_new_requests_task(enterprise_customer_uuid):
+    """
+    Task to send new-request emails to admins.
+
+    Args:
+        enterprise_customer_uuid (str): enterprise customer uuid identifier
+    Raises:
+        HTTPError if Braze client callfails with an HTTPError
+    """
+    lms_client = LmsApiClient()
```
nsprenkle, line 32: Is this something we should be instantiating every time or just once? Do we need any sort of handling in case the client was not set up correctly?

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+    enterprise_customer_data = lms_client.get_enterprise_customer_data(enterprise_customer_uuid)
+
+    config_model = apps.get_model('subsidy_request.SubsidyRequestCustomerConfiguration')
```
schenedx, line 35: Can this task work without getting the config model here? What if config model is no longer there?

enterprise_access/apps/subsidy_request/management/commands/python_study_group.py
```diff
@@ -0,0 +1,43 @@
...
+        'Spin off celery tasks to send enterprise admins an email that lists requests '
+        'that were created since this was last run.'
+    )
+
+    def handle(self, *args, **options):
+        configs = SubsidyRequestCustomerConfiguration.objects.filter(
+            subsidy_requests_enabled=True
+        )
+
+        for config in configs:
```
rgraber, line 41: Why not just pass the whole config object since we need it later anyway?

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+    customer_config = config_model.objects.get(
```
nsprenkle, line 36: Ever a case where this would find not 1 result, where `get` would fail? Need added handling.

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+    ]
+    try:
+        braze_client.send_campaign_message(
+            settings.BRAZE_NEW_REQUESTS_NOTIFICATION_CAMPAIGN,
+            recipients=recipients,
+            trigger_properties=braze_trigger_properties,
+        )
+    except HTTPError as exc:
+        logger.exception(exc)
+        raise
```
schenedx, line 105: When we raise exception here, what happens? Would it affect other tasks?

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+    # so if created is greater than last remind date, it means
+    # it was created after cron was last run
+    if customer_config.last_remind_date is not None:
+        subsidy_requests = subsidy_requests.filter(created__gte=customer_config.last_remind_date)
+
+    subsidy_requests = subsidy_requests.order_by("-created")
+
+    if not subsidy_requests:
+        logger.info(
+            'No new subsidy requests. Not sending new requests '
```
nedbat, line 56: It's better to add strings together than use implicit concatenation.

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+
+@shared_task(base=LoggedTaskWithRetry)
+def psg_send_admins_email_with_new_requests_task(enterprise_customer_uuid):
+    """
+    Task to send new-request emails to admins.
+
+    Args:
+        enterprise_customer_uuid (str): enterprise customer uuid identifier
+    Raises:
+        HTTPError if Braze client callfails with an HTTPError
```
dianekaplan, line 30:
```suggestion
        HTTPError if Braze client call fails with an HTTPError
```

rgraber, line 30:
```suggestion
        HTTPError if Braze client call fails with an HTTPError
```

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+from enterprise_access.apps.api_client.discovery_client import DiscoveryApiClient
+from enterprise_access.apps.api_client.lms_client import LmsApiClient
+from enterprise_access.apps.subsidy_request.constants import SubsidyRequestStates
+from enterprise_access.tasks import LoggedTaskWithRetry
+from enterprise_access.utils import get_subsidy_model
+
+logger = logging.getLogger(__name__)
+
+@shared_task(base=LoggedTaskWithRetry)
+def psg_send_admins_email_with_new_requests_task(enterprise_customer_uuid):
```
MichaelRoytman, line 23: Can we make this task work for a set of `SubsidyRequestCustomerConfiguration` as opposed to just one? Running this on each `SubsidyRequestCustomerConfiguration` could be wasteful, and a bulk task would be more efficient.

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+    # Filter when we last run this unless we never ran before
+    # "future" is greater than "past"
+    # so if created is greater than last remind date, it means
+    # it was created after cron was last run
+    if customer_config.last_remind_date is not None:
+        subsidy_requests = subsidy_requests.filter(created__gte=customer_config.last_remind_date)
+
+    subsidy_requests = subsidy_requests.order_by("-created")
+
+    if not subsidy_requests:
```
nsprenkle, line 54: I think this will always be `False` since it's a query object. Test based on count?

enterprise_access/apps/subsidy_request/management/commands/python_study_group.py
nedbat: Especially since the first sentence here is not a summary at all.

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+@shared_task(base=LoggedTaskWithRetry)
+def psg_send_admins_email_with_new_requests_task(enterprise_customer_uuid):
+    """
+    Task to send new-request emails to admins.
+
+    Args:
+        enterprise_customer_uuid (str): enterprise customer uuid identifier
+    Raises:
+        HTTPError if Braze client callfails with an HTTPError
+    """
```
ashultz0, line 31: this method is just a long stream of consciouness and I'd love to have it broken up, but sometimes not worth it for this sort of task

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+
+    braze_trigger_properties = {}
+    enterprise_slug = enterprise_customer_data['slug']
+
+    if subsidy_model == apps.get_model('subsidy_request.LicenseRequest'):
+        subsidy_string = 'subscriptions'
+    else:
+        subsidy_string = 'coupons'
+
+    url = f'{settings.ENTERPRISE_ADMIN_PORTAL_URL}/{enterprise_slug}/admin/{subsidy_string}/manage-requests'
```
johnnagro, line 69: `urljoin` is safer than string concatenation here

schenedx (approved):
Left some comments inline. Nothing major. Looks pretty good.

enterprise_access/apps/subsidy_request/management/commands/python_study_group.py
```diff
@@ -0,0 +1,43 @@
...
+    can simply give config models a field that holds how frequently admins want to be notified
+    and then update the cronjob to run more frequently (a "higher resolution") to let us
+    include all use cases.
+    """
+    help = (
+        'Spin off celery tasks to send enterprise admins an email that lists requests '
+        'that were created since this was last run.'
+    )
+
+    def handle(self, *args, **options):
```
MichaelRoytman, line 36: Many management commands have arguments for sleep time, batch size, etc. What is the expected use of this command? Would those arguments be valuable for the anticipated number of `SubsidyRequestCustomerConfiguration`?

enterprise_access/apps/subsidy_request/psg_tasks.py
nedbat, line 30: "callfails" isn't a word :)

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+
+    if not subsidy_requests:
+        logger.info(
+            'No new subsidy requests. Not sending new requests '
+            f'email to admins for enterprise {enterprise_customer_uuid}.'
+            )
+        return
+
+    braze_trigger_properties = {}
+    enterprise_slug = enterprise_customer_data['slug']
```
nsprenkle, line 62: Ever a case where this data would not exist?

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+        state=SubsidyRequestStates.REQUESTED,
+    )
+    # Filter when we last run this unless we never ran before
+    # "future" is greater than "past"
+    # so if created is greater than last remind date, it means
+    # it was created after cron was last run
+    if customer_config.last_remind_date is not None:
+        subsidy_requests = subsidy_requests.filter(created__gte=customer_config.last_remind_date)
+
+    subsidy_requests = subsidy_requests.order_by("-created")
```
ashultz0, line 52: subsidy request set scattered and confusing

enterprise_access/apps/subsidy_request/psg_tasks.py
MichaelRoytman, line 32: The body of this task is really long. Could you break this out into meaningful methods so it's easier to understand?

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+        enterprise_customer_uuid=enterprise_customer_uuid,
+    )
+
+    subsidy_model = get_subsidy_model(customer_config.subsidy_type)
+    subsidy_requests = subsidy_model.objects.filter(
+        enterprise_customer_uuid=enterprise_customer_uuid,
+        state=SubsidyRequestStates.REQUESTED,
+    )
+    # Filter when we last run this unless we never ran before
+    # "future" is greater than "past"
```
nedbat, line 46: With no period after "before", i had trouble parsing this text.

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+            )
+        return
+
+    braze_trigger_properties = {}
+    enterprise_slug = enterprise_customer_data['slug']
+
+    if subsidy_model == apps.get_model('subsidy_request.LicenseRequest'):
+        subsidy_string = 'subscriptions'
+    else:
+        subsidy_string = 'coupons'
```
ashultz0, line 67: this whole if is a wtf

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+        logger.info(
+            'No new subsidy requests. Not sending new requests '
+            f'email to admins for enterprise {enterprise_customer_uuid}.'
+            )
+        return
+
+    braze_trigger_properties = {}
+    enterprise_slug = enterprise_customer_data['slug']
+
+    if subsidy_model == apps.get_model('subsidy_request.LicenseRequest'):
```
nsprenkle, line 64: 'I do not think [this code] means what you think it means.' ~ Inigo Montoya

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+        course_title = subsidy_request.course_title
+
+        braze_trigger_properties['requests'].append({
+            'user_email': user_email,
+            'course_title': course_title,
+        })
+
+    admin_users = enterprise_customer_data['admin_users']
+
+    logger.info(
```
ashultz0, line 85: this would be a logical place to break the method

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+
+        user_email = subsidy_request.user.email
+        course_title = subsidy_request.course_title
+
+        braze_trigger_properties['requests'].append({
+            'user_email': user_email,
+            'course_title': course_title,
+        })
+
+    admin_users = enterprise_customer_data['admin_users']
```
nsprenkle, line 83: Again, case where this key might not exist?

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+        for admin_user in admin_users
+    ]
+    try:
+        braze_client.send_campaign_message(
+            settings.BRAZE_NEW_REQUESTS_NOTIFICATION_CAMPAIGN,
+            recipients=recipients,
+            trigger_properties=braze_trigger_properties,
+        )
+    except HTTPError as exc:
+        logger.exception(exc)
```
rgraber, line 104: Don't need to pass `exc` here if you're using logger.exception. Also would be useful to add a message for clarity/grepping Splunk logs.

enterprise_access/apps/subsidy_request/psg_tasks.py
nedbat, line 64: It seems odd to check the type of the model instead of the value of `customer_config.subsidy_type`.

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+
+    url = f'{settings.ENTERPRISE_ADMIN_PORTAL_URL}/{enterprise_slug}/admin/{subsidy_string}/manage-requests'
+    braze_trigger_properties['manage_requests_url'] = url
+
+    braze_trigger_properties['requests'] = []
+    for subsidy_request in subsidy_requests:
+
+        user_email = subsidy_request.user.email
+        course_title = subsidy_request.course_title
+
```
nedbat, line 77: I wouldn't assign these to two variables just to use them once a few lines lower down.

enterprise_access/apps/subsidy_request/psg_tasks.py
rgraber, line 46: I don't understand this comment. Why are future and past in quotes?

enterprise_access/apps/subsidy_request/psg_tasks.py
MichaelRoytman, line 54: I _think_ you want to use `exists` here. I'm not sure if this will evaluate correctly (i.e. is the QuerySet evaluated by such a statement?). (If I had more time, I'd just look this up myself.)

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+            'course_title': course_title,
+        })
+
+    admin_users = enterprise_customer_data['admin_users']
+
+    logger.info(
+        f'Sending new-requests email to admins for enterprise {enterprise_customer_uuid}. '
+        f'The email includes {len(subsidy_requests)} subsidy requests.'
+    )
+    braze_client = BrazeApiClient()
```
nsprenkle, line 89: Similar question, is there any issue here with calling this every time we run this management command? Any handling around the API client not returning/setting up correctly?

rgraber (changes requested):
Task should be unit tested

nsprenkle (commented):
PSG Testing 🐍

enterprise_access/apps/subsidy_request/management/commands/python_study_group.py
MichaelRoytman, line 17: I appreciate the detail in this comment.

enterprise_access/apps/subsidy_request/psg_tasks.py
ashultz0: probably one function to get the appropriate subsidy request and turn them into trigger objects, one to send the mail, and a frame func around it

enterprise_access/apps/subsidy_request/psg_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+        braze_client.send_campaign_message(
+            settings.BRAZE_NEW_REQUESTS_NOTIFICATION_CAMPAIGN,
+            recipients=recipients,
+            trigger_properties=braze_trigger_properties,
+        )
+    except HTTPError as exc:
+        logger.exception(exc)
+        raise
+
+    customer_config.last_remind_date = datetime.now()
```
nedbat, line 107: Do you have a window here where a new request could appear after this function started, but before you capture now(), and so a reminder is missed?

PR #84, opened 2022-10-25
docs: Remove "event bus" from TODO comments

timmc-edx:
This is no longer planned as part of the event bus work, although it's probably still a good idea.


robrap (commented):
Here are my pending comments.

robrap (approved):
@timmc-edx: I say we land this PR so we can close our ticket. 

Separately, @macdiesel and @christopappas can decide when and if to do the full revert of https://github.com/edx/enterprise-access/pull/51, since they aren't using any of this. We didn't actually implement this, so it doesn't really make sense for us to be responsible for the clean up.

PR #91, opened 2023-01-23
feat: ENT-6692 - Added app and models for subsidy access policy

irfanuddinahmad:
Description: This PR adds the Subsidy Access django application

JIRA: [ENT-6692](https://2u-internal.atlassian.net/browse/ENT-6692)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -0,0 +1,192 @@
...
+        default=0,
+    )
+
+    def can_redeem(self, learner_id, content_key):
+        spent_amount = subsidy_client.amount_spent_for_learner(
+            subsidy_uuid = self.subsidy_uuid,
+            learner_uuid = learner_id,
+            )
+        #TODO the following should also incorporate the course price
+        if spent_amount < self.per_learner_spend_limit:
```
irfanuddinahmad, line 169: @iloveagent57 This should account for the price of the course in the transaction ... right? Price to be fetched from which client?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -0,0 +1,192 @@
...
+        default=0,
+    )
+
+    def can_redeem(self, learner_id, content_key):
+            group_amount_spent = subsidy_client.amount_spent_for_group_and_catalog(
+                subsidy_uuid = self.subsidy_uuid,
+                group_uuid = self.group_uuid,
+                catalog_uuid = self.catalog_uuid,
+                )
+            if  group_amount_spent < self.spend_limit:
```
irfanuddinahmad, line 189: @iloveagent57 Course price here as well ... right?

enterprise_access/apps/subsidy_access_policy/models.py
iloveagent57: Yup, exactly.  This should be fetched from the Subsidy API/client.  I'll try to get you a real API specification soon so that this logic can be filled out a little more.

enterprise_access/apps/subsidy_access_policy/models.py
iloveagent57: Yup, the amount previously spent here is the aggregate of all transaction amounts for the "group" (which in v1 of this project, is just the whole enterprise).

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -0,0 +1,192 @@
...
+        return None
+
+    def has_redeemed(self, learner_id, content_key):
+        """
+        Check if the subsidy has been redeemed.
+        """
+        if self.access_method == AccessMethods.DIRECT:
+            return subsidy_client.has_redeemed(self.subsidy_uuid, learner_id, content_key)
+        elif self.access_method == AccessMethods.REQUEST:
+            return subsidy_client.has_requested(self.subsidy_uuid, learner_id, content_key)
```
iloveagent57, line 76: Note: the prototype code is probably wrong about this branch.  If the access method is browse & request, it should really query the enterprise-access B&R API.  We may need to update the models/API for B&R to accept a subsidy uuid as a filter condition.

enterprise_access/apps/subsidy_access_policy/mocks.py
```diff
@@ -0,0 +1,81 @@
...
+        """Subsidy service api"""
+        return mock.MagicMock(subsidy_uuid, learner_id, content_key)
+
+    @classmethod
+    def has_requested(cls, subsidy_uuid, learner_id, content_key):
+        """Subsidy service api"""
+        return mock.MagicMock(subsidy_uuid, learner_id, content_key)
+
+    @classmethod
+    def get_license_for_learner(cls, subsidy_uuid, learner_id):
```
saleem-latif, line 59: nit: Please rename `license` to `subsidy` in this PR.

enterprise_access/apps/subsidy_access_policy/mocks.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    API client for the groups service.
+    """
+    @classmethod
+    def is_redeemable(cls, subsidy_uuid, learner_id, content_key):
+        """Subsidy service api"""
+        return mock.MagicMock(subsidy_uuid, learner_id, content_key)
+
+    @classmethod
+    def redeem(cls, subsidy_uuid, learner_id, content_key):
```
saleem-latif, line 39: Do we want to change this to `create_transaction`? @iloveagent57 
So that is more inline with https://github.com/openedx/enterprise-subsidy/blob/180b1b3dd9e7dddb5c9c7f6175b424bae685a3a5/docs/decisions/0002-initial-api-specification.rst#post-enterprise-subsidy-transaction

saleem-latif (changes requested):

enterprise_access/apps/subsidy_access_policy/mocks.py
irfanuddinahmad: @iloveagent57 should we change this name?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -0,0 +1,201 @@
...
+
+    uuid = models.UUIDField(
+        primary_key=True,
+        default=uuid4,
+        editable=False,
+        unique=True,
+    )
+    version = models.IntegerField(default=0)
+    active = models.BooleanField(default=False)
+    group_uuid = models.UUIDField(null=True, blank=True, db_index=True)
```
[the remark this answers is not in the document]
irfanuddinahmad: There could be a case where the group and catalog are not attached to the subsidy

enterprise_access/apps/subsidy_access_policy/mocks.py
irfanuddinahmad: @moconnell1453 @iloveagent57 Can we create the [access policy repo](https://github.com/openedx/edx-enterprise-subsidy-access-policy) also?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+        abstract = True
+
+    uuid = models.UUIDField(
+        primary_key=True,
+        default=uuid4,
+        editable=False,
+        unique=True,
+    )
+    version = models.IntegerField(default=0)
```
pwnage101, line 33: I *think* such a version field would overlap in purpose with django simple history, which I think I'd rather see instead of an explicit version field.

enterprise_access/apps/subsidy_access_policy/mocks.py
irfanuddinahmad: No

iloveagent57 (approved):

PR #100, opened 2023-02-08
feat: access policy list and redeem api endpoints

muhammad-ammar:


docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+       'quantity': 19900,
+       'unit': 'USD_CENTS',
+       'reference_id': 1234,
+       'reference_table': 'enrollments',
+       'subsidy_access_policy_uuid': 'a-policy-uuid',
+       'metadata': {...},
+       'created': 'created-datetime',
+       'modified': 'modified-datetime',
+       'reversals': []
+   }
```
muhammad-ammar, line 66: This is not finalized.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -32,7 +38,7 @@ class Meta:
     )
     active = models.BooleanField(default=False)
     group_uuid = models.UUIDField(null=True, blank=True, db_index=True)
```
muhammad-ammar, line 40: I believe `null` should be set to `False` because for now we will always receive an `enterprise_customer_uuid` and we will treating that as `group_id`

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Enterprise Access Policy Transactions
+**************************************
+
+POST access policy redeem transaction
+======================================
+**/api/v1/policy/redeem/**
+
+Creats a new transaction to redeem an entitlement by making a request to ``POST /api/v1/subsidies/.../transactions/``
+A side-effect of a successful POST request here is the creation of a course enrollment or entitlement
```
irfanuddinahmad, line 35: Lets not use the word 'entitlement'

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+that "fulfills" the ledger transaction.
+
+Inputs
+------
+
+- ``customer-id`` (POST data, required): The uuid of the customer. For now it will be an enterprise customer uuid.
```
irfanuddinahmad, line 41: Should we call this group-id a.k.a enterprise-uuid

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+Outputs
+-------
+Returns data about the transaction.
+
+::
+
+   {
+       'uuid': 'the-transaction-uuid',
+       'status': 'completed',
+       'idempotency_key': 'the-idempotency-key',
```
irfanuddinahmad, line 54: I think the transaction uuid will be unique and can be used for duplicate check

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+       'content_key': 'demox_1234+2T2023',
+       'quantity': 19900,
+       'unit': 'USD_CENTS',
+       'reference_id': 1234,
+       'reference_table': 'enrollments',
+       'subsidy_access_policy_uuid': 'a-policy-uuid',
+       'metadata': {...},
+       'created': 'created-datetime',
+       'modified': 'modified-datetime',
+       'reversals': []
```
irfanuddinahmad, line 65: reversals is a list of what kind?

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+       'status': 'completed',
+       'idempotency_key': 'the-idempotency-key',
+       'learner_id': 54321,
+       'content_key': 'demox_1234+2T2023',
+       'quantity': 19900,
+       'unit': 'USD_CENTS',
+       'reference_id': 1234,
+       'reference_table': 'enrollments',
+       'subsidy_access_policy_uuid': 'a-policy-uuid',
+       'metadata': {...},
```
irfanuddinahmad, line 62: metadata contains which data?

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+       'created': 'created-datetime',
+       'modified': 'modified-datetime',
+       'reversals': []
+   }
+
+Permissions
+-----------
+
+enterprise_learner
+  Should only create the transaction if the requesting user has implicit (JWT) or explicit (DB-defined)
```
irfanuddinahmad, line 72: Do we plan DB defined permissions?

enterprise_access/apps/api_client/enterprise_catalog_client.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        Check whether the specified enterprise catalog contains the given content.
+
+        Arguments:
+            catalog_uuid (UUID): UUID of the enterprise catalog to check.
+            content_ids (list of str): List of content ids to check whether the catalog contains. The endpoint does not
+                differentiate between course_run_ids and program_uuids so they can be used interchangeably. 
+
+        Returns:
+            bool: Whether the given content_ids were found in the specified enterprise catalog.
```
irfanuddinahmad, line 27: it should return a bool with each of the content_ids ... no?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -54,9 +60,12 @@ def can_redeem(self, learner_id, content_key):
         """
         Check that a given learner can redeem the given content.
         """
-        if not catalog_client.catalog_contains_content(self.catalog_uuid, content_key):
+        enterprise_catalog_api_client = EnterpriseCatalogApiClient()
+        lms_api_client = LmsApiClient()
+
+        if not enterprise_catalog_api_client.contains_content_items(self.catalog_uuid, [content_key]):
```
irfanuddinahmad, line 66: We will always have one content_key here ... no?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -87,6 +95,46 @@ def has_redeemed(self, learner_id, content_key):
...
+        # ```
+
+        sorted_policies = sorted(
+            redeemable_policies,
+            # p.policy_type_priority is a class property that's a lower number for
+            # LearnerCreditSubsidyAccessPolicy and higher for SubscriptionSubsidyAccessPolicy.
+            key=lambda p: (p.priority, p.subsidy.balance),
+        )
+        # Simply pick the first policy:
+        return sorted_policies[0]
```
irfanuddinahmad, line 136: add a TODO here

enterprise_access/apps/subsidy_access_policy/views.py
```diff
@@ -0,0 +1,41 @@
...
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+
+
+class AccessPolicyAPIView(APIView):
+    """
+    """
+    authentication_classes = [JwtAuthentication]
+    permission_classes = [permissions.IsAuthenticated]
+
+    def post(self, request, format=None):
```
irfanuddinahmad, line 16: this is the redeem url .. right?

enterprise_access/apps/api_client/enterprise_catalog_client.py
muhammad-ammar: enterprise catalog endpoint only gives a single bool against all the content_ids we pass.

enterprise_access/apps/subsidy_access_policy/views.py
muhammad-ammar: yes

docs/decisions/0003-initial-api-specification.rst
muhammad-ammar: The response is not finalized.

docs/decisions/0003-initial-api-specification.rst
muhammad-ammar: The response is not finalized.

docs/decisions/0003-initial-api-specification.rst
muhammad-ammar: The response is not finalized.

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+
+
+Enterprise Access Policy Transactions
+**************************************
+
+POST access policy redeem transaction
+======================================
+**/api/v1/policy/redeem/**
+
+Creats a new transaction to redeem an entitlement by making a request to ``POST /api/v1/subsidies/.../transactions/``
```
iloveagent57, line 34: "...making a request to the enterprise-subsidy ``POST /api/v1/subsidies/.../transactions/`` endpoint."

docs/decisions/0003-initial-api-specification.rst
iloveagent57: You could say something like "redeem subsidy value by making a request..."

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+**/api/v1/policy/**
+The root URL for all actions allowed on access policy apis.
+
+
+Enterprise Access Policy Transactions
+**************************************
+
+POST access policy redeem transaction
+======================================
+**/api/v1/policy/redeem/**
```
iloveagent57, line 32: This route should probably have the policy uuid as part of the path. `/api/v1/policy/[policy_uuid]/redeem`

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Creats a new transaction to redeem an entitlement by making a request to ``POST /api/v1/subsidies/.../transactions/``
+A side-effect of a successful POST request here is the creation of a course enrollment or entitlement
+that "fulfills" the ledger transaction.
+
+Inputs
+------
+
+- ``group_id`` (POST data, required): The uuid of the customer. For now it will be an enterprise customer uuid.
```
iloveagent57, line 41: I think we can leave `group_id` out for now.  The customer uuid can be inferred from the `policy_uuid` in the route.

enterprise_access/apps/api/v1/views.py
```diff
@@ -740,3 +748,58 @@ def partial_update(self, request, *args, **kwargs):
...
+class AccessPolicyRedeemAPIView(APIView):
+    """
+    API to redeem a policy.
+    """
+    authentication_classes = [JwtAuthentication]
+    permission_classes = [permissions.IsAuthenticated]
+
+    @permission_required(
+        constants.REQUESTS_ADMIN_LEARNER_ACCESS_PERMISSION,
+        fn=lambda request: request.data.get('group_id')
```
iloveagent57, line 762: If the route for this view contains the policy uuid, this permission_required `fn` would need to change.  You could use something like what license-manager does for a similar use-case: https://github.com/openedx/license-manager/blob/b4e538da2bcd66bbf2638d0dfcaa2467e0a27a7d/license_manager/apps/api/v1/views.py#L1359-L1362
, which basically means creating a utility/helper function to trace through related objects to get the relevant enterprise customer uuid: https://github.com/openedx/license-manager/blob/b4e538da2bcd66bbf2638d0dfcaa2467e0a27a7d/license_manager/apps/api/utils.py#L100-L132

docs/decisions/0003-initial-api-specification.rst
iloveagent57: After thinking through this a little bit, one tricky thing is that there are multiple types of policies encapsulated by different models, e.g. `PerLearnerEnrollmentCapLearnerCreditAccessPolicy` vs. `CappedEnrollmentLearnerCreditAccessPolicy`, and then whatever ones we'll eventually have for licenses. This could make it a little complicated to have a single route like `/api/v1/policy` without knowing exactly which model a view or viewset would need to lookup the PK in.  I'll start a conversation about this.

iloveagent57 (commented):
Here's some comments for now - the last comment exposed kind of a big question that I'll start a conversation about.

docs/decisions/0003-initial-api-specification.rst
iloveagent57: OK, John posted a link/ideas in Slack about trying to fix this up with Proxy models.  To go along with that, here's what I think a more detailed API spec looks like:

The primary rest API endpoint is `redeem` which can accept a GET or a POST.

### Create the redemption

`POST /api/v1/policy/[policy_uuid]/redeem`

This is is scoped to a policy identifier because it saves the computation that likely just happened in preceding GET redeem call (which has to compute against the subsidy and catalog APIs, which is network calls and real work), and also it makes future features of picking amongst multiple policies in a UI more feasible.

Successful POSTs should return an object in the response data containing some identifier indicating that enrollment/entitlement occurred and a URL that the learner-portal interstitial page should redirect to.  We still need to work out the exact implementation details of this.

### Checking redeemability
`GET /api/v1/policy/redeem?enterprise_customer_uuid=[customer_uuid]`

1. Should return a list of policies which could redeem for the given (user, content) pair in the request data.
2. If the (user, content) pair is not redeemable in any policy for the customer, endpoint should return an empty list.
3. Each item in the response results list is the serialized policy, containing at least the course key, access method, catalog id, and priority/preference of the policy.
4. Each item in the list will contain a field indicating the priority/preference of that item.  This makes it clear/explicit from the client perspective how the priority is being chosen.  The results list should be ordered by the values specified in this field.

Why not return a single object or null?  A list is the most generic thing to return, it supports use cases where clients need to know about all redeemable policies out-of-the-box.  It’s also more naturally RESTful (it aligns tightly to the actual data relationships between DB records).

irfanuddinahmad (approved):
LGTM

PR #109, opened 2023-03-06
feat: ENT-6842 added subsidy locks to the policy api

irfanuddinahmad:


enterprise_access/apps/api/v1/views.py
```diff
@@ -816,11 +820,23 @@ def redeem(self, request, *args, **kwargs):
...
+            lock_acquired = acquire_subsidy_lock(
+                policy.subsidy_uuid,
+                django_cache_timeout=SUBSIDY_LOCK_TIMEOUT_SECONDS,
+            )
+            if not lock_acquired:
+                return Response(
+                    data='Enrollment currently locked for this subsidy.',
+                    status=status.HTTP_423_LOCKED,
+                )
+            if policy.can_redeem(learner_id, content_key):
```
irfanuddinahmad, line 833: @iloveagent57 [copying your code](https://github.com/openedx/license-manager/blob/f901ea018e8525170e887ff46abed03cd257cebe/license_manager/apps/api/v1/views.py#L742) from license manager. Should there be a lock_id that is sent across to the subsidy api so that it can verify that the calling client is indeed the one holding the lock?

irfanuddinahmad: Also, the lock_id might need to be sent in such a manner that the receiving api can check that it is not spoofed

irfanuddinahmad: The lock_id is essentially the cross-service and unique transaction identifier

irfanuddinahmad: should we retry to acquire the lock if we fail to acquire it ... what should be the retry algorithm ... exponential backoff?

irfanuddinahmad:
> Irfan may I suggest an ADR to explain the locking mechanism we discussed this morning to go along with this PR?

@macdiesel added the ADR

docs/decisions/0004-add-access-policy-functionality.rst
```diff
@@ -270,6 +270,38 @@ then after that prioritizing subsidies with smaller balances.  Here’s a protot
      # Simply pick the first policy:
      return sorted_policies[0]
 
+Concurrency and Transaction Isolation
```
muhammad-ammar, line 273: Do we need to update this section(text + code) or I am missing something? Here we are describing multiple locking constraints. But in the actual implementation we are just locking on `policy uuid`.

enterprise_access/apps/api/v1/views.py
```diff
@@ -816,11 +816,20 @@ def redeem(self, request, *args, **kwargs):
 
         learner_id = serializer.data['learner_id']
         content_key = serializer.data['content_key']
+        try:
+            lock_acquired = policy.acquire_lock(learner_id, content_key)
```
muhammad-ammar, line 820: `acquire_lock` takes no argument.

enterprise_access/apps/api/v1/views.py
```diff
@@ -816,11 +816,20 @@ def redeem(self, request, *args, **kwargs):
...
-            response,
-            status=status.HTTP_200_OK,
-        )
+            return Response(
+                response,
+                status=status.HTTP_200_OK,
+            )
+        finally:
+            if lock_acquired:
+                policy.release_lock(learner_id, content_key)
```
muhammad-ammar, line 835: `release_lock` takes not argument

muhammad-ammar (changes requested):
Overall looks good. Left some minor comments.

docs/decisions/0004-add-access-policy-functionality.rst
irfanuddinahmad: added learner and content ids to the locking calls ... to be used as needed by the proxy models

macdiesel (approved):

PR #111, opened 2023-03-07
docs: ADR about locking design for access policies.

iloveagent57:


docs/decisions/0005-access-policy-locks.rst
```diff
@@ -0,0 +1,72 @@
...
+  which originate from the policy.
+
+During redemption, we'll acquire a lock key that is scoped to the ``uuid`` of the policy through
+which redemption occurs.  This means that no other, concurrent request can redeem through this policy
+while this lock is held.  This is the simplest implementation of a lock key choice we can make
+that covers all the access policy types described above.
+
+Consequences
+************
+
```
adamstankiewicz, line 55: [clarification] This locking implies that several learners from the same enterprise customer attempting to redeem content using the same policy concurrently won't be possible. In practice, will this lock duration be relatively instantaneous to not block other learners from attempting to enroll via the same policy at the same time as you?

Is the client (MFE) or the API responsible for retrying if a redemption request comes in when the policy is already locked?

iloveagent57: Changed as follows to answer these questions:
> Locking the entire policy when the policy only applies a per-learner spend cap is fairly restrictive.
  It implies that several learners from the same enterprise customer won't be able to
  concurrently redeem via the same policy.  In practice, this should be a relatively rare occurrence;
  the system throughput is not expected to be particularly high, although the lock duration could
  be on the order of seconds. In this scenario, the client would have to be responsible for any
  retry or communication to the user that a lock could not be acquired.
  We can eventually (or perhaps, immediately) choose to override the lock key choice in the implementing class of 
  that policy type to take the learner identifier into account, which would mitigate the negative effects
  in this scenario.

adamstankiewicz: 👨‍🍳 💋

adamstankiewicz (approved):
LGTM!

PR #112, opened 2023-03-09
feat: Added CRUD API for subsidy policies

sameenfatima78:
**GET list with group_uuid**
Screen Shot 2023-03-13 at 1 23 55 PM

**GET list with group_uuid validation and POST form**
Screen Shot 2023-03-13 at 1 24 04 PM

**Retrive/Delete/Patch**
Screen Shot 2023-03-13 at 1 22 02 PM

### DIFF
--- a/docs/decisions/0003-initial-api-specification.rst
+++ b/docs/decisions/0003-initial-api-specification.rst
@@ -66,9 +66,9 @@ A sample response can be seen below.
 Permissions
 -----------
 
-enterprise_learner
+enterprise_learner or enterprise_admin
   Should only create the transaction if the requesting user has implicit (JWT) or explicit (DB-defined)
-  ``enterprise_learner`` role assigned.
+  ``enterprise_learner`` or ``enterprise_admin`` role assigned.
 
 POST access policy redeem transaction
 ======================================
@@ -93,3 +93,145 @@ Permissions
 enterprise_learner
   Should only create the transaction if the requesting user has implicit (JWT) or explicit (DB-defined)
   ``enterprise_learner`` role assigned.
+
+
+CRUD operations on SubsidyAccessPolicy
+======================================
+
+GET **/api/v1/admin/policy/**
+
+List all policies linked to a group.
+
+Inputs
+------
+
+- ``group_uuid`` (Query Param, required): The uuid of the customer. For now it will be an enterprise customer uuid.
+
+Outputs
+-------
+A sample response can be seen below.
+
+::
+
+   [
+        {
+            "uuid": "d0cd25f9-0b73-49c3-8299-dc9751a12ef5",
+            "policy_type": "PerLearnerEnrollmentCreditAccessPolicy",
+            "description": "sdad",
+            "active": true,
+            "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
+            "catalog_uuid": "214cf999-5964-4a2e-afa8-c62461558315",
+            "subsidy_uuid": "214cf999-5964-4a2e-afa8-c62461342198",
+            "access_method": "direct",
+            "per_learner_enrollment_limit": 0,
+            "per_learner_spend_limit": 0,
+            "spend_limit": 5
+        },
+        {
+            "uuid": "a16b960d-0ddd-4af2-a596-9991cd5508da",
+            "policy_type": "CappedEnrollmentLearnerCreditAccessPolicy",
+            "description": "sdad",
+            "active": true,
+            "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
+            "catalog_uuid": "214cf999-5964-4a2e-afa8-c62461558315",
+            "subsidy_uuid": "214cf999-5964-4a2e-afa8-c62461342198",
+            "access_method": "direct",
+            "per_learner_enrollment_limit": 3,
+            "per_learner_spend_limit": 0,
+            "spend_limit": 0
+        },
+]
+
+Permissions
+-----------
+
+enterprise_admin
+  Should only create the transaction if the requesting user has implicit (JWT) or explicit (DB-defined) ``enterprise_admin`` role assigned.
+
+
+**GET /api/v1/admin/policy/<policy_uuid>/**
+
+Retrieve a subsidy access policy instance.
+
+Inputs
+------
+
+- ``policy_uuid`` (URL, required): The uuid of the customer. For now it will be an enterprise customer uuid.
+
+Outputs
+-------
+A sample response can be seen below.
+
+::
+
+    {
+        "uuid": "d0cd25f9-0b73-49c3-8299-dc9751a12ef5",
+        "policy_type": "PerLearnerEnrollmentCreditAccessPolicy",
+        "description": "sdad",
+        "active": true,
+        "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
+        "catalog_uuid": "214cf999-5964-4a2e-afa8-c62461558315",
+        "subsidy_uuid": "214cf999-5964-4a2e-afa8-c62461342198",
+        "access_method": "direct",
+        "per_learner_enrollment_limit": 3,
+        "per_learner_spend_limit": 0,
+        "spend_limit": 0
+    }
+
+Permissions
+-----------
+
+enterprise_admin
+  Should only create the transaction if the requesting user has implicit (JWT) or explicit (DB-defined) ``enterprise_admin`` role assigned.
+
+**POST /api/v1/admin/policy/**
+
+Create a subsidy access policy instance after validating the request data.
+
+Inputs
+------
+
+- ``payload`` (request data, required): Payload data for POST request.
+
+A sample **request** can be seen below.
+
+::
+
+    {
+        "policy_type": "PerLearnerEnrollmentCreditAccessPolicy",
+        "description": "updated description",
+        "active": true,
+        "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
+        "catalog_uuid": "214cf999-5964-4a2e-afa8-c62461558315",
+        "subsidy_uuid": "214cf999-5964-4a2e-afa8-c62461342198",
+        "access_method": "direct",
+        "per_learner_enrollment_limit": 3,
+        "per_learner_spend_limit": 0,
+        "spend_limit": 0
+    }
+
+Outputs
+-------
+A sample response can be seen below.
+
+::
+
+    {
+        "policy_type": "PerLearnerEnrollmentCreditAccessPolicy",
+        "description": "updated description",
+        "active": true,
+        "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
+        "catalog_uuid": "214cf999-5964-4a2e-afa8-c62461558315",
+        "subsidy_uuid": "214cf999-5964-4a2e-afa8-c62461342198",
+        "access_method": "direct",
+        "per_learner_enrollment_limit": 3,
+        "per_learner_spend_limit": 0,
+        "spend_limit": 0
+    }
+
+
+Permissions
+-----------
+
+enterprise_admin
+  Should only create the transaction if the requesting user has implicit (JWT) or explicit (DB-defined) ``enterprise_admin`` role assigned.
--- a/enterprise_access/apps/api/serializers.py
+++ b/enterprise_access/apps/api/serializers.py
@@ -1,12 +1,22 @@
 """
 Serializers for Enterprise Access API v1.
 """
+import logging
 
+from django.apps import apps
 from django.urls import reverse
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from rest_framework import serializers
 
+from enterprise_access.apps.subsidy_access_policy.constants import (
+    NON_CREDIT_LIMIT_POLICY_TYPES,
+    POLICY_TYPE_CREDIT_LIMIT_FIELDS,
+    POLICY_TYPE_FIELD_MAPPER,
+    POLICY_TYPES_WITH_CREDIT_LIMIT,
+    AccessMethods,
+    PolicyTypes
+)
 from enterprise_access.apps.subsidy_access_policy.models import SubsidyAccessPolicy
 from enterprise_access.apps.subsidy_request.models import (
     CouponCodeRequest,
@@ -14,6 +24,8 @@ from enterprise_access.apps.subsidy_request.models import (
     SubsidyRequest,
     SubsidyRequestCustomerConfiguration
 )
+
+logger = logging.getLogger(__name__)
 
 
 class SubsidyRequestSerializer(serializers.ModelSerializer):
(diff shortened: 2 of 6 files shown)


enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+    def create(self, request, *args, **kwargs):
+        """
+        create action for SubsidyAccessPolicyCRUDViewset. Handles creation of policy after validation
+        """
+        policy_data = request.data
+        serializer = self.get_serializer(data=policy_data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
muhammad-ammar, line 800: this can also be written like
```python
serializer.is_valid(raise_exception=True)
serializer.save()
return Response(serializer.data, status=status.HTTP_201_CREATED)
```

enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+        Returns:
+            policy instance
+        """
+        if policy_uuid:
+            try:
+                instance = SubsidyAccessPolicy.objects.get(uuid=policy_uuid)
+            except ObjectDoesNotExist as exc:
+                error_msg = f'Policy UUID provided ({policy_uuid}) is not a valid UUID'
+                logger.exception(error_msg)
+                raise SubsidyAccessPolicyError(error_msg, status.HTTP_400_BAD_REQUEST) from exc
```
muhammad-ammar, line 817: can be written as
```python
policy = get_object_or_404(SubsidyAccessPolicy, pk=policy_uuid)
```

enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+        try:
+            instance = self._validate_policy(policy_uuid, policy_data)
+        except SubsidyAccessPolicyError as exc:
+            logger.exception(exc)
+            return Response(exc.message, exc.http_status_code)
+        serializer = self.get_serializer(instance, data=policy_data, partial=True)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_202_ACCEPTED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
muhammad-ammar, line 840:
```python
serializer.is_valid(raise_exception=True)
serializer.save()
return Response(serializer.data, status=status.HTTP_201_CREATED)
```

enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+            except ObjectDoesNotExist as exc:
+                error_msg = f'Policy UUID provided ({policy_uuid}) is not a valid UUID'
+                logger.exception(error_msg)
+                raise SubsidyAccessPolicyError(error_msg, status.HTTP_400_BAD_REQUEST) from exc
+            # for updates, ensure that uuid is not editable
+            if instance and policy_data.get('uuid') != str(instance.uuid):
+                error_msg = f'Policy UUID cannot be changed'
+                logger.exception(error_msg)
+                raise SubsidyAccessPolicyError(error_msg, status.HTTP_400_BAD_REQUEST)
+            return instance
```
muhammad-ammar, line 823: I think this can be moved into serializer.

On a second thought, `uuid` of the policy will be auto-generated. So I think we should not consider uuid coming from client. In that case I think we don't need to do any validation for uuid for put/post/patch?

enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+        """
+        Used for http patch method. Updates the policy data passed in request after validation.
+        """
+        policy_data = request.data
+        policy_uuid = self.kwargs.get('uuid')
+        try:
+            instance = self._validate_policy(policy_uuid, policy_data)
+        except SubsidyAccessPolicyError as exc:
+            logger.exception(exc)
+            return Response(exc.message, exc.http_status_code)
```
muhammad-ammar, line 835: I think this will be simplified as per my comments above.

enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+        if not group_uuid:
+            msg = '"group_uuid" query param is required'
+            return Response(msg, status=status.HTTP_400_BAD_REQUEST)
+        return super().list(request, *args, **kwargs)
+
+    def retrieve(self, request, *args, **kwargs):
+        """
+        Retrieve action for SubsidyAccessPolicyCRUDViewset
+        """
+        return super().retrieve(request, *args, **kwargs)
```
muhammad-ammar, line 856: I think this is the default implementation of `retrieve`? If yes, then can we remove `retrieve`?

enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def list(self, request, *args, **kwargs):
+        """
+        List action for SubsidyAccessPolicyCRUDViewset. Show all policies linked with the group uuid passed in request.
+        """
+        group_uuid = self.requested_group_uuid
+        if not group_uuid:
+            msg = '"group_uuid" query param is required'
+            return Response(msg, status=status.HTTP_400_BAD_REQUEST)
```
muhammad-ammar, line 849: It may be a good idea to have a same format for all errors returned to simplify the error handling on client side?

enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+    def get_queryset(self):
+        """
+        Queryset for the SubsidyAccessPolicyCRUDViewset. The admins should only be able to perform CRUD operations
+        on the enterprise's policies they belong to.
+        """
+        queryset = SubsidyAccessPolicy.objects.order_by('-created')
+        group_uuid = self.requested_group_uuid
+        if group_uuid:
+            queryset = queryset.filter(group_uuid=group_uuid)
+        return queryset
```
muhammad-ammar, line 867: `requests.has_admin_access` permission will be checked before the view code is executed? If yes, then I think we will always get `group_uuid` because `group_uuid` is required for permission code + others. If yes, then I think we can remove `if group_uuid:` above.

enterprise_access/apps/api/v1/views.py
```diff
@@ -746,7 +746,128 @@ def partial_update(self, request, *args, **kwargs):
...
+     Viewset for Subsidy Acess Policy CRUD operations.
+     """
+
+    permission_classes = (permissions.IsAuthenticated,)
+    serializer_class = serializers.SubsidyAccessPolicyCRUDSerializer
+    authentication_classes = (JwtAuthentication,)
+    filter_backends = (filters.OrderingFilter, DjangoFilterBackend,)
+    filterset_fields = ('group_uuid', 'policy_type',)
+    pagination_class = PaginationWithPageCount
+    http_method_names = ['get', 'head', 'options', 'post', 'patch', 'delete']
```
muhammad-ammar, line 760: Any idea what is the default value of this? Do we really need to add all the actions?

enterprise_access/apps/api/serializers.py
```diff
@@ -141,7 +148,69 @@ def validate_content_key(self, value):
 
         return value
 
-class SubsidiyAccessPolicyListSerializer(SubsidiyAccessPolicyRedeemSerializer):  # pylint: disable=abstract-method
+
+class SubsidyAccessPolicyCRUDSerializer(serializers.ModelSerializer):
+    """
+    Serializer to validate policy data for CRUD operations.
+    """
+    uuid = serializers.UUIDField(read_only=True)
```
muhammad-ammar, line 156: Looking at this now. Does that mean `uuid` will not be used in any write operation?

enterprise_access/apps/api/serializers.py
```diff
@@ -141,7 +148,69 @@ def validate_content_key(self, value):
...
+    policy_type = serializers.ChoiceField(choices=PolicyTypes.CHOICES)
```
muhammad-ammar, line 157: `policy_type` is kind of private field and used to track the type of the actual(`proxy`) class of the policy. This should not be publicly exposed.

enterprise_access/apps/api/serializers.py
```diff
@@ -141,7 +148,69 @@ def validate_content_key(self, value):
...
+    description = serializers.CharField(max_length=None, min_length=None, allow_blank=False, trim_whitespace=True)
+    active = serializers.BooleanField()
+    group_uuid = serializers.UUIDField(allow_null=False)
+    catalog_uuid = serializers.UUIDField(allow_null=False)
+    subsidy_uuid = serializers.UUIDField(allow_null=False)
+    access_method = serializers.ChoiceField(choices=AccessMethods.CHOICES)
+
+    per_learner_enrollment_limit = serializers.IntegerField()
+    per_learner_spend_limit = serializers.IntegerField()
+    spend_limit = serializers.IntegerField()
```
muhammad-ammar, line 167: The above 3 fields are not applicable to every policy. So should we check/validate or specify that which field will be used for which policy?

enterprise_access/apps/subsidy_access_policy/constants.py
```diff
@@ -25,6 +25,16 @@ class SegmentEvents:
     SUBSIDY_ACCESS_POLICY_CREATED = 'edx.server.enterprise-access.enrollment-lifecycle.subsidy-access-policy.created'
     SUBSIDY_ACCESS_POLICY_REDEEMED = 'edx.server.enterprise-access.enrollment-lifecycle.subsidy-access-policy.redeemed'
 
+class PolicyTypes:
```
muhammad-ammar, line 28: I think we don't need this as I explained earlier. These types are for internal use only.

enterprise_access/apps/api/serializers.py
muhammad-ammar: I think we would need this to create an object of specific proxy type. Lets have a call tomorrow morning to discuss this.

enterprise_access/apps/api/v1/urls.py
```diff
@@ -9,7 +9,8 @@
 
 router = DefaultRouter()
 
-router.register("policy", views.SubsidyAccessPolicyViewset, 'policy')
+router.register("admin/policy", views.SubsidyAccessPolicyCRUDViewset, 'admin-policy')
+router.register("policy", views.SubsidyAccessPolicyRedeemViewset, 'policy')
```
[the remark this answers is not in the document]
sameenfatima78: Yeah, I just thought to keep the original one also in use as our problem was already solved with `admin/policy`. Also, according to [ENT-6856](https://2u-internal.atlassian.net/browse/ENT-6856), we are required to return redeemable policies at `/policy`

enterprise_access/apps/api/serializers.py
sameenfatima78: Yep, it will be auto-generated via [model](https://github.com/openedx/enterprise-access/blob/main/enterprise_access/apps/subsidy_access_policy/models.py#L37) and we would want to ensure nobody changes the value while updating the instance. so that's why we added `read-only`

enterprise_access/apps/api/serializers.py
sameenfatima78: Yep, we would also need this to add validation logic around "showing relevant limit field based on selected policy type "

enterprise_access/apps/api/v1/views.py
sameenfatima78: Good catch, I will check if removing this line would have any effect on the listed actions.

enterprise_access/apps/api/v1/views.py
sameenfatima78: For POST/PUT, yes. For PATCH, the interface is more flexible in customizing a payload. So admin can easily add a key-value for uuid in the payload. Just being extra cautious here. I had initially added this logic in serializer but obviously, it automatically removes uuid from payload so this condition `policy_data.get('uuid') != str(instance.uuid)` always passes as left side is always None.

enterprise_access/apps/api/v1/views.py
sameenfatima78: Yep. I had added it just for some local testing. It's removed now.

enterprise_access/apps/api/serializers.py
sameenfatima78: Or maybe we could just discard the other two limits if they are non-zero in serializer validate method?

enterprise_access/apps/api/v1/views.py
sameenfatima78: These are all the methods that are allowed by default https://github.com/django/django/blob/16411b8400ad08f90c236bb2e18f65c655f903f8/django/views/generic/base.py#L34

enterprise_access/apps/subsidy_access_policy/constants.py
sameenfatima78: As per new discussion, we do need it.

muhammad-ammar (approved):

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -93,3 +93,145 @@ Permissions
...
+======================================
+
+GET **/api/v1/admin/policy/**
+
+List all policies linked to a group.
+
+Inputs
+------
+
+- ``group_uuid`` (Query Param, required): The uuid of the customer. For now it will be an enterprise customer uuid.
```
johnnagro, line 108: i am wondering if we should use `enterprise_customer_uuid` here rather than group. group is going to have a very specific meaning in the future - a way to group learners within an enterprise. we will still need to be able to query based on enterprise customer to find all policies regardless of the (future) groups associated with the policies.

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -93,3 +93,145 @@ Permissions
...
+
+::
+
+   [
+        {
+            "uuid": "d0cd25f9-0b73-49c3-8299-dc9751a12ef5",
+            "policy_type": "PerLearnerEnrollmentCreditAccessPolicy",
+            "description": "sdad",
+            "active": true,
+            "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
```
johnnagro, line 122: we probably want `enterprise_customer_uuid` here, `group` in this context will have a specific, important meaning in the future.

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -93,3 +93,145 @@ Permissions
...
+A sample response can be seen below.
+
+::
+
+    {
+        "uuid": "d0cd25f9-0b73-49c3-8299-dc9751a12ef5",
+        "policy_type": "PerLearnerEnrollmentCreditAccessPolicy",
+        "description": "sdad",
+        "active": true,
+        "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
```
johnnagro, line 172: we probably want `enterprise_customer_uuid` here, `group` in this context will have a specific, important meaning in the future.

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -93,3 +93,145 @@ Permissions
...
+
+A sample **request** can be seen below.
+
+::
+
+    {
+        "policy_type": "PerLearnerEnrollmentCreditAccessPolicy",
+        "description": "updated description",
+        "active": true,
+        "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
```
johnnagro, line 204: we probably want `enterprise_customer_uuid` here, `group` in this context will have a specific, important meaning in the future.

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -93,3 +93,145 @@ Permissions
...
+-------
+A sample response can be seen below.
+
+::
+
+    {
+        "policy_type": "PerLearnerEnrollmentCreditAccessPolicy",
+        "description": "updated description",
+        "active": true,
+        "group_uuid": "214cf999-5964-4a2e-afa8-c62461558211",
```
johnnagro, line 223: we probably want `enterprise_customer_uuid` here, `group` in this context will have a specific, important meaning in the future.

johnnagro (changes requested):
we should discuss the use of `group` versus `enterprise_customer` so that we can come to a consensus either way

docs/decisions/0003-initial-api-specification.rst
macdiesel: I would like to +1 this. I think that the language here needs to be very precise.

macdiesel (approved):

docs/decisions/0003-initial-api-specification.rst
sameenfatima78: Created https://2u-internal.atlassian.net/browse/ENT-6902 to take care of this.

PR #115, opened 2023-03-13
feat: ENT-6843 dispatched event for subsidy redemption

irfanuddinahmad:
Description: This PR adds an event on the openedx event bus upon redemption of a subsidy

JIRA: https://2u-internal.atlassian.net/browse/ENT-6843

### DIFF
--- a/enterprise_access/apps/api/serializers.py
+++ b/enterprise_access/apps/api/serializers.py
@@ -123,7 +123,7 @@ class SubsidyRequestCustomerConfigurationSerializer(serializers.ModelSerializer)
         return super().update(instance, validated_data)
 
 
-class SubsidiyAccessPolicyRedeemSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+class SubsidyAccessPolicyRedeemSerializer(serializers.Serializer):  # pylint: disable=abstract-method
     """
     Serializer to validate policy redeem request POST data.
     """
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/api/v1/views.py
```diff
@@ -826,7 +828,11 @@ def redeem(self, request, *args, **kwargs):
                 )
             if policy.can_redeem(learner_id, content_key):
                 response = policy.redeem(learner_id, content_key)
-
+                if settings.KAFKA_ENABLED:  # pragma: no cover
```
muhammad-ammar, line 831: I think we should move the `if` condition inside the `send_subsidy_redemption_event_to_event_bus`. The caller of this method should not take the responsibility to check if `Kafka` is enabled/disabled.

muhammad-ammar (approved):
LGTM. Left one minor comment.

saleem-latif (approved):

enterprise_access/apps/api/v1/views.py
irfanuddinahmad: good point ... done

enterprise_access/apps/api/serializers.py
```diff
@@ -123,7 +123,7 @@ def update(self, instance, validated_data):
         return super().update(instance, validated_data)
 
 
-class SubsidiyAccessPolicyRedeemSerializer(serializers.Serializer):  # pylint: disable=abstract-method
+class SubsidyAccessPolicyRedeemSerializer(serializers.Serializer):  # pylint: disable=abstract-method
```
iloveagent57, line 126: ah, good catch

iloveagent57 (approved):

PR #116, opened 2023-03-14
ENT-6863: feat: Added enterprise subsidy api client to communicate with subsidy app.

saleem-latif:
__Jira Ticket:__ [ENT-6863](https://2u-internal.atlassian.net/browse/ENT-6863)

__Description:__
This PR adds subsidy API client to handle communication with the subsidy app.


iloveagent57 (approved):
👍 This is a good starting point!

PR #117, opened 2023-03-14
feat: add policy redemption endpoint to return list of redemptions belong to a learner

muhammad-ammar:


saleem-latif (approved):

sameenfatima78 (approved):

enterprise_access/apps/api/v1/views.py
```diff
@@ -925,3 +925,26 @@ def redeem(self, request, *args, **kwargs):
...
+        redemptions = []
+        policies = SubsidyAccessPolicy.objects.filter(group_uuid=group_uuid)
+        for policy in policies:
+            redemption = policy.has_redeemed(learner_id, content_key)
+            redemptions.append(redemption)
+
+        return Response(
+            redemptions,
+            status=status.HTTP_200_OK,
+        )
```
iloveagent57, line 950: I think we want to run `redemptions` through a serializer before returning it, right?

docs/decisions/0003-initial-api-specification.rst
```diff
@@ -94,6 +94,29 @@ enterprise_learner
   Should only create the transaction if the requesting user has implicit (JWT) or explicit (DB-defined)
   ``enterprise_learner`` role assigned.
 
+GET access policy redemption
+===============================
+**/api/v1/policy/redemption/**
```
iloveagent57, line 99: We probably want to pluralize `redemptions`.

Also, for these URL routes, I wonder if something like this would be easier to deal with:
1. `/api/v1/policy/<enterprise-customer-uuid>/?learner_id=...&content_key=...` to list redeemable policies
2. `/api/v1/policy/<enterprise-customer-uuid>/redeem/<policy_uuid>/?learner_id=...&content_key=...` to start a redemption command.
3. `/api/v1/policy/<enterprise-customer-uuid>/redemptions/?learner_id=...&content_key=...` to query which redemptions already exist for the (learner, content) pair.

I know (2) introduces some nested routing which is a little complex.  If that gets hairy, we could make `policy_uuid` a required part of the POST body, or a query param.

iloveagent57 (commented):
Let me know how complex it is to switch around that URL routing.  If that would end up blocking work or blocking manual testing, we can schedule it as a follow-on ticket/PR.

enterprise_access/apps/api/v1/views.py
muhammad-ammar: redemptions is a list of responses where each response is the return value of `policy.has_redeemed` received from enterprise-subsidy. This is not finalized yet due to unavailability of communication between enterprise-access and enterprise-subsidy. This whole viewset will be revisited once all the endpoints that are needed will be implemented.

docs/decisions/0003-initial-api-specification.rst
muhammad-ammar: Should be easy to go with above suggested routing. Let me try.

enterprise_access/apps/api/v1/views.py
iloveagent57: Ah, ok, makes sense.

docs/decisions/0003-initial-api-specification.rst
muhammad-ammar: The easiest way possible is to not use the DRF Router and manually define the urls. Looking for more ways.

muhammad-ammar: The above routing is possible using the below + some minor modifications in the viewsest.

```python
policy_list = views.SubsidyAccessPolicyRedeemViewset.as_view({
    'get': 'list',
})

policy_redeem = views.SubsidyAccessPolicyRedeemViewset.as_view({
    'post': 'redeem'
})

policy_redemption = views.SubsidyAccessPolicyRedeemViewset.as_view({
    'get': 'redemption',
})

urlpatterns = [
    path('policy/<uuid:enterprise_customer_uuid>/', policy_list, name='policy-list'),
    path('policy/<uuid:enterprise_customer_uuid>/redeem/<uuid:policy_uuid>/', policy_redeem, name='policy-redeem'),
    path('policy/<uuid:enterprise_customer_uuid>/redemption', policy_redemption, name='policy-redemption'),
]
```

But now looking at the 2nd url route above, I am not sure why we need `enterprise_customer_uuid` at all? We don't need `enterprise_customer_uuid` in the [redeem view](https://github.com/openedx/enterprise-access/blob/main/enterprise_access/apps/api/v1/views.py#L906-L938). First url route above will only return the policies that are associated with an enterprise so I believe we don't need `enterprise_customer_uuid` in 2nd url route.

An HTTP GET `http://localhost:18270/api/v1/policy/12aacfee-8ffa-4cb3-bed1-059565a57f06/?learner_id=1234&content_key=course-v1:edX+edXPrivacy101+3T2020` will return the below response. Please see the `policy_redemption_url` below.
```json
[
    {
        "uuid": "9e63269a-1e80-4371-b7fe-a8878973b4d6",
        "policy_redemption_url": "/api/v1/policy/9e63269a-1e80-4371-b7fe-a8878973b4d6/",
        "policy_type": "SubscriptionAccessPolicy",
        "enterprise_customer_uuid": "12aacfee-8ffa-4cb3-bed1-059565a57f06",
        "description": "Subscription access policy",
        "active": true,
        "catalog_uuid": "4ccd4516-0c4d-45b5-b653-b28e158322c2",
        "subsidy_uuid": "103cdc29-d633-4384-87b0-bc699f2307c5",
        "group_uuid": "12aacfee-8ffa-4cb3-bed1-059565a57f06",
        "access_method": "direct",
        "per_learner_enrollment_limit": 0,
        "per_learner_spend_limit": 0,
        "spend_limit": 0
    }
]
```

muhammad-ammar: @iloveagent57 Can we live with the existing urls for now and merge this PR?

iloveagent57: >But now looking at the 2nd url route above, I am not sure why we need enterprise_customer_uuid at all? We don't need enterprise_customer_uuid in the [redeem view](https://github.com/openedx/enterprise-access/blob/main/enterprise_access/apps/api/v1/views.py#L906-L938). First url route above will only return the policies that are associated with an enterprise so I believe we don't need enterprise_customer_uuid in 2nd url route.

Ah, you make a very good point.  Yes, I think it's fine to merge for now and come back to this in a later PR.

PR #118, opened 2023-03-14
feat: ENT-6843 added events for access policy create-update

irfanuddinahmad:
Description: This PR adds support for pushing access policy create/update events to the openedx event bus

JIRA: https://2u-internal.atlassian.net/browse/ENT-6843


iloveagent57 (approved):
Looks good.  I wonder if it'll be simpler to send these event types on a Django `post_save` signal to make the API implementations a little cleaner?

PR #120, opened 2023-03-16
fix: replace group_uuid with enterprise_customer_uuid

muhammad-ammar:


enterprise_access/apps/api/v1/tests/test_views.py
```diff
@@ -1711,7 +1711,7 @@ def test_list_policies_for_enterprise(self):
 
     def test_list_without_group_uuid_query_param(self):
```
sameenfatima78, line 1712: test name had group_uuid in it, can we replace it with enterprise_uuid?

sameenfatima78 (approved):
one minor change requested, otherwise all good!

enterprise_access/apps/api/v1/tests/test_views.py
muhammad-ammar: good catch.

muhammad-ammar: done

PR #121, opened 2023-03-20
feat: new endpoint to return policies with credit available

sameenfatima78:


sameenfatima78 (commented):
@adamstankiewicz Want to know your thoughts on discarding `remaining_balance_per_user` from the API and just using `remaining_balance` instead. I think we can always tell from the `policy_type` if the remaining balance represents credit that is available to just the learner or is it useable enterprise wide. e:g for `PerLearnerSpendCreditPolicy`, `remaining_balance` would be the amount the current learner can redeem. For `CappedEnrollmentLearnerCreditAccessPolicy`, the `remaining_balance` will tell how much credit is available to all learners. Let me know what you think? 👀  

https://github.com/openedx/enterprise-access/blob/main/enterprise_access/apps/subsidy_access_policy/models.py#L272

adamstankiewicz:
> Want to know your thoughts on discarding `remaining_balance_per_user` from the API and just using `remaining_balance` instead. I think we can always tell from the `policy_type` if the remaining balance represents credit that is available to just the learner or is it useable enterprise wide.

@sameenfatima78 What about an access policy that has both an enterprise-wide spent limit and a per-learner spend limit together as a single policy (this is a valid use case we have today)? In that case, wouldn't we need to have the 2 fields for enterprise-wide and learner-specific balance remaining?

Also, from the frontends perspective, I don't believe we could (or should) rely on the `policy_type` to deduce what `remaining_balance` means semantically because that would likely result in some conditional logic in the frontend to change how it used the `remaining_balance` field based on hardcoded `policy_type` names?

sameenfatima78:
> 
> @sameenfatima78 What about an access policy that has both an enterprise-wide spent limit and a per-learner spend limit together as a single policy (this is a valid use case we have today)? In that case, wouldn't we need to have the 2 fields for enterprise-wide and learner-specific balance remaining?
> 
> Also, from the frontends perspective, I don't believe we could (or should) rely on the `policy_type` to deduce what `remaining_balance` means semantically because that would likely result in some conditional logic in the frontend to change how it used the `remaining_balance` field based on hardcoded `policy_type` names?

Hmm, this makes sense. I agree with all your points. But I'm still confused what would `remaining_balance` represent for policy types that are only meant to store learner-specific credits, and how will we calculate it? Like for `PerLearnerSpendCreditPolicy`, we can calculate `remaining_balance_per_user` by taking the learner's current spent and subtract it from `per_learner_spend_limit`, what about `remaining_balance`?

adamstankiewicz:
@sameenfatima78 I believe even in the case for `PerLearnerSpendCreditPolicy`, the `remaining_balance` would still represent the total remaining balance for the associated subsidy.

`remaining_balance` is largely a pass-thru for how much total balance is left on the associated subsidy's ledger.

Say a `PerLearnerSpendCreditPolicy` has a `per_learner_spend_limit` of $500, but the subsidy ledger balance only has $250 remaining. In this case, a learner who hasn't spent any learner credit yet would get a response similar to:

```json
{
  "remaining_balance": 250,
  "remaining_balance_for_user": 500,
}
```

In the enterprise learner portal MFE, for the case described above, we rely on both `remaining_balance` and `remaining_balance_for_user` to determine whether any given enterprise offer has low/no balance remaining.

Pseudocode:

```jsx
const isOfferLowOnBalance = (offer) => {
  // does remaining balance have less than a certain threshold, e.g. only 10% funds remaining?
  const hasLowBalance = offer.remainingBalance <= lowOfferDollarThreshold;

  // does user's remaining balance have less than a predefined dollar amount, e.g. $149?
  const hasLowUserBalance = offer.remainingBalanceForUser <= lowOfferUserDollarThreshold;

  return hasLowBalance || hasLowUserBalance;
}
```

```jsx
const isOfferOutOfBalance = (offer) => {
  // does remaining balance have less than a predefined dollar amount, e.g. $99?
  const hasNoBalance = offer.remainingBalance <= outOfOfferDollarThreshold;

  // does user's remaining balance have less than a predefined dollar amount, e.g. $99?
  const hasNoUserBalance = offer.remainingBalanceForUser <= outOfOfferUserDollarThreshold;

  return hasNoBalance || hasNoUserBalance;
}
```

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -260,14 +260,29 @@ class Meta:
...
+    def credit_available(self, learner_id=None):
+        if self.remaining_credit(learner_id) > 0:
+            return True
+        return False
+
+    def remaining_credit(self, learner_id=None):
+        """
+        Returns the remaining redeemable credit for the policy.
+        """
+        learner_transaction_count = subsidy_client.transactions_for_learner(
```
adamstankiewicz, line 280: [curious] Is the intent for `transactions_for_learner` only to return transactions in a "committed" (i.e., fulfilled) state and without any reversals that would negate the transaction?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -285,15 +300,30 @@ class Meta:
...
+    def credit_available(self, learner_id=None):
+        if self.remaining_credit(learner_id) > 0:
+            return True
+        return False
+
+    def remaining_credit(self, learner_id=None):
+        """
+        Returns the remaining redeemable credit for the policy.
+        """
+        spent_amount = subsidy_client.amount_spent_for_learner(
```
adamstankiewicz, line 321: [curious] Is the intent for `amount_spent_for_learner ` only to return the amount spent for transactions in a "committed" (i.e., fulfilled) state and without any reversals that would negate the transaction?

enterprise_access/apps/subsidy_access_policy/models.py
adamstankiewicz: For example, say a learner enrolls in a course and then unenrolls with the 14 refund day window so a transaction reversal is created to add funds back to the ledger balance. In that case, would there be 2 transactions for the learner that negate each other?

pwnage101: I think existing Transactions in any state besides "committed" should count towards the consumed credit for a policy because they represent in-process attempts to redeem content, and ignoring other states could result in a race condition of over-allocation of redemptions.  That said, policy-level locking should prevent this race condition in the first place.

pwnage101: Regarding reversals: I can't speak to those, it sounds like a product question for standup.

sameenfatima78:
> @sameenfatima78 I believe even in the case for `PerLearnerSpendCreditPolicy`, the `remaining_balance` would still represent the total remaining balance for the associated subsidy.
> 
> `remaining_balance` is largely a pass-thru for how much total balance is left on the associated subsidy's ledger.
> 
> Say a `PerLearnerSpendCreditPolicy` has a `per_learner_spend_limit` of $500, but the subsidy ledger balance only has $250 remaining. In this case, a learner who hasn't spent any learner credit yet would get a response similar to:
> 
> ```json
> {
>   "remaining_balance": 250,
>   "remaining_balance_for_user": 500,
> }
> ```
> 
> In the enterprise learner portal MFE, for the case described above, we rely on both `remaining_balance` and `remaining_balance_for_user` to determine whether any given enterprise offer has low/no balance remaining.
> 
> Pseudocode:
> 
> ```js
> const isOfferLowOnBalance = (offer) => {
>   // does remaining balance have less than a certain threshold, e.g. only 10% funds remaining?
>   const hasLowBalance = offer.remainingBalance <= lowOfferDollarThreshold;
> 
>   // does user's remaining balance have less than a predefined dollar amount, e.g. $149?
>   const hasLowUserBalance = offer.remainingBalanceForUser <= lowOfferUserDollarThreshold;
> 
>   return hasLowBalance || hasLowUserBalance;
> }
> ```
> 
> ```js
> const isOfferOutOfBalance = (offer) => {
>   // does remaining balance have less than a predefined dollar amount, e.g. $99?
>   const hasNoBalance = offer.remainingBalance <= outOfOfferDollarThreshold;
> 
>   // does user's remaining balance have less than a predefined dollar amount, e.g. $99?
>   const hasNoUserBalance = offer.remainingBalanceForUser <= outOfOfferUserDollarThreshold;
> 
>   return hasNoBalance || hasNoUserBalance;
> }
> ```

@adamstankiewicz Thank you for the detailed answer, things are much clearer now. I have incorporated your feedback and added the fields "remaining_balance_per_user" and "remaining_balance" as suggested. Could you please take another look?

enterprise_access/apps/subsidy_access_policy/models.py
sameenfatima78: That's how I'm hoping this endpoint in `enterprise-subsidy` should work.

enterprise_access/apps/subsidy_access_policy/models.py
sameenfatima78: @adamstankiewicz That's a good point. I think this logic should be handled at `enterprise-subsidy` side.

adamstankiewicz: > I think existing Transactions in any state besides "committed" should count towards the consumed credit for a policy because they represent in-process attempts to redeem content

@pwnage101 Agreed, I could see Transactions in a created/pending state also count towards the consumed credit, but what about a transaction that we've failed to fulfill (e.g., learner not actually enrolled)? Looking at [`TransactionStateChoices`](https://github.com/openedx/openedx-ledger/blob/35e2b1ce22740f649ec3385adfa352e94797104d/openedx_ledger/models.py#L25-L38), I wonder if there should be an error state for transactions?

> Regarding reversals: I can't speak to those, it sounds like a product question for standup.

The reversals suggestion comes from a recent bug in ecommerce with existing learner credit where we didn't consider refunds when querying for learners' prior orders, so they weren't able to re-enroll because ecommerce thought they had already placed an order. I'll bring this up in standup then (likely a Milestone 2 thing anyways).

iloveagent57 (approved):

PR #128, opened 2023-03-29
fix: fix attribute error with subsidy client 0.2.4

iloveagent57:


johnnagro (approved):
⛵

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -259,10 +259,7 @@ def release_lock(self, learner_id, content_key):  # pylint: disable=unused-argum
         Release a previously acquired lock for transaction isolation.
         TODO: use this method.
         """
-        return release_subsidy_policy_lock(
-            self.uuid,
-            django_cache_timeout=SUBSIDY_POLICY_LOCK_TIMEOUT_SECONDS,
```
johnnagro, line 264: this kwarg didnt need to be passed in, and wasn'st being plucked out, so was being passed into the content_key creator, which was resulting in a different cache key

johnnagro (approved):
⛵

PR #134, opened 2023-04-20
feat: redemption metadata

johnnagro:
# Description

- This PR contains an ADR and the start of what modifications would be required to enact it.
- This ADR intends to outline modifications to existing API endpoint(s) such that MFEs can pass additional metadata into the redemption flow. This extension was devised to support required GEAG meatadata such as DOB and terms acceptance dates.
- if folks generally vibe with this, I will flesh out the PR to include a full implementation

# References

- https://github.com/openedx/edx-enterprise-subsidy-client/pull/15

### DIFF
--- a/docs/decisions/0008-additional-redemption-metadata.rst
+++ b/docs/decisions/0008-additional-redemption-metadata.rst
@@ -9,12 +9,12 @@ Proposed
 Context
 =======
 
-This document intends to outline modifications to existing API endpoint(s) such that MFEs can pass additional metadata into the redemption flow. This extention was devised to support required GEAG meatadata such as DOB and terms acceptance dates.
+This document intends to outline modifications to existing API endpoint(s) such that MFEs can pass additional metadata into the redemption flow. This extension was devised to support required GEAG meatadata such as DOB and terms acceptance dates.
 
 Decision
 =======
 
-The GEAG system requires additional metadata about a learner in order to process an allocation. This additional metadata (name, dob, etc) are collected during the enrollment flow on the edX side before enrollment. This additional metadata is not persisted anywhere on the edX side. Because of the lack of persistence and the placement of this information collection on the edX side of the process we decided to create a faicility for the frontend to pass along additional metadata into the redemption flow.
+The GEAG system requires additional metadata about a learner in order to process an allocation. This additional metadata (name, dob, etc) are collected during the enrollment flow on the edX side before enrollment. This additional metadata is not persisted anywhere on the edX side. Because of the lack of persistence and the placement of this information collection on the edX side of the process we decided to create a facility for the frontend to pass along additional metadata into the redemption flow.
 
 
 POST access policy redeem transaction
--- a/docs/decisions/0008-additional-redemption-metadata.rst
+++ b/docs/decisions/0008-additional-redemption-metadata.rst
@@ -11,6 +11,12 @@ Context
 
 This document intends to outline modifications to existing API endpoint(s) such that MFEs can pass additional metadata into the redemption flow. This extention was devised to support required GEAG meatadata such as DOB and terms acceptance dates.
 
+Decision
+=======
+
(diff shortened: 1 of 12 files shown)


docs/decisions/0008-additional-redemption-metadata.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Status
+======
+
+Proposed
+
+Context
+=======
+
+This document intends to outline modifications to existing API endpoint(s) such that MFEs can pass additional metadata into the redemption flow. This extension was devised to support required GEAG meatadata such as DOB and terms acceptance dates.
```
adamstankiewicz, line 12: nit: consider defining GEAG before using the acronym.

docs/decisions/0008-additional-redemption-metadata.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Decision
+=======
+
+The GEAG system requires additional metadata about a learner in order to process an allocation. This additional metadata (name, dob, etc) are collected during the enrollment flow on the edX side before enrollment. This additional metadata is not persisted anywhere on the edX side. Because of the lack of persistence and the placement of this information collection on the edX side of the process we decided to create a facility for the frontend to pass along additional metadata into the redemption flow.
```
adamstankiewicz, line 17: > Because of the lack of persistence...

[clarification] Is part of the decision here to start persisting the metadata (likely alongside the transaction I'd assume)? Might be worth expanding on the decision to explicitly mention that, if so.

adamstankiewicz, line 17: Does there need to be or should there be any validation on the `redeem` endpoint such that we can verify the necessary fields in `metadata` are passed in order to make the GEAG allocation before attempting to call `enterprise-subsidy`/GEAG? Or is the intent for this API to largely pass through the validation errors returned by GEAG?

docs/decisions/0008-additional-redemption-metadata.rst
```diff
@@ -0,0 +1,55 @@
...
+
+
+POST access policy redeem transaction
+======================================
+**/api/v1/policy/<policy_uuid>/redeem/**
+
+This is the existing API endpoint to eedeem subsidy value by making a request to enterprise-subsidy service.
```
adamstankiewicz, line 24: nit: typo. "eedeem" -> "redeem"

docs/decisions/0008-additional-redemption-metadata.rst
```diff
@@ -9,12 +9,12 @@ Proposed
 Context
 =======
 
-This document intends to outline modifications to existing API endpoint(s) such that MFEs can pass additional metadata into the redemption flow. This extension was devised to support required GEAG meatadata such as DOB and terms acceptance dates.
+This document intends to outline modifications to existing API endpoint(s) such that MFEs can pass additional metadata into the redemption flow. This extension was devised to support required GetSmarter Enteprise Api Gateway (GEAG) meatadata such as DOB and terms acceptance dates.
```
adamstankiewicz, line 12: nit: typo "Ente**r**prise"

docs/decisions/0008-additional-redemption-metadata.rst
johnnagro: By lack of persistence i meant that the MFE isn't saving it anywhere - such as the user profile - which would allow the subsidy system to retrieve it for enrollment into GEAG. Because its scope of existence is only this enrollment flow we'll need to pass it directly into the redemption flow. TBD on how much of this information will be persisted by the Subsidy service.

johnnagro: IMHO any validation involved with this data will be done by the subsidy service since has knowledge of the fulfillment systems - in this case enterprise access is mostly just passing it along and collecting the response to send back to the caller.

adamstankiewicz: Makes sense to me!

adamstankiewicz (approved):
This looks good to me, once quality/CI is passing.

PR #135, opened 2023-04-21
fix: policy_redemption_url should be a full URI

pwnage101:
ENT-6992

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_views.py
@@ -2,6 +2,7 @@
 Tests for Enterprise Access API v1 views.
 """
 import random
+import re
 from unittest.mock import call, patch
 from uuid import UUID, uuid4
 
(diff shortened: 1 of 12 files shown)


enterprise_access/apps/api/v1/tests/test_views.py
```diff
@@ -1604,14 +1604,26 @@ def test_list_all_redeemable_only_policies(self):
         response = self.client.get(SUBSIDY_ACCESS_POLICY_LIST_ENDPOINT, query_params)
         response_json = self.load_json(response.content)
 
-        # Verify that api response only includes the redeemable policies
-        assert response_json == [
-            dict(SubsidyAccessPolicyRedeemableSerializer([self.redeemable_policy], many=True).data[0])
```
pwnage101, line 1609: IMO, testing a serializer with the same serializer is self-fulfilling.  It's what led to this bug creeping in after all!

adamstankiewicz (approved):

iloveagent57 (approved):

PR #137, opened 2023-04-25
fix: update response for can-redeem API endpoint to remove business logic on client

adamstankiewicz:
# Description

In the `/api/v1/polic/enterprise-customer/<enterprise_customer_uuid>/can-redeem/` API endpoint:

* Adds missing migrations about removed models.
* Updated the model fields around per-learner enrollment and spend limits to default to `None` instead of `0` and, related, updated the `can_redeem` validation logic to account for `None`. This allows policies to be created with no limits enforced on a policy.
* Renames `subsidy_access_policy` to `redeemable_subsidy_access_policy` in the response to more clearly communicate that it is indeed redeemable to consumers.
* Adds a `can_redeem` helper boolean property in the response when `redeemable_subsidy_access_policy` is returned so that the frontend doesn't need to compute its own boolean.
* Adds a `has_successful_redemption` helper boolean property in the response so the frontend doesn't necessarily need to iterate through all redemptions in the returned list to find if one of them has `state: "committed"`. May eventually be used to determine whether learner is already enrolled. Not currently used given course page in the learner portal needs to support existing enrollments from other non-EMET (legacy) subsidies.

### DIFF
--- a/enterprise_access/apps/api/v1/views.py
+++ b/enterprise_access/apps/api/v1/views.py
@@ -1238,10 +1238,13 @@ class SubsidyAccessPolicyRedeemViewset(UserDetailsFromJwtMixin, PermissionRequir
                 resolved_policy = SubsidyAccessPolicy.resolve_policy(redeemable_policies)
                 serialized_policy = serializers.SubsidyAccessPolicyRedeemableSerializer(resolved_policy).data
 
+            has_successful_redemption = any(redemption['state'] == 'committed' for redemption in redemptions)
             can_redeem_for_content_response = {
                 "content_key": content_key,
                 "redemptions": redemptions,
-                "subsidy_access_policy": serialized_policy,
+                "has_redeemed": has_successful_redemption,
+                "redeemable_subsidy_access_policy": serialized_policy,
+                "can_redeem": bool(serialized_policy),
                 "reasons": reasons,
             }
             response.append(can_redeem_for_content_response)
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -85,17 +85,17 @@ class SubsidyAccessPolicy(TimeStampedModel):
     per_learner_enrollment_limit = models.IntegerField(
         null=True,
         blank=True,
-        default=0,
+        default=None,
     )
     per_learner_spend_limit = models.IntegerField(
         null=True,
         blank=True,
-        default=0,
+        default=None,
     )
     spend_limit = models.IntegerField(
         null=True,
         blank=True,
-        default=0,
+        default=None,
     )
 
     history = HistoricalRecords()
@@ -375,7 +375,11 @@ class PerLearnerEnrollmentCreditAccessPolicy(SubsidyAccessPolicy, CreditPolicyMi
         Checks if the given learner_id has a number of existing subsidy transactions
         LTE to the learner enrollment cap declared by this policy.
         """
-        if len(self.transactions_for_learner(learner_id)['transactions']) < self.per_learner_enrollment_limit:
+        if self.per_learner_enrollment_limit is None:
+            return super().can_redeem(learner_id, content_key)
+
+        learner_transactions_count = len(self.transactions_for_learner(learner_id)['transactions'])
+        if learner_transactions_count < self.per_learner_enrollment_limit:
             return super().can_redeem(learner_id, content_key)
 
         return (False, REASON_LEARNER_MAX_ENROLLMENTS_REACHED)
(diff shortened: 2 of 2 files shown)


enterprise_access/apps/subsidy_access_policy/migrations/0008_default_limits_none.py
```diff
@@ -0,0 +1,43 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('subsidy_access_policy', '0007_remove_unused_models'),
+    ]
+
+    operations = [
```
adamstankiewicz, line 12: Changes the `default` for these fields to be `None` instead of `0`.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -375,7 +375,11 @@ def can_redeem(self, learner_id, content_key):
         Checks if the given learner_id has a number of existing subsidy transactions
         LTE to the learner enrollment cap declared by this policy.
         """
-        if len(self.transactions_for_learner(learner_id)['transactions']) < self.per_learner_enrollment_limit:
+        if self.per_learner_enrollment_limit is None:
+            return super().can_redeem(learner_id, content_key)
+
+        learner_transactions_count = len(self.transactions_for_learner(learner_id)['transactions'])
+        if learner_transactions_count < self.per_learner_enrollment_limit:
```
adamstankiewicz, line 382: Previously, the default for `self.per_learner_enrollment_limit` was `0`, which prevents the policy from being redeemed even though this policy intends to have no per-learner limits.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -375,7 +375,11 @@ def can_redeem(self, learner_id, content_key):
         Checks if the given learner_id has a number of existing subsidy transactions
         LTE to the learner enrollment cap declared by this policy.
         """
-        if len(self.transactions_for_learner(learner_id)['transactions']) < self.per_learner_enrollment_limit:
+        if self.per_learner_enrollment_limit is None:
+            return super().can_redeem(learner_id, content_key)
```
pwnage101, line 379: minor nit: It's more intuitive to me if the main branch ends with `super().can_redeem()` at the bottom of the method body, and alternate branches terminate with the error response.  This is similar how we typically structure view functions, where the success case is achieved by reaching the bottom of the function, and error responses are littered in the middle.

enterprise_access/apps/subsidy_access_policy/migrations/0007_remove_unused_models.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ('subsidy_access_policy', '0006_group_uuid_optional'),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name='CappedEnrollmentLearnerCreditAccessPolicy',
+        ),
+        migrations.DeleteModel(
+            name='SubscriptionAccessPolicy',
```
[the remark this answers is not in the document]
pwnage101: LGTM, I remember these getting removed in heart, but not necessarily in practice.

enterprise_access/apps/api/v1/views.py
```diff
@@ -1241,7 +1241,9 @@ def can_redeem(self, request, enterprise_customer_uuid=None):
             can_redeem_for_content_response = {
                 "content_key": content_key,
                 "redemptions": redemptions,
-                "subsidy_access_policy": serialized_policy,
+                "has_successful_redemption": any(redemption['state'] == 'committed' for redemption in redemptions),
```
pwnage101, line 1244: Hm, "committed" is a magic string, but enterprise-access doesn't install openedx-ledger (which gives us openedx_ledger.models.TransactionStateChoices) so maybe just set this at the top of the file?

```python
TRANSACTION_STATE_COMMITTED = "committed"
```

pwnage101 (approved):
Just nits!

enterprise_access/apps/subsidy_access_policy/models.py
adamstankiewicz: Refactored and will push up a new commit for re-review.

.gitignore
```diff
@@ -52,6 +52,8 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
+pii_report
+reports
```
adamstankiewicz, line 56: Generated after running `make validate` in `make app-shell`.

enterprise_access/apps/subsidy_access_policy/constants.py
```diff
@@ -55,3 +55,26 @@ class PolicyTypes:
...
+        Indicates that the content has been redeemed, and a reference to the redemption result (often an enrollment ID)
+        is stored in the reference_id field of the transaction.
+
+    FAILED
+        Indidcates that the attempt to redeem the content in the target LMS encountered an error.
+    """
+
+    CREATED = 'created'
+    PENDING = 'pending'
+    COMMITTED = 'committed'
```
adamstankiewicz, line 79: Only "committed" is used in this code base right now, but if we're bringing in one of the states, might as well bring them all in so they're available.

PR #140, opened 2023-04-26
fix: account for active flag on redeemable policies

adamstankiewicz:
The `subsidy-access-policy` app considers inactive (i.e., `active=False`) policies as redeemable. For example, calling the `http://localhost:18270/api/v1/policy/enterprise-customer/a067aacd-667d-4bd6-81e2-e332ef5ca1cd/can-redeem/?content_key=course-v1%3AedX%2BS2023%2B1T2023` endpoint, which has a policy marked as `active=False` that is otherwise redeemable, returns the inactive policy, which is not correct given `active=False`.

To fix this, we first check the generic `can_redeem` defined in the extended `SubsidyAccessPolicy` model before checking the policy-specific `can_redeem` business rules.

This ensures policies with `active=False` don't attempt to validate their own business (i.e., having to make API calls to `enterprise-subsidy`, etc.) if the policy isn't active to begin with.

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -20,6 +20,7 @@ from enterprise_access.apps.subsidy_access_policy.constants import CREDIT_POLICY
 
 POLICY_LOCK_RESOURCE_NAME = "subsidy_access_policy"
 
+REASON_POLICY_NOT_ACTIVE = "Subsidy access policy is not active."
 REASON_CONTENT_NOT_IN_CATALOG = "Requested content_key not contained in policy's catalog."
 REASON_LEARNER_NOT_IN_ENTERPRISE = "Learner not part of enterprise associated with the access policy."
 REASON_NOT_ENOUGH_VALUE_IN_SUBSIDY = "Not enough remaining value in subsidy to redeem requested content."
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -375,17 +378,21 @@ def can_redeem(self, learner_id, content_key):
         Checks if the given learner_id has a number of existing subsidy transactions
         LTE to the learner enrollment cap declared by this policy.
         """
-        has_per_learner_enrollment_limit = self.per_learner_enrollment_limit is not None
+        # perform generic access checks
+        should_attempt_redemption, reason = super().can_redeem(learner_id, content_key)
+        if not should_attempt_redemption:
+            return (False, reason)
 
+        has_per_learner_enrollment_limit = self.per_learner_enrollment_limit is not None
```
iloveagent57, line 386: I like how you've switched this around to a "fall into success" pattern!

iloveagent57 (approved):
One additional thought: it was worth filtering out the inactive policies at the API layer?
https://github.com/openedx/enterprise-access/blob/face38ff11c4bd1aed52f5611ff08b60ebcfdc90/enterprise_access/apps/api/v1/views.py#L936-L938

enterprise_access/apps/subsidy_access_policy/models.py
adamstankiewicz: Agreed, per the recommendation of @pwnage101! :)

adamstankiewicz:
> One additional thought: it was worth filtering out the inactive policies at the API layer?

@iloveagent57 Oh, good callout. I agree, I think it makes to sense to filter for `active=True` where you suggested (pushed up a new commit for this). I think it also makes sense to leave the validation in the `can_redeem` method of `SubsidyAccessPolicy`, too, though for any instance of checking redeem-ability outside of the API, so I'll leave that in place.

PR #141, opened 2023-04-30
feat: update response payload for can-redeem

adamstankiewicz:
Changes the `http://localhost:18270/api/v1/policy/enterprise-customer/{enterprise_customer_uuid}/can-redeem/?content_key=XXX` API endpoint's response payload in order to return the redeemable subsidy access policy (if any) for each provided content key as well as a top-level redeemable subsidy access policy (i.e., first redeemable policy, if any) that would apply to the parent course (versus a policy for a specific content key. Used by Learner Portal course page to determine whether a course may be subsidized at all, without the MFE needing to iterate through the list of policies to find one that has `can_redeem: true`):

**Original**

```js
[
  {
    "content_key": "course-v1:demox+1234+2T2023_1",
    "redemptions": [],
    "subsidy_access_policy": {
      "uuid": "56744a36-93ac-4e6c-b998-a2a1899f2ae4",
      "policy_redemption_url": "http://policy-redemption.url",
      // ...
    },
    "can_redeem": true,
    "reasons": []
  }
]
```

**New**

```js
{
  "redeemable_subsidy_access_policy": {
    "uuid": "56744a36-93ac-4e6c-b998-a2a1899f2ae4"
    // ...
  },
  "redeemability_by_content_key": [
    {
      "content_key": "course-v1:demox+1234+2T2023_1",
      "redemptions": [],
      "subsidy_access_policy": {
        "uuid": "56744a36-93ac-4e6c-b998-a2a1899f2ae4",
        "policy_redemption_url": "http://policy-redemption.url",
        // ...
      },
      "can_redeem": true,
      "reasons": []
    }
  ]
}
```

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_views.py
@@ -1804,26 +1804,32 @@ class TestSubsidyAccessPolicyRedeemViewset(TestSubsidyRequestViewSet):
             'content_key': ['course-v1:edX+edXPrivacy101+3T2020', 'course-v1:edX+edXPrivacy101+3T2020_2'],
         }
         response = self.client.get(self.subsidy_access_policy_can_redeem_endpoint, query_params)
-        response_list = response.json()
+        response_dict = response.json()
 
         # Make sure we got responses for all two content_keys requested.
-        assert len(response_list) == 2
+        assert len(response_dict["redeemability_per_content_key"]) == 2
+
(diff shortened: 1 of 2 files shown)


adamstankiewicz:
Ended up going with a different approach where we use the course run key associated with `advertised_course_run_uuid` on the course object from course-discovery's `/api/v1/courses/` API endpoint in the learner portal. This endpoint isn't aware of the top-level course to be able to surface the `advertised_course_run_uuid`.

PR #145, opened 2023-05-02
feat: lay down some drf-spec basics. ENT-6948

iloveagent57:
It's spec-tac-u-lahhh!

To test:
- checkout this branch
- make sure devserver reloads
- Visit http://localhost:18270/api-docs/ (swagger docs at existing route, but powered by drf-spec now, so some things work a little nicer)
- Visit http://localhost:18270/api/schema/redoc/, observe the glory

image

I've tidy-d up the `can_redeem` docstring for now to render a little cleaner - it'll need some follow-up work to get its complicated/nested response serialization represented with a serializer class.

### DIFF
--- a/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
@@ -8,8 +8,8 @@ from contextlib import suppress
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
-from django.http.request import QueryDict
 from django_filters.rest_framework import DjangoFilterBackend
+from drf_spectacular.utils import extend_schema, extend_schema_view
 from edx_enterprise_subsidy_client import EnterpriseSubsidyAPIClient
 from edx_rbac.mixins import PermissionRequiredMixin
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
@@ -42,6 +42,16 @@ from .utils import PaginationWithPageCount
 logger = logging.getLogger(__name__)
 
 
+@extend_schema_view(
+    retrieve=extend_schema(
+        tags=['Subsidy Access Policy CRUD'],
+        summary='Retrieve subsidy access policy.',
+    ),
+    delete=extend_schema(
+        tags=['Subsidy Access Policy CRUD'],
+        summary='Delete subsidy access policy.',
+    ),
+)
 class SubsidyAccessPolicyCRUDViewset(PermissionRequiredMixin, viewsets.ModelViewSet):
     """
      Viewset for Subsidy Access Policy CRUD operations.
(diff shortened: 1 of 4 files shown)


johnnagro (approved):
praise be 🙌

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -400,6 +441,11 @@ def get_redemptions_by_policy_uuid(self, enterprise_customer_uuid, lms_user_id,
 
         return redemptions_by_policy_uuid
 
+    @extend_schema(
+        tags=['Subsidy Access Policy Redemption'],
+        summary='Retrieve redemption.',
+        parameters=[serializers.SubsidyAccessPolicyRedemptionRequestSerializer],
```
pwnage101, line 447: why use the `parameters` kwarg instead of `request`?

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -419,6 +465,11 @@ def redemption(self, request, *args, **kwargs):
             status=status.HTTP_200_OK,
         )
 
+    @extend_schema(
+        tags=['Subsidy Access Policy Redemption'],
+        summary='Can redeem.',
+        parameters=[serializers.SubsidyAccessPolicyCanRedeemRequestSerializer],
```
pwnage101, line 471: I did a weird thing with this request serializer to inject the enterprise_customer_uuid URL location parameter into the serializer.  That probably makes the redoc result of this incorrectly represent enterprise_customer_uuid as a query parameter.  Probably what I should have done was inject it as a bit of serializer context or something other than a parameter...maybe you can do that here?

pwnage101: oh wait, how the heck does it work (correctly appears as a path parameter): 
image

pwnage101: no wait, enterprise_customer_uuid appears twice, the first one is probably inferred magically, and the second one borrowed from the serializer.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
iloveagent57: request seems to “want” to use the serializer you give it to describe the signature of the request body, which we’re typically not doing for many of our GET views.
If we were filtering in the views, we could rely on the view’s filterset_class …we might be able to lightly refactor some existing views to do that instead.
the parameters kwarg of extend_schema lets you pass in OpenApiParameter() objects and optionally a serializer class or instance from which fields will be treated as the available query params.
We can circle back to the finer points of this discussion once we eliminate any of the superfluous views that currently exist for access policy - I expect that the list() and credits_available() methods actually aren’t needed, but I have to re-read the spec and chat with Adam and Marlon about exactly what’s needed.

PR #150, opened 2023-05-08
feat: add `list_price` to response of can_redeem

pwnage101:
This also bundles other improvements:
* output serializers are defined for the can_redeem endpoint, enabling much higher quality redoc pages
* added a ContentKeyField to replace the ValidateContentKeyMixin which was harder to use and overly complicated.

ENT-7137

### DIFF
--- a/enterprise_access/apps/api/serializers/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/serializers/subsidy_access_policy.py
@@ -97,59 +97,52 @@ class SubsidyAccessPolicyCRUDSerializer(serializers.ModelSerializer):
         return attrs
 
 
-class ValidateContentKeyMixin:
+class ContentKeyField(serializers.CharField):
     """
-    Mixin to provide validation function for the `content_key` serializer field.
-
-    Requires the inheriting serializer to declare a `content_key` field.  Supports one or many content keys (CharField
-    or ListField-of-CharField).
+    Serializer Field for Content Keys, created to add basic opaque-keys-based validation.
     """
-    def validate_content_key(self, value):
-        """
-        Validate `content_key` field.
-        """
-        content_keys = [value] if isinstance(value, str) else value
-        for content_key in content_keys:
-            try:
-                CourseKey.from_string(content_key)
-            except InvalidKeyError as exc:
-                raise serializers.ValidationError(f"Invalid content_key: {content_key}") from exc
-        return value
+    def to_internal_value(self, data):
+        content_key = data
+        try:
+            CourseKey.from_string(content_key)
+        except InvalidKeyError as exc:
+            raise serializers.ValidationError(f"Invalid content_key: {content_key}") from exc
+        return super().to_internal_value(content_key)
 
 
 # pylint: disable=abstract-method
-class SubsidyAccessPolicyRedeemRequestSerializer(ValidateContentKeyMixin, serializers.Serializer):
+class SubsidyAccessPolicyRedeemRequestSerializer(serializers.Serializer):
     """
     Request Serializer to validate policy redeem endpoint POST data.
 
     For view: SubsidyAccessPolicyRedeemViewset.redeem
     """
     lms_user_id = serializers.IntegerField(required=True)
-    content_key = serializers.CharField(required=True)
+    content_key = ContentKeyField(required=True)
     metadata = serializers.JSONField(required=False, allow_null=True)
 
 
 # pylint: disable=abstract-method
-class SubsidyAccessPolicyListRequestSerializer(ValidateContentKeyMixin, serializers.Serializer):
+class SubsidyAccessPolicyListRequestSerializer(serializers.Serializer):
     """
     Request Serializer to validate policy list endpoint query params.
 
     For view: SubsidyAccessPolicyRedeemViewset.list
     """
     lms_user_id = serializers.IntegerField(required=True)
-    content_key = serializers.CharField(required=True)
+    content_key = ContentKeyField(required=True)
     enterprise_customer_uuid = serializers.UUIDField(required=True)
 
 
 # pylint: disable=abstract-method
-class SubsidyAccessPolicyRedemptionRequestSerializer(ValidateContentKeyMixin, serializers.Serializer):
+class SubsidyAccessPolicyRedemptionRequestSerializer(serializers.Serializer):
     """
     Request Serializer to validate policy redemption endpoint query params.
 
     For view: SubsidyAccessPolicyRedeemViewset.redemption
     """
     lms_user_id = serializers.IntegerField(required=True)
-    content_key = serializers.CharField(required=True)
+    content_key = ContentKeyField(required=True)
     enterprise_customer_uuid = serializers.UUIDField(required=True)
 
 
@@ -165,18 +158,14 @@ class SubsidyAccessPolicyCreditsAvailableRequestSerializer(serializers.Serialize
 
 
 # pylint: disable=abstract-method
-class SubsidyAccessPolicyCanRedeemRequestSerializer(ValidateContentKeyMixin, serializers.Serializer):
+class SubsidyAccessPolicyCanRedeemRequestSerializer(serializers.Serializer):
     """
     Request serializer to validate can_redeem endpoint query params.
 
     For view: SubsidyAccessPolicyRedeemViewset.can_redeem
     """
-    # enterprise_customer_uuid = serializers.UUIDField(
-    #     required=False,
-    #     help_text='The enterprise customer UUID under which policies will be queried for redeemability.',
-    # )
     content_key = serializers.ListField(
-        child=serializers.CharField(required=True),
+        child=ContentKeyField(required=True),
         allow_empty=False,
         help_text='Content keys about which redeemability will be queried.',
     )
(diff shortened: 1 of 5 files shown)


enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -566,22 +591,41 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
                 for redemption in redemptions
             )
-            can_redeem_for_content_response = {
+
+            # Determine the price for content for display purposes only.
+            try:
+                content_metadata = self.subsidy_client.get_subsidy_content_data(enterprise_customer_uuid, content_key)
+                # Note that the "content_price" key is guaranteed to exist, but the value may be None.  Also, the
+                # upstream field type is a string representing a float, and we passthrough that verbatim.
+                content_price = content_metadata["content_price"]
```
pwnage101, line 610: FYI, the code as written just passes-through a string representing decimal dollars (which is how the list_price is stored in the content metadata upstream), rather than integer cents used in the rest of this service, including in other API responses.  My personal opinion would be to have this convert the output to be consistent with the rest of the quantities returned by this API, but I also was just referencing the API spec: https://github.com/openedx/enterprise-access/blob/6ad37aac9b3fbdf594c2409fd8b7ad8aa7ab3555/docs/decisions/0006-api-specification-for-enterprise-mfes.rst#L1

Open to suggestions!

pwnage101, line 610: FYI, I just looked at the API spec and discovered that integer cents was called for, not a passthrough string representing decimal dollars. https://github.com/openedx/enterprise-access/blob/6ad37aac9b3fbdf594c2409fd8b7ad8aa7ab3555/docs/decisions/0006-api-specification-for-enterprise-mfes.rst#L107

I plan to convert this before merging.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -566,22 +591,41 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
                 "reasons": reasons,
             }
-            response.append(can_redeem_for_content_response)
+            element_responses.append(element_response)
 
-        return Response(response, status=status.HTTP_200_OK)
+        response_serializer = serializers.SubsidyAccessPolicyCanRedeemResponseSerializer(
+            {"results": element_responses}
+        )
+        return Response(response_serializer.data, status=status.HTTP_200_OK)
```
iloveagent57, line 631: Nice, way easier to reason about everything wrapped up in here now.

iloveagent57 (approved):
lgtm, I'll take another look once you do that stuff with `list_price`.

pwnage101:
see 2nd commit for changes.  changed the list_price output to an integer, and also refactored the way of mocking the API to get around the lru_cache persisting between tests.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -566,22 +591,43 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
+                "redeemable_subsidy_access_policy": resolved_policy,
+                "can_redeem": bool(resolved_policy),
                 "reasons": reasons,
             }
-            response.append(can_redeem_for_content_response)
+            element_responses.append(element_response)
 
-        return Response(response, status=status.HTTP_200_OK)
+        response_serializer = serializers.SubsidyAccessPolicyCanRedeemResponseSerializer(
+            {"results": element_responses}
```
adamstankiewicz, line 631: This is technically a breaking change to update the response from a list to a dict with a `results` list, as the client (learner portal MFE) currently expects a list to be returned. If we change this, it will likely result in a JavaScript error when parsing its response in the learner portal course page on stage.

A couple links to learner portal source where we parse the `can-redeem` response:
* https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/2e8ebf2beac823651019338352dd2841d7b860c2/src/components/course/data/hooks.jsx#L552-L557
* https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/2e8ebf2beac823651019338352dd2841d7b860c2/src/components/course/data/hooks.jsx#L567

image

A few things to consider: 
1. be OK with potentially breaking stage learner portal course page until it is updated to reflect this change shortly after (will not impact production).
2. version the endpoint.
3. update learner portal course page to handle both style of responses and then remove the old way.
4. something else?

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -566,22 +591,43 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
             )
-            can_redeem_for_content_response = {
+
+            # Determine the price for content for display purposes only.
+            try:
+                content_metadata = self.subsidy_client.get_subsidy_content_data(enterprise_customer_uuid, content_key)
+                # Note that the "content_price" key is guaranteed to exist, but the value may be None.  Also, the
+                # upstream field type is a string representing a decimal dollar.
+                list_price_decimal_dollars_str = content_metadata["content_price"]
+                list_price_integer_cents = int(float(list_price_decimal_dollars_str) * 100)
```
adamstankiewicz, line 612: [curious] Is it worthwhile to return both the integer cents and the decimal dollars? It feels a bit odd for the API to convert the list price to cents only for the caller of the API (learner portal) to convert it back to decimal dollars. A potential idea:

```js
{
  "list_price": {
    "USD": 299,
    "USD_CENTS": 29900,
  }
}
```

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -96,59 +96,52 @@ def validate(self, attrs):
         return attrs
 
 
-class ValidateContentKeyMixin:
+class ContentKeyField(serializers.CharField):
```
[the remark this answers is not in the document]
adamstankiewicz: +1 this is a nice addition :)

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
adamstankiewicz: Related, locally, I'm seeing a course price of $299.00 in course-discovery be returned as `299000` in the `can-redeem` response, which would equate to `$29,900`.

It looks like `get_subsidy_content_data` is already returning integer cents?

adamstankiewicz, line 612: If `list_price_decimal_dollars_str` can be `None`, won't an exception be raised here: `float(list_price_decimal_dollars_str)`?

```python
>>> float(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: float() argument must be a string or a number, not 'NoneType'
```

The try/except doesn't catch `TypeError` though, so I believe this exception would be raised.

pwnage101: handled case, and added new test.

pwnage101: I think I would agree about being explicit about the type, since the unit in this case doesn't directly pertain to any unit of any particular ledger/subsidy.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: Good catch, I forgot to mention this.  I re-worked the endpoint to revert it back to original behavior, with ample code comments to explain why we chose to deviate from the industry norm.  Also, I re-worked it in a way that the redoc page looks accurate:
image

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: updated to match your proposed format.  Note that `usd` is a float, and `usd_cents` is an integer.

adamstankiewicz (approved):
LGTM!

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -248,6 +250,21 @@ class SubsidyAccessPolicyRedeemViewset(UserDetailsFromJwtMixin, PermissionRequir
     permission_required = 'requests.has_learner_or_admin_access'
     http_method_names = ['get', 'post']
 
+    @classmethod
+    @functools.lru_cache(maxsize=None)
+    def get_subsidy_client(cls):
+        """
+        A request-cached EnterpriseSubsidyAPIClient instance.
```
iloveagent57, line 257: FWIW, I think `lru_cache` will cache for the lifetime of the process, not the request lifetime.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -626,22 +608,57 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
+                    list_price_integer_cents = int(list_price_decimal_dollars * 100)
+                else:
+                    list_price_decimal_dollars = None
+                    list_price_integer_cents = None
+            except requests.exceptions.HTTPError:
+                # No need to record a failure reason here because evaluate_policies() -> policy.can_redeem() already
+                # caught any problematic or non-existent content keys, and provided a reason it could not be redeemed.
+                # Just set a None content_price and move along.
+                list_price_decimal_dollars = None
+                list_price_integer_cents = None
```
iloveagent57, line 640: This method body is getting really long - is there an obvious place to wrap this in a helper method or something?  Or maybe the serializer can do the translation of string to numbers?

iloveagent57 (approved):
Looks good, two small notes.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: I'd love to consolidate this into the serializer, but in light of my long todo list I'm going to defer this and add a comment.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: Interesting, that does seem to match up with my experience with these having lasting memory between unit tests, which in theory should destroy and create new instances of these views and models.  I'll update all the docstrings for now to say that they're process-cached since I think there's enough of a combination of sentiment and evidence to support the claim.  Note that `@cached_property` will have the same behavior since it's effectively backed by `lru_cache`.

PR #154, opened 2023-05-11
Aed/crud view tests

iloveagent57:


enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -23,15 +29,309 @@
...
+        (
+            {'system_wide_role': 'some-other-role', 'context': str(TEST_ENTERPRISE_UUID)},
+            status.HTTP_403_FORBIDDEN,
+        ),
+        # A good admin role, but in a context/customer we don't have, get's you a 403.
+        (
+            {'system_wide_role': SYSTEM_ENTERPRISE_ADMIN_ROLE, 'context': str(uuid4())},
+            status.HTTP_403_FORBIDDEN,
+        ),
+        # A learner role is also fine, but in a context/customer we don't have, get's you a 403.
```
johnnagro, line 85: if the customer is missing - should we offer a 404 instead? makes reasoning client-side easier?

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -23,15 +29,309 @@
...
+        (
+            {'system_wide_role': SYSTEM_ENTERPRISE_LEARNER_ROLE, 'context': str(uuid4())},
+            status.HTTP_403_FORBIDDEN,
+        ),
+        # An operator role is fine, too, but in a context/customer we don't have, get's you a 403.
```
johnnagro, line 90: if the customer is missing - should we offer a 404 instead? makes reasoning client-side easier?

johnnagro (approved):
one non-blocking discussion point

PR #155, opened 2023-05-11
[devstack] provision a SystemWideEnterpriseUserRoleAssignment

pwnage101:
This automates a currently-manual step during devstack setup.

### DIFF
--- a/provision-enterprise-access.sh
+++ b/provision-enterprise-access.sh
@@ -30,6 +30,11 @@ docker exec -t enterprise_access.app bash -c "echo 'from django.contrib.auth imp
 echo -e "${GREEN}Provisioning ${name}_worker in LMS...${NC}"
 docker exec -t edx.devstack.lms  bash -c "source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms --settings=devstack_docker manage_user ${name}_worker ${name}_worker@example.com --staff --superuser"
 
+# Create system wide enterprise role assignment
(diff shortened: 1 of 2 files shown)


provision-enterprise-access.sh
```diff
@@ -30,6 +30,11 @@ docker exec -t enterprise_access.app bash -c "echo 'from django.contrib.auth imp
 echo -e "${GREEN}Provisioning ${name}_worker in LMS...${NC}"
 docker exec -t edx.devstack.lms  bash -c "source /edx/app/edxapp/edxapp_env && python /edx/app/edxapp/edx-platform/manage.py lms --settings=devstack_docker manage_user ${name}_worker ${name}_worker@example.com --staff --superuser"
 
+# Create system wide enterprise role assignment
+# TODO: this is a pretty complex oneline, we should probably eventually convert this to a management command.
```
adamstankiewicz, line 34: I believe this will also be relevant to other Enterprise IDAs as well? I think you have to manually create this system-wide role assignment for every Enterprise IDA, I think. E.g., `enterprise-catalog` doesn't have this created during provisioning either.

Of course, probably not something to prioritize right now, but worth pursuing at some point.

adamstankiewicz (approved):

provision-enterprise-access.sh
pwnage101: indeed, if we like this particular oneline, we could repeat it for enterprise-subsidy, enterprise-catalog, license-manager, and more?

PR #156, opened 2023-05-11
docs: [snowflake] Take transaction state into consideration for balance calculation

pwnage101:
Correctly account for the state of Transactions and Reversals to enhance Subsidy balance calculation to calculate both "final" and "pending" balance.  Note this only impacts the demonstration/reference SQL query, not anything that actually gets run.

### DIFF
--- a/scripts/lc2_generate_test_data/snowflake.sql
+++ b/scripts/lc2_generate_test_data/snowflake.sql
@@ -92,15 +92,17 @@ order by subsidy.title, transaction.created
 
 -- calculate balance of subsidy A (expect it to be $851 = $1000 - $49 - $49 + $49):
 with all_quantities as (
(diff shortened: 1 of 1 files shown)


scripts/lc2_generate_test_data/snowflake.sql
```diff
@@ -69,18 +69,32 @@ order by subsidy.title, transaction.created
 order by subsidy.title, reversal.created
 ;
 
--- list OCM enrollments created for testing:
+-- List OCM enrollments created for testing.  Link the following chain of models:
```
pwnage101, line 72: This is just a bonus for @alex-sheehan-edx

iloveagent57 (approved):
Good idea!

PR #158, opened 2023-05-13
fix: list price on `can-redeem` API endpoint

adamstankiewicz:
Current implementation assumes the upstream `content_price` from subsidy will be a string like `"299.00"`; however, the `content_metadata` [API endpoint](https://enterprise-subsidy.edx.org/api/schema/redoc/#tag/api/operation/api_v1_content_metadata_retrieve) in subsidy returns USD cents like `29900`.

This PR updates the `list_price` API field in enterprise-access's `can-redeem` API response to return the correct denominations of USD and USD cents, e.g.:

```json
[{
	"list_price": {
		"usd": 299,
		"usd_cents": 29900
	}
}]
```

instead of the current state of saying the course is $29,900 USD:

```json
[{
	"list_price": {
		"usd": 29900,
		"usd_cents": 2990000
	}
}]
```

### DIFF
--- a/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
@@ -669,7 +669,7 @@ class SubsidyAccessPolicyRedeemViewset(UserDetailsFromJwtMixin, PermissionRequir
                 list_price_integer_cents = content_metadata["content_price"]
                 # TODO: simplify this function by consolidating this conversion logic into the response serializer:
                 if list_price_integer_cents is not None:
-                    list_price_decimal_dollars = float(list_price_integer_cents / 100)
+                    list_price_decimal_dollars = float(list_price_integer_cents) / 100
                 else:
                     list_price_decimal_dollars = None
             except requests.exceptions.HTTPError:
--- a/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
@@ -665,16 +665,13 @@ class SubsidyAccessPolicyRedeemViewset(UserDetailsFromJwtMixin, PermissionRequir
             # Determine the price for content for display purposes only.
             try:
                 content_metadata = self.subsidy_client.get_subsidy_content_data(enterprise_customer_uuid, content_key)
-                # Note that the "content_price" key is guaranteed to exist, but the value may be None.  Also, the
-                # upstream field type is a string representing a decimal dollar.
-                list_price_decimal_dollars_str = content_metadata["content_price"]
+                # Note that the "content_price" key is guaranteed to exist, but the value may be None.
+                list_price_integer_cents = content_metadata["content_price"]
                 # TODO: simplify this function by consolidating this conversion logic into the response serializer:
-                if list_price_decimal_dollars_str:
-                    list_price_decimal_dollars = float(list_price_decimal_dollars_str)
-                    list_price_integer_cents = int(list_price_decimal_dollars * 100)
+                if list_price_integer_cents is not None:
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -665,16 +665,13 @@ def can_redeem(self, request, enterprise_customer_uuid):
             # Determine the price for content for display purposes only.
             try:
                 content_metadata = self.subsidy_client.get_subsidy_content_data(enterprise_customer_uuid, content_key)
-                # Note that the "content_price" key is guaranteed to exist, but the value may be None.  Also, the
-                # upstream field type is a string representing a decimal dollar.
```
adamstankiewicz, line 669: > Also, the upstream field type is a string representing a decimal dollar.

This doesn't appear to be true. The `content-metadata` API in subsidy returns USD cents: https://enterprise-subsidy.edx.org/api/schema/redoc/#tag/api/operation/api_v1_content_metadata_retrieve

```json
{
    "content_uuid": "eeeb2cad-7e7b-4efa-a61e-e42b75751e70",
    "content_key": "edX+DM",
    "course_run_uuid": "d178f6d4-cd7a-4c6d-b1c8-f8099433173a",
    "course_run_key": "course-v1:edX+DM+1T2023",
    "source": "2u",
    "mode": "paid-executive-education",
    "content_price": 81800,
    "geag_variant_id": "7a3334ce-c587-4757-8c8c-d5fe9f6ed88b"
}
```

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -369,8 +365,9 @@ def resolve_policy(cls, redeemable_policies):
         """
         # For now, we inefficiently make one call per subsidy record.
         subsidy_uuids = set(redeemable_policy.subsidy_uuid for redeemable_policy in redeemable_policies)
+        subsidy_client = get_versioned_subsidy_client()
         subsidy_balances = {
-            subsidy_uuid: cls.subsidy_client.retrieve_subsidy(subsidy_uuid)["current_balance"]
+            subsidy_uuid: subsidy_client.retrieve_subsidy(subsidy_uuid)["current_balance"]
```
adamstankiewicz, line 370: Without making this change to get the versions subsidy client (i.e., `cls.subsidy_client`), I'm seeing getting a "No attribute `retrieve_subsidy` exists on `subsidy_client`" type of error.

I can replicate this in the Django shell, where using the class method approach to access the `subsidy_client` property errors but using the `subsidy_client` method on a single instance of `SubsidyAccessPolicy` does have the `retrieve_subsidy` attribute.

Is there an alternative way of fixing this?

iloveagent57: You can decorate the property definition with a `@staticmethod` (and remove the `self` parameter from it).

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
iloveagent57: todo for our future selves: we should add some better docstrings to the subsidy client package to make this clearer.

iloveagent57 (approved):

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: Ah, for some reason I read the enterprise-subsidy code and thought it was a passthrough.  However, it seems to already contain conversion logic: https://github.com/openedx/enterprise-subsidy/blob/837fdcde75d7b0450506f637f410785a7d6f3d3e/enterprise_subsidy/apps/content_metadata/api.py#L54

pwnage101 (approved):

PR #161, opened 2023-05-15
feat: can_redeem() now checks against spend_limit. ENT-7153

iloveagent57:
https://2u-internal.atlassian.net/browse/ENT-7153


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -235,6 +246,21 @@ def transactions_for_learner_and_content(self, lms_user_id, content_key):
...
+        the ``spend_limit`` set by this policy.
+        """
+        if self.spend_limit is None:
+            return False
+
+        spent_amount = self.aggregates_for_policy().get('total_quantity') or 0
+        course_price = self.subsidy_client.get_subsidy_content_data(
+            self.enterprise_customer_uuid,
+            content_key
+        )['content_price']
```
adamstankiewicz, line 261: [clarification] Will `course_price` always be a value that we can use to add to `spent_amount`? E.g., if `get_subsidy_content_data` returns `content_price: None` without getting an error like:

```
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

johnnagro (approved):

PR #169, opened 2023-05-16
fix: reverting request-id logger changes

johnnagro:
reverting the logger component until i can figure out why splunk is not picking up what i'm putting down.

- https://github.com/openedx/enterprise-access/pull/162
- https://github.com/openedx/enterprise-access/pull/164
- https://github.com/openedx/enterprise-access/pull/166
- https://github.com/openedx/enterprise-access/pull/167


adamstankiewicz (approved):
:sadpanda:

PR #172, opened 2023-05-17
refactor: use settings url for get_policy_redemption_url

emrosarioa:
The `get_policy_redemption_url` method will return the proper URL based on configuration settings instead of the current request scheme.


enterprise_access/apps/api/tests/test_serializers.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        Test that the get_policy_redemption_url method returns the correct
+        URL for the policy redemption.
+        """
+
+        serializer = SubsidyAccessPolicyRedeemableResponseSerializer(self.non_redeemable_policy)
+
+        data = serializer.data
+        self.assertIn("policy_redemption_url", data)
+        expected_url = urljoin(settings.ENTERPRISE_ACCESS_URL, self.subsidy_access_policy_redeem_endpoint)
```
iloveagent57, line 37: nit: might be worth declaring this as the exact string literal you expect, to make sure that we don't accidentally get double slashes between the base url and the redeem API route, as sometimes happens with `urljoin()`.

iloveagent57 (approved):

enterprise_access/settings/devstack.py
```diff
@@ -79,6 +79,7 @@
 ENTERPRISE_ADMIN_PORTAL_URL = 'http://localhost:1991'
 ENTERPRISE_CATALOG_URL = 'http://enterprise.catalog.app:18160'
 ENTERPRISE_SUBSIDY_URL = 'http://enterprise-subsidy.app:18280'
+ENTERPRISE_ACCESS_URL = 'http://enterprise-access.app:18270'
```
adamstankiewicz, line 82: Because this URL is being used in the API response for the client to use as a web request, `enterprise-access.app` won't work as that the internal devstack network url (i.e., for service-to-service communication). In this case, we need the URL to be `http://localhost:18270` for the MFE to be able to make the request.

Running this branch locally, and clicking "Enroll" button in learner portal to call the API at `http://enterprise-access.app:18270` throws an error.

adamstankiewicz (changes requested):

adamstankiewicz (approved):

PR #178, opened 2023-05-22
feat: cache content metadata and containment

iloveagent57:
https://2u-internal.atlassian.net/browse/ENT-7199
Features:
* Adds a `content_metadata_api` module to encapsulate the fetching and caching of content metadata and catalog inclusion.
* The `SubsidyAccessPolicy` is modified to interact with this module for content needs, rather than going directly to the catalog or subsidy client.
* Introduces optional kwarg to `can_redeem()` to skip the check for learner inclusion in a customer.  We now call `can_redeem` with that kwarg set to True from the view, which is being requested by an already authN'd user, who a priori must be a member of the customer.
* Adds optional settings `CACHE_KEY_VERSION_STAMP` and `CONTENT_METADATA_CACHE_TIMEOUT` to control key-based cache eviction and timeouts for cached content metadata, respectively.
* I've tried to update the tests to be slightly more sane in terms of all the mocking of integration points, but it's an uphill battle.
* Subtle, one-line win: `resolve_policy` can early return if there's only one policy to find the priority of.  This saves a subsidy API balance fetch.

Notes:
* We had previously considered eliminating the call to check catalog inclusion, in favor of relying on the subsidy `content-metadata` calls, where 404 response to infer that the content is not in the catalog.  However, the `content-metadata` endpoint provided by the subsidy service (and ultimately by the enterprise-catalog service) returns metadata records as long as they are associated with _any_ catalog for the provided customer uuid.  We’ll have cases where one customer has multiple catalogs, so we still have to do a catalog inclusion check.
* In the future, we could modify the [customer content-metadata endpoint](https://github.com/openedx/enterprise-catalog/blob/f0e625dc2007776b464f872d199dbe4f607da686/enterprise_catalog/apps/api/v1/views/enterprise_customer.py#L183) in enterprise-catalog to take an optional catalog identifier by which the response should be filtered, which could get us down to 1 catalog call.
* For now, I've gone in the direction of relying on django-caching + request caching (thanks, `TieredCache`) to make the redeemability checks more performant.

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/utils.py
+++ b/enterprise_access/apps/subsidy_access_policy/utils.py
@@ -2,7 +2,13 @@
 Utils for subsidy_access_policy
 """
 from django.conf import settings
+from edx_django_utils.cache import RequestCache
 from edx_enterprise_subsidy_client import get_enterprise_subsidy_api_client
+
+from enterprise_access import __version__ as code_version
+
+CACHE_KEY_SEP = ':'
+CACHE_NAMESPACE = 'subsidy_access_policy'
 
 
 def get_versioned_subsidy_client():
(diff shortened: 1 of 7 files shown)


enterprise_access/apps/subsidy_access_policy/utils.py
```diff
@@ -14,3 +20,23 @@ def get_versioned_subsidy_client():
     if getattr(settings, 'ENTERPRISE_SUBSIDY_API_CLIENT_VERSION', None):
         kwargs['version'] = int(settings.ENTERPRISE_SUBSIDY_API_CLIENT_VERSION)
     return get_enterprise_subsidy_api_client(**kwargs)
+
+
+def versioned_cache_key(*args):
```
macdiesel, line 25: Good stuff.

macdiesel (approved):
Nice shimming.

PR #179, opened 2023-05-22
fix: can_redeem - fix invalid enterprise 4xx responses.

pwnage101:
In some cases, these returned 500 errors, but now they're all 4xx.

ENT-7111

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
@@ -954,3 +954,59 @@ class TestSubsidyAccessPolicyRedeemViewset(APITestWithMocks):
         assert response.json() == {
             'detail': f'Could not determine price for content_key: {test_content_key}',
         }
+
+    @mock.patch('enterprise_access.apps.api.v1.views.subsidy_access_policy.LmsApiClient')
+    @ddt.data(
+        # A role that's not mapped to any feature perms will get you a 403.
+        (
+            {'system_wide_role': 'some-other-role', 'context': str(TEST_ENTERPRISE_UUID)},
+            status.HTTP_403_FORBIDDEN,
+        ),
+        # Real learners requesting a fake enterprise get 403 since edx-rbac stops the request.
+        (
+            {'system_wide_role': SYSTEM_ENTERPRISE_LEARNER_ROLE, 'context': TEST_ENTERPRISE_UUID},
+            status.HTTP_403_FORBIDDEN,
+        ),
+        # Real enterprise admins requesting a fake enterprise get 403 since edx-rbac stops the request.
+        (
+            {'system_wide_role': SYSTEM_ENTERPRISE_ADMIN_ROLE, 'context': TEST_ENTERPRISE_UUID},
+            status.HTTP_403_FORBIDDEN,
+        ),
+        # Real all-access operators requesting a fake enterprise get 404 since our own view code stops the request.
+        (
+            {'system_wide_role': SYSTEM_ENTERPRISE_OPERATOR_ROLE, 'context': '*'},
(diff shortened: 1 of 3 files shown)


enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -954,3 +954,67 @@ def test_can_redeem_policy_no_price(self, mock_lms_client):
         assert response.json() == {
             'detail': f'Could not determine price for content_key: {test_content_key}',
         }
+
+    @mock.patch('enterprise_access.apps.api.v1.views.SubsidyAccessPolicyRedeemViewset.subsidy_client')
```
iloveagent57, line 958: My latest commit got rid of this client, you'll want to mock something like this instead:
```python
        mocked_content_data_from_view = {
            "content_uuid": str(uuid4()),
            "content_key": test_content_key,
            "source": "edX",
            "content_price": None,
        }

        with mock.patch(
            'enterprise_access.apps.api.v1.views.subsidy_access_policy.get_and_cache_content_metadata',
            return_value=mocked_content_data_from_view,
        ):
            # ...yer codez
```
Your mock doesn't _have_ to be a context manager, I just liked that style to sort of mentally distinguish the mock of the price fetch in the view vs. the mock of the price fetch down in the `can_redeem()` logic.

pwnage101: turns out I didn't even need to mock that function since the code doesn't get that far anyway.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -652,8 +652,12 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
             if not redemptions and not redeemable_policies:
                 lms_client = LmsApiClient()
-                enterprise_customer_data = lms_client.get_enterprise_customer_data(enterprise_customer_uuid)
-                enterprise_admin_users = enterprise_customer_data.get('admin_users')
+                try:
+                    enterprise_customer_data = lms_client.get_enterprise_customer_data(enterprise_customer_uuid)
+                    enterprise_admin_users = enterprise_customer_data.get('admin_users')
+                except requests.exceptions.HTTPError as exc:
+                    logger.warning(f"enterprise_customer {enterprise_customer_uuid} not found.")
+                    raise NotFound(detail="Could not find given enterprise_customer.") from exc
```
iloveagent57, line 660: I know it's unlikely that this call fails in isolation - is it worth failing the whole request if we can't get the admin users?  Could we try and throw a warning in the logs on exception, but continue on with the rest of the logic in this view?

emrosarioa: I like this idea!

pwnage101: The reason I opted to fail to whole request is because this error does not pertain to any one particular requested course_key.  It is an issue of the UUID in the URL location not existing.  If we could spare an extra API call, I would have moved this check to the top of the view, but that would incur performance cost every time the view is called rather than just in error cases.

iloveagent57: Cool, fair point.  I think https://github.com/openedx/enterprise-access/pull/183 kind of accomplishes the "check if anything exists for the customer UUID" bit (albeit by virtue of checking for an active policy, not necessarily the existence of a customer).

iloveagent57 (approved):

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: indeed, that does elegantly solve the problem.

PR #183, opened 2023-05-24
feat: return 404 from can-redeem when no active policies for customer

iloveagent57:
Looking for feedback on whether 404 is appropriate, vs. maybe a 200 with something in the `reasons` list?
The `NotFound` in this PR does come with a helpful detail:
image


pwnage101 (approved):

pwnage101:
I think 404 is fine because policies are perhaps the main object that can_redeem is concerned with.

johnnagro:
folded into another PR

PR #188, opened 2023-05-26
feat: concurrency experimentation

johnnagro:
One approach to concurrently collecting redeem-ability information about a list of course keys is to implement the concurrency via a celery task group, and wait to collect the results. This is easier than rolling your own ThreadPool or ThreadPool-like concurrency management. It also requires less changing our existing logic - which would be required if we were batching up the API requests in parallel. It may seem odd to use celery here but it is a fully-featured, configurable concurrency framework.


johnnagro:
not going to pursue this right now

PR #191, opened 2023-05-30
feat: create and send an idempotency key to subsidy redeem()

pwnage101:
ENT-7204

All related PRs:
* https://github.com/openedx/enterprise-access/pull/191
* https://github.com/openedx/enterprise-subsidy/pull/113
* https://github.com/openedx/edx-enterprise-subsidy-client/pull/23

### DIFF
--- a/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
@@ -320,7 +320,7 @@ class SubsidyAccessPolicyRedeemViewset(UserDetailsFromJwtMixin, PermissionRequir
         all_policies_for_enterprise = self.get_queryset()
         for policy in all_policies_for_enterprise:
             try:
-                redeemable, reason = policy.can_redeem(lms_user_id, content_key, skip_customer_user_check=True)
+                redeemable, reason, _ = policy.can_redeem(lms_user_id, content_key, skip_customer_user_check=True)
             except ContentPriceNullException as exc:
                 logger.warning(f'{exc} when checking can_redeem() for {enterprise_customer_uuid}')
                 raise RedemptionRequestException(
@@ -454,15 +454,21 @@ class SubsidyAccessPolicyRedeemViewset(UserDetailsFromJwtMixin, PermissionRequir
             # point, we should also consider migrating this logic into the policy model so that different policy types
             # that have different locking needs can supply different lock kwargs.
             with policy.lock():
-                if policy.can_redeem(lms_user_id, content_key):
-                    redemption_result = policy.redeem(lms_user_id, content_key, metadata)
+                can_redeem, reason, existing_transactions = policy.can_redeem(lms_user_id, content_key)
+                if can_redeem:
+                    redemption_result = policy.redeem(lms_user_id, content_key, existing_transactions, metadata)
                     send_subsidy_redemption_event_to_event_bus(
                         SUBSIDY_REDEEMED.event_type,
                         serializer.data
                     )
                     return Response(redemption_result, status=status.HTTP_200_OK)
                 else:
-                    raise RedemptionRequestException()
+                    raise RedemptionRequestException(
+                        detail=self._get_reasons_for_no_redeemable_policies(
+                            policy.enterprise_customer_uuid,
+                            {reason: [policy]}
+                        )
+                    )
         except SubsidyAccessPolicyLockAttemptFailed as exc:
             logger.exception(exc)
             raise SubsidyAccessPolicyLockedException() from exc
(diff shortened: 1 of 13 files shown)


enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -454,15 +454,19 @@ def redeem(self, request, *args, **kwargs):
             # point, we should also consider migrating this logic into the policy model so that different policy types
             # that have different locking needs can supply different lock kwargs.
             with policy.lock():
-                if policy.can_redeem(lms_user_id, content_key):
-                    redemption_result = policy.redeem(lms_user_id, content_key, metadata)
+                can_redeem, reason, existing_transactions = policy.can_redeem(lms_user_id, content_key)
+                if can_redeem:
+                    historical_redemptions = [r.uuid for r in existing_transactions if r.state == 'committed']
```
johnnagro, line 459: hmm i think for the sake of idempotency, we need all historical transactions - not just successful ones.

johnnagro (changes requested):

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
iloveagent57: ...or maybe have your key-generating utility function do the sorting so the caller doesn't have to think too hard?

pwnage101: oh i typed 'failed' when I meant to type 'pending' in my force-push.  Fixing again.  Once I get unit tests up this should be covered too.

enterprise_access/apps/subsidy_access_policy/utils.py
```diff
@@ -40,3 +50,16 @@ def request_cache():
...
+    """
+    Create a key that allows a transaction to be created idempotently.
+    """
+    idpk_data = {
+        tx_key: value
+        for tx_key, value in metadata.items()
+        if tx_key in TRANSACTION_METADATA_KEYS
+    }
+    hashed_metadata = hashlib.md5(str(idpk_data).encode()).hexdigest()
+    return f'{LEDGERED_SUBSIDY_IDEMPOTENCY_KEY_PREFIX}-{subsidy_uuid}-{hashed_metadata}'
```
pwnage101, line 65: Is it cool that I'm dropping the quantity from the transaction idempotency key sent by enterprise-access?  That context just doesn't exist at the point of calling the subsidy redeem API endpoint from access, and logically I don't think it's needed for de-duplicating requests anyway

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -454,15 +454,22 @@ def redeem(self, request, *args, **kwargs):
             # point, we should also consider migrating this logic into the policy model so that different policy types
             # that have different locking needs can supply different lock kwargs.
             with policy.lock():
-                if policy.can_redeem(lms_user_id, content_key):
-                    redemption_result = policy.redeem(lms_user_id, content_key, metadata)
+                can_redeem, reason, existing_transactions = policy.can_redeem(lms_user_id, content_key)
+                if can_redeem:
+                    historical_redemptions = [r.uuid for r in existing_transactions if r.state != 'failed']
```
adamstankiewicz, line 459: nit: should 'failed' be using `TransactionStateChoices.FAILED`?

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: I'm re-opening this thread due to developing a fairly strong opinion that "failed" and "reversed" transactions should version an idempotency key.  Please see full explanation in docstring for `_redemptions_for_idempotency_key()`.  Thoughts?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -287,48 +288,101 @@ def will_exceed_spend_limit(self, content_key, content_metadata=None):
...
+
+        return (True, None, existing_transactions)
+
+    def _redemptions_for_idempotency_key(self, all_transactions):
+        """
+        Select the historical redemptions (transactions) that may contribute to the idempotency key.
+
+        Currently, only failed or reversed transactions qualify.  That means for every new failed/reversed transaction
+        in the history, a new transaction with a unique idempotency_key can be created even for the same
+        subsidy/policy/learner/content.  In all other cases, the same idempotency_key as the last redeem attempt will be
```
iloveagent57, line 333: I would just add something here like: "thus allowing us to re-attempt a previously failed or reversed transaction for that same (subsidy, policy, learner, content) combination."

iloveagent57 (approved):

PR #198, opened 2023-06-07
fix: account for fact that transaction quantities are almost always negative

iloveagent57:
https://2u-internal.atlassian.net/browse/ENT-7262

Before this fix, learner-level and policy-level spend limits were not enforced because a sum of negative numbers will always be less than a spend cap defined as a positive integer.

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -273,6 +273,19 @@ class SubsidyAccessPolicy(TimeStampedModel):
             'aggregates': response_payload['aggregates'],
         }
 
+    @staticmethod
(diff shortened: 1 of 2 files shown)


alex-sheehan-edx (approved):
Do we want a test that checks `can_redeem` will throw an exception?

adamstankiewicz (approved):

iloveagent57:
well, here's a copypasta of the test since github PRs are currently real slow to update:
```python
def test_content_would_exceed_limit_positive_spent_amount(self):
        """
        Ensures that passing a positive spent_amount will raise an exception.
        """
        with self.assertRaisesRegex(Exception, 'Expected a sum of transaction quantities <= 0'):
            self.per_learner_enroll_policy.content_would_exceed_limit(10, 100, 15)
```

pwnage101 (approved):

PR #201, opened 2023-06-09
feat: overall improvements to custom policy fields

pwnage101:
* Create policy endpoint responds with 400 error on missing custom field.
* Create policy endpoint responds with 400 error on extraneous custom field.
* Redoc now has all help_texts inherited from the model fields.
* Custom policy field descrpitions are more helpful.
* Custom policy fields are more maintainable.

ENT-7113


enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -87,13 +107,28 @@ def create(self, validated_data):
...
+                f"Missing fields for {policy_type} policy type: {policy_class.REQUIRED_CUSTOM_FIELDS}."
+            )
+        # Here's the "no more" part:
+        unused_custom_fields = set(policy_class.ALL_CUSTOM_FIELDS) - set(policy_class.REQUIRED_CUSTOM_FIELDS)
+        if unused_custom_fields.intersection(attrs.keys()):
+            custom_policy_field_errors.append(
+                f"Extraneous fields for {policy_type} policy type: {list(unused_custom_fields)}."
+            )
+        if custom_policy_field_errors:
+            raise serializers.ValidationError(custom_policy_field_errors)
```
iloveagent57, line 130: Would it be simpler to do something like
```python
if set(attrs.keys()) != set(policy_class.REQUIRED_CUSTOM_FIELDS):
    errors.append(f"Provided fields {attrs.keys()} does not exactly match the required custom fields {policy_class.REQUIRED_CUSTOM_FIELDS}")
```

iloveagent57 (approved):
Looks good, minor code suggestion!

enterprise_access/apps/api/serializers/subsidy_access_policy.py
pwnage101: attrs.keys() likely contains a bunch of other non-custom fields such as "enterprise_customer_uuid" and "catalog_uuid" though.   That said, if I consolidated this code to just one error it would be a shorter function.

PR #203, opened 2023-06-20
feat: idempotent policies

brobro10000:
Checks existing active policies with identical customer uuids, subsidy uuids, catalog uuids and access method and returns the policy if it exist, otherwise create a new policy.

pending tests

### DIFF
--- a/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
@@ -145,6 +145,13 @@ class SubsidyAccessPolicyCRUDViewset(PermissionRequiredMixin, viewsets.ModelView
     # impossible for learners to create policies.
     permission_required = REQUESTS_ADMIN_LEARNER_ACCESS_PERMISSION
 
+    def __init__(self, *args, **kwargs):
+        self.extra_context = {}
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -145,6 +145,13 @@ class SubsidyAccessPolicyCRUDViewset(PermissionRequiredMixin, viewsets.ModelView
     # impossible for learners to create policies.
     permission_required = REQUESTS_ADMIN_LEARNER_ACCESS_PERMISSION
 
+    def __init__(self, *args, **kwargs):
+        self.extra_context = {}
+        super().__init__(*args, **kwargs)
+
+    def set_idempotent_policy(self, idempotent):
+        self.extra_context['idempotent'] = idempotent
```
pwnage101, line 153: The function name seems ambiguous to me.  Like, isn't the policy always idempotent, why do we have to call a function to make it idempotent?  IMO:
```suggestion
    def set_policy_created(self, created):
        self.extra_context['created'] = created
```
then, above, call it with True and False in both branches.

pwnage101 (changes requested):
just one thing about naming, the rest LGTM!

pwnage101 (approved):

PR #207, opened 2023-06-27
checkpoint

iloveagent57:


enterprise_access/apps/api/v1/exceptions.py
```diff
@@ -0,0 +1,143 @@
...
+                self.code = constants.SubsidyRedemptionErrorCodes.FULFILLMENT_ERROR
+
+        self.user_message = constants.SubsidyFulfillmentErrorReasons.USER_MESSAGES_BY_REASON.get(self.reason)
+
+    def _get_subsidy_fulfillment_error_message(self, subsidy_error_detail):
+        """
+        ``subsidy_error_detail`` is either a string describing an error message,
+        a dict with a "message" key describing an error message, or a list of such
+        dicts.  This helper method widdles any of those things down into a single
+        error message string.
```
pwnage101, line 117: Is this just the result of the subsidy API being indecisive about error formatting and being written by >1 person?

macdiesel: Oof yes, can we fix the offending call?

iloveagent57: I wrote to be defensive about the types of messages that can exist from a DRF APIException: https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py

PR #209, opened 2023-06-30
docs: cache overview doc

iloveagent57:


macdiesel (approved):
This looks good. Is it worth adding some more detail as our logic on why we decided to cache these items and why for that amount of time?

PR #214, opened 2023-07-12
fix: history tracking for SubsidyAccessPolicy proxy models

pwnage101:
Before this commit, history tracking on SubsidyAccessPolicy and all child models was non-functional. Now, all history is now tracked in the pre-existing `historicalsubsidyaccesspolicy` table, including changes to all child proxy models of SubsidyAccessPolicy.

At the time of writing, django-simple-history does not support tracking proxy models natively.  This upstream issue contains a few workarounds, and we incorporated one of them:

https://github.com/jazzband/django-simple-history/issues/544

One downside is that it creates migration clutter.  The migration file included in this PR contains only no-ops, literally sqlmigrate returns `BEGIN; COMMIT;`.  As far as I can tell, this causes no harm.

NOTE: This does not retroactively track creation and update records. For historical integrity, we should at least consider running this command to give the current instances initial change records:

```
python manage.py populate_history --auto
```

https://django-simple-history.readthedocs.io/en/latest/quick_start.html#existing-projects

ENT-7150


iloveagent57 (approved):
This is magical AF and I love-hate it.

PR #215, opened 2023-07-12
refactor: demote admins and make permission names more explicit

pwnage101:
Before this change, enterprise admins could access policy API endpoints which should have been exclusive to openedx operators.  Now, they have been demoted to learner-level policy permissions.

Also, make all the feature role and permission names more explicit.

ENT-7150


iloveagent57 (approved):
Good catch!

PR #220, opened 2023-07-18
feat: support update/partial_update in SubsidyAccessPolicyViewSet.

iloveagent57:
ENT-7149

We want to refactor the policy update API endpoint in a way that makes the permission-ing more obvious and clearly defines the signature of the API using distinct request and response serializers.
This PR changes the SubsidyAccessPolicyViewSet to define `update()` and `parital_update()` actions.
Note, this viewset is already hooked up to a router at the `/api/v1/subsidy-access-policies/` path.

The following fields are allowed to be updated:
- description
- catalog_uuid
- subsidy_uuid
- access_method
- per_learner_enrollment_limit
- per_learner_spend_limit
- spend_limit

Requesting to modify any other field will result in a `ValidationError`.
We use the existing `SubsidyAccessPolicyResponseSerialize`r to output the updated policy record.

Permissions: only the system wide roles that ultimately map to the `SUBSIDY_ACCESS_POLICY_WRITE_PERMISSION` permission are allowed to update policy records.

### DIFF
--- a/enterprise_access/apps/api/filters/base.py
+++ b/enterprise_access/apps/api/filters/base.py
@@ -29,6 +29,6 @@ class NoFilterOnRetrieveBackend(drf_filters.DjangoFilterBackend):
         """
         Returns None if this is a retrieve() operation.
         """
-        if view.action == 'retrieve':
+        if view.action in ('retrieve', 'update', 'partial_update'):
             return None
         return super().get_filterset_class(view, queryset)
--- a/enterprise_access/apps/api/serializers/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/serializers/subsidy_access_policy.py
@@ -17,6 +17,23 @@ from enterprise_access.apps.subsidy_access_policy.models import SubsidyAccessPol
 logger = logging.getLogger(__name__)
 
 
+def validate_no_extra_custom_fields_for_policy_class(policy_class, fields_and_values):
+    """
+    Validates that no fields are included in the serializer
+    data pertaining to one type of SubsidyAccessPolicy that are
+    only valid for *another* type of SubsidyAccessPolicy.
+    e.g. ``per_learner_spend_limit`` is not a valid field for creating
+    or updating a ``PerLearnerEnrollmentSubsidyAccessPolicy`` record.
+    """
+    unused_custom_fields = set(policy_class.ALL_CUSTOM_FIELDS) - set(policy_class.REQUIRED_CUSTOM_FIELDS)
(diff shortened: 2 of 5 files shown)


enterprise_access/apps/api/filters/base.py
```diff
@@ -29,6 +29,6 @@ def get_filterset_class(self, view, queryset=None):
         """
         Returns None if this is a retrieve() operation.
         """
-        if view.action == 'retrieve':
+        if view.action in ('retrieve', 'update', 'partial_update'):
             return None
```
pwnage101, line 33: outdated docstring

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -239,6 +239,81 @@ class SubsidyAccessPolicyDeleteRequestSerializer(serializers.Serializer):
...
+
+    def validate(self, attrs):
+        """
+        Raises a ValidationError if any field not explicitly declared
+        as a field in this serializer defintion is provided as input.
+        """
+        unknown = sorted(set(self.initial_data) - set(self.fields))
+        if unknown:
+            raise serializers.ValidationError("Field(s) are not updatable: {}".format(", ".join(unknown)))
+        return attrs
```
pwnage101, line 306: Does this need to copy/inherit any logic to check that "no more" custom fields are provided than allowed for the policy type given?

https://github.com/openedx/enterprise-access/blob/888b77d40fd666b39c85a7283a31b4f7c78b6c02/enterprise_access/apps/api/serializers/subsidy_access_policy.py#L141-L146

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -352,6 +359,107 @@ def test_destroy_view(self, request_payload, expected_change_reason):
...
+            spend_limit=5,
+            active=False,
+        )
+
+        request_payload = {
+            'description': 'the new description',
+            'active': True,
+            'catalog_uuid': str(uuid4()),
+            'subsidy_uuid': str(uuid4()),
+            'access_method': 'assigned',
```
pwnage101, line 384: may as well use `AccessMethods.ASSIGNED`?

pwnage101 (approved):
LGTM besides nits!

enterprise_access/apps/api/filters/base.py
iloveagent57: fixed in the other PR: https://github.com/openedx/enterprise-access/pull/221/files

PR #223, opened 2023-07-24
feat: Adds additional error reasons to SubsidyAccessPolicy

brobro10000:
Adds more reasons and an additional error message for learner credit in the `SubsidyAccessPolicy` model for the `can_redeem` endpoint. 

Additional logic needed within the enterprise-subsidy repository to specify if a subsidy is not active by checking `isActive` within the subsidy `can_redeem` endpoint

Pending Tests

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -21,8 +21,9 @@ from .constants import (
     REASON_LEARNER_MAX_SPEND_REACHED,
     REASON_LEARNER_NOT_IN_ENTERPRISE,
     REASON_NOT_ENOUGH_VALUE_IN_SUBSIDY,
-    REASON_POLICY_NOT_ACTIVE,
+    REASON_POLICY_EXPIRED,
     REASON_POLICY_SPEND_LIMIT_REACHED,
+    REASON_SUBSIDY_EXPIRED,
     AccessMethods,
     TransactionStateChoices
 )
@@ -311,6 +312,7 @@ class SubsidyAccessPolicy(TimeStampedModel):
 
         content_price = self.get_content_price(content_key, content_metadata=content_metadata)
         spent_amount = self.aggregates_for_policy().get('total_quantity') or 0
+
         return self.content_would_exceed_limit(spent_amount, self.spend_limit, content_price)
 
     def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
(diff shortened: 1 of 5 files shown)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -321,10 +323,14 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
...
-                * third a list of any transactions represending existing redemptions (any state).
+                * third a list of any transactions representing existing redemptions (any state).
         """
+        content_metadata = self.get_content_metadata(content_key)
+
         if not self.active:
-            return (False, REASON_POLICY_NOT_ACTIVE, [])
+            if not self.will_exceed_spend_limit(content_key, content_metadata=content_metadata):
+                return (False, REASON_POLICY_NO_FUNDS, [])
+            return (False, REASON_POLICY_EXPIRED, [])
```
brobro10000, line 333: Will be abstracted to handle an expired policy, and a policy that has no funds

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -340,13 +346,12 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
...
-        if not subsidy_can_redeem_payload.get('can_redeem', False):
-            return (False, REASON_NOT_ENOUGH_VALUE_IN_SUBSIDY, existing_transactions)
-
-        content_metadata = self.get_content_metadata(content_key)
         if not content_metadata:
             return (False, REASON_CONTENT_NOT_IN_CATALOG, existing_transactions)
 
+        if not subsidy_can_redeem_payload.get('can_redeem', False):
+            return (False, REASON_NOT_ENOUGH_VALUE_IN_SUBSIDY, existing_transactions)
+
```
brobro10000, line 354: Logic swapped to avoid additional call into the subsidies `can_redeem`. Check if the `content_metadata` exist in the catalog before checking on a subsidy level.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -321,32 +324,48 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
...
+        active_subsidy = subsidy_can_redeem_payload.get('active', False)
+        existing_transactions = subsidy_can_redeem_payload.get('all_transactions', [])
+
+        # inactive subsidy
+        if not active_subsidy:
+            return (False, REASON_SUBSIDY_EXPIRED, [])
+
+        # inactive policy with spend remaining (could have redeemed)
+        if not self.active and not self.will_exceed_spend_limit(content_key, content_metadata=content_metadata):
```
macdiesel, line 344: This and the changes on like 149 are a little sus i think. I synced with @brobro10000 on slack with some suggestions. He's going to take another look at the logic in here.

brobro10000: Going to take another look at it after our discussion yesterday 👍🏽

macdiesel (approved):
I think this makes the logic around the error handling easier to read. Nice work.

PR #224, opened 2023-07-26
feat: change which reasons are selected for a non-redeemable policy

brobro10000:
Reorders the messaging from can_redeem to match figmas expected order of operation for error messaging return


pwnage101 (commented):
I think if the order of the reasons has an impact, we should document the rationale of the ordering in the docstring of this function

pwnage101 (approved):
LGTM!

PR #226, opened 2023-07-27
fix: when we fetch inter-service data in can_redeem matters for perf

iloveagent57:
We should wait to fetch data from enterprise-subsidy until we know we actually need it.

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -333,15 +333,9 @@ class SubsidyAccessPolicy(TimeStampedModel):
                 * second element contains a reason code if the content is not redeemable,
                 * third a list of any transactions representing existing redemptions (any state).
         """
-        content_metadata = self.get_content_metadata(content_key)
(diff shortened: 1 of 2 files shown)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -352,20 +346,31 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
...
-        # inactive subsidy
+        # We want to wait to do these checks that might require a call
+        # to the enterprise-subsidy service until we *know* we'll need the data.
+        subsidy_can_redeem_payload = self.subsidy_client.can_redeem(
+            self.subsidy_uuid,
+            lms_user_id,
+            content_key,
+        )
+
+        active_subsidy = subsidy_can_redeem_payload.get('active', False)
```
adamstankiewicz, line 367: [suggestion/nit] Consider adding a code comment that this `active` refers to a computed property that takes into account start/expiration dates so there's less confusion between the subsidy `active` value versus the policy's `is_active` flag?

adamstankiewicz (approved):
LGTM. Are there test cases to update/add to account for these changes?

PR #230, opened 2023-08-01
build: switch to mysql 8

johnnagro:
## Description
- https://2u-internal.atlassian.net/browse/ENT-7470
- standing on the shoulders of giants https://github.com/openedx/enterprise-subsidy/pull/141

### Testing instructions

```
make dev.up
make dev.dbcopy8
make app-restart
```
Then login to the admin site to ensure your data was copied: http://localhost:18270/admin/ 
Note that the new mysql 8 container is running side-by-side with mysql 5.7 for now.  For mysql 8, the container's 3306 port is bound to the host's 3406 port, so use 3406 if you use a mysql IDE or something locally.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,20 @@ services:
     volumes:
       - enterprise_access_mysql57:/var/lib/mysql
 
+  mysql80:
+    command: mysqld --character-set-server=utf8 --collation-server=utf8_general_ci
+    container_name: enterprise_access.mysql80
+    environment:
+      MYSQL_ROOT_PASSWORD: ""
+      MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
+    # Oracle-packaged version includes a `linux/arm64/v8` version, needed for
+    # machines with Apple Silicon CPUs (Mac M1, M2)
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -11,6 +11,22 @@ services:
...
+    environment:
+      MYSQL_ROOT_PASSWORD: ""
+      MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
+    # Oracle-packaged version includes a `linux/arm64/v8` version, needed for
+    # machines with Apple Silicon CPUs (Mac M1, M2)
+    image: mysql:8.0.33-oracle
+    networks:
+      - devstack_default
+    ports:
+      - "3406:3306"
```
iloveagent57, line 26: You know, we should actually just not bind this here or in enterprise-subsidy, or choose a different host port for each.  Bringing containers up as-is results in "Bind for 0.0.0.0:3406 failed: port is already allocated".  I think I only bound a port in the enterprise-subsidy PR because I was copying some prior art from devstack.

iloveagent57 (approved):
I'll open an enterprise-subsidy PR to not bind to the host's 3406 port.

PR #237, opened 2023-08-15
feat: add new field in credits_available serializer

sameenfatima78:


jajjibhai008 (approved):
LGTM 👍

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -210,6 +210,12 @@ def remaining_balance(self):
         """
         return self.subsidy_balance()
 
+    def subsidy_expiration_datetime(self):
+        """
+        Returns the expiration date for the associated subsidy ledger.
+        """
+        return self.subsidy_record().get('expiration_datetime')
```
iloveagent57, line 217: It's probably worth storing the result of `self.subsidy_record()` in the request cache at this point.  The serializer above now calls `obj.remaining_balance()` and `obj.subsidy_expiration_datetime()`, each of which will result in an API call to fetch the subsidy record associated with the policy object.  We should add a function like `get_and_cache_subsidy_record(subsidy_uuid)` to the `subsidy_api.py` file and call that new caching function from `self.subsidy_record()`.

PR #239, opened 2023-08-17
feat: Add new content_assignments app and models stubs

pwnage101:
ENT-7528


enterprise_access/apps/content_assignments/models.py
```diff
@@ -0,0 +1,99 @@
...
+        AssignmentPolicy,
+        related_name="assignments",
+        on_delete=models.CASCADE,
+        db_index=True,
+        help_text="AssignmentPolicy defining the lifecycle rules of this assignment.",
+    )
+    learner_email = models.EmailField(
+        null=False,
+        blank=False,
+        unique=True,
```
iloveagent57, line 45: This can't be unique, a single email might be assigned multiple contents.

iloveagent57: We'll want to index this field, too.

enterprise_access/apps/content_assignments/models.py
```diff
@@ -0,0 +1,99 @@
...
+        blank=False,
+        null=False,
+        choices=LearnerContentAssignmentStateChoices.CHOICES,
+        default=LearnerContentAssignmentStateChoices.ALLOCATED,
+        help_text=(
+            "The current state of the LearnerContentAssignment. One of: "
+            f"{[choice[1] for choice in LearnerContentAssignmentStateChoices.CHOICES]}"
+        ),
+    )
+    transaction = models.UUIDField(
```
iloveagent57, line 83: Nit: probably `transaction_uuid`, since it's not a FK to another model.

enterprise_access/apps/content_assignments/models.py
iloveagent57: Doesn't have to be addressed in this PR, but we'll also eventually want a composite UNIQUE key on (email, content) or (email, content, state), depending on the approach we pick for dealing with the allocated -> errored -> (admin re-assigns) -> allocated lifecycle flow.

pwnage101: whoops, carryover from the place I copied from

enterprise_access/apps/content_assignments/models.py
```diff
@@ -0,0 +1,99 @@
...
+from .constants import LearnerContentAssignmentStateChoices
+
+
+class AssignmentPolicy(TimeStampedModel):
+    """
+    Manage the creation and lifecycle of LearnerContentAssignments according to configurable rules.
+    """
+    subsidy_access_policy = models.ForeignKey(
+        SubsidyAccessPolicy,
+        related_name="assignment_policy",
```
pwnage101, line 19: Also just updated this from "assignment_policy_config" to "assignment_policy".

enterprise_access/apps/content_assignments/models.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+    state = models.CharField(
+        max_length=255,
+        blank=False,
+        null=False,
+        choices=LearnerContentAssignmentStateChoices.CHOICES,
+        default=LearnerContentAssignmentStateChoices.ALLOCATED,
+        help_text=(
+            "The current state of the LearnerContentAssignment. One of: "
+            f"{[choice[0] for choice in LearnerContentAssignmentStateChoices.CHOICES]}"
```
pwnage101, line 82: Also updated this from choice[1]  to choice[0] so that the actual database values are shown instead of the pretty names.

enterprise_access/apps/content_assignments/models.py
```diff
@@ -0,0 +1,101 @@
...
+from django.db import models
+from django_extensions.db.models import TimeStampedModel
+from simple_history.models import HistoricalRecords
+
+from enterprise_access.apps.subsidy_access_policy.models import SubsidyAccessPolicy
+
+from .constants import LearnerContentAssignmentStateChoices
+
+
+class AssignmentPolicy(TimeStampedModel):
```
iloveagent57, line 13: Nit: We've been using a convention of throwing a `uuid` primary key on most of our new model definitions for a while, could you throw that on these two models instead of the auto big int?

iloveagent57 (approved):
👍 IMHO, this is good to go once you switch the `id` PK to `uuid`.

enterprise_access/apps/content_assignments/models.py
pwnage101: ah, of course

PR #240, opened 2023-08-17
feat: multi-policy resolution

johnnagro:
## Description
- https://2u-internal.atlassian.net/browse/ENT-7484
- when multiple redeemable policies are present, resolve them first by type, then expiration date, then balance
- utilize `RequestCache` to avoid multiple round-trips to subsidy service
- gate this experimental functionality behind a setting, allowing us to observe performance in a controlled fashion on STAGE


alex-sheehan-edx (approved):
👍 makes sense to me? `¯\_(ツ)_/¯`

iloveagent57 (approved):
LGTM! Is there an existing test case somewhere that the priority is respected above the expiration date and balance?

johnnagro:
> LGTM! Is there an existing test case somewhere that the priority is respected above the expiration date and balance?

I'll add one for priority - preexisting code and i assumed it was there but dont see one.

PR #244, opened 2023-08-22
feat: more policy logging

johnnagro:
- adding some logging to assert when the gate is disabled


alex-sheehan-edx (approved):
idk why your tests are failing but lgtm

PR #247, opened 2023-08-24
feat: expose display_name via REST API. ENT-7595

iloveagent57:
https://2u-internal.atlassian.net/browse/ENT-7595

Builds on top of https://github.com/openedx/enterprise-access/pull/246

### DIFF
--- a/enterprise_access/apps/api/serializers/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/serializers/subsidy_access_policy.py
@@ -43,6 +43,7 @@ class SubsidyAccessPolicyResponseSerializer(serializers.ModelSerializer):
         fields = [
             'uuid',
             'policy_type',
+            'display_name',
             'description',
             'active',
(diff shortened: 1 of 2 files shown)


enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -89,6 +91,12 @@ class Meta:
         read_only_fields = ['uuid']
         extra_kwargs = {
             'uuid': {'read_only': True},
+            'display_name': {
+                'allow_blank': False,
```
adamstankiewicz, line 95: [clarification] If this field has `'allow_blank': False` initially, does this mean the provisioning tool will not work correctly until any corresponding changes are made on the provisioning tool? If that's the case, should we allow blanks on the API side of things for now until the provisioning tool is passing along `display_name` in its payload?

adamstankiewicz (approved):

PR #262, opened 2023-09-13
fix: unsupported operand type for remaining balance

sameenfatima78:
This change handles the scenario when `spent_amount` comes to be `None`.

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -721,14 +721,18 @@ class PerLearnerEnrollmentCreditAccessPolicy(CreditPolicyMixin, SubsidyAccessPol
         return (True, None, existing_redemptions)
 
     def credit_available(self, lms_user_id=None):
-        if self.remaining_balance_per_user(lms_user_id) > 0:
-            return True
-        return False
+        remaining_balance_per_user = self.remaining_balance_per_user(lms_user_id)
+        return (remaining_balance_per_user is not None) and remaining_balance_per_user > 0
 
     def remaining_balance_per_user(self, lms_user_id=None):
         """
         Returns the remaining redeemable credit for the user.
+        Returns None if `per_learner_enrollment_limit` is not set.
         """
+        if self.per_learner_enrollment_limit is None:
+            return None
+        if self.per_learner_enrollment_limit <= 0:
+            return 0
         existing_transaction_count = len(self.transactions_for_learner(lms_user_id)['transactions'])
         return self.per_learner_enrollment_limit - existing_transaction_count
 
@@ -777,12 +781,18 @@ class PerLearnerSpendCreditAccessPolicy(CreditPolicyMixin, SubsidyAccessPolicy):
         return (True, None, existing_redemptions)
 
     def credit_available(self, lms_user_id=None):
-        return self.remaining_balance_per_user(lms_user_id) > 0
+        remaining_balance_per_user = self.remaining_balance_per_user(lms_user_id)
+        return (remaining_balance_per_user is not None) and remaining_balance_per_user > 0
 
     def remaining_balance_per_user(self, lms_user_id=None):
         """
         Returns the remaining redeemable credit for the user.
+        Returns None if `per_learner_spend_limit` is not set.
         """
+        if self.per_learner_spend_limit is None:
+            return None
+        if self.per_learner_spend_limit <= 0:
+            return 0
         spent_amount = self.transactions_for_learner(lms_user_id)['aggregates'].get('total_quantity') or 0
         return self.per_learner_spend_limit - spent_amount
 


saleem-latif (approved):

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -784,7 +784,7 @@ def remaining_balance_per_user(self, lms_user_id=None):
         Returns the remaining redeemable credit for the user.
         """
         spent_amount = self.transactions_for_learner(lms_user_id)['aggregates'].get('total_quantity') or 0
-        return self.per_learner_spend_limit - spent_amount
+        return self.per_learner_spend_limit - int(spent_amount or 0)
```
adamstankiewicz, line 787: [clarification/curious] What situation leads to `spent_amount` being `None`? On line 786, if `self.transactions_for_learner(lms_user_id)['aggregates'].get('total_quantity')` were to return `None`, wouldn't `spent_amount` already be falling back to `0` due to the `or 0` on the previous line?

sameenfatima78: That's what I thought too but I deployed the changes on stage today for learner credit subsidy box and got this [error](https://splunk.edx.org/en-US/app/search/search?sid=1694609260.2240668) from access service. 👀

adamstankiewicz: I see. I think the error is based on `self.per_learner_spend_limit` being `None`, not `spent_amount`. The `per_learner_spend_limit` model field can be `None` if the budget/policy type does not deal with a per-learner spend limit, where the math would be `None - 0`.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -730,7 +730,7 @@ def remaining_balance_per_user(self, lms_user_id=None):
         Returns the remaining redeemable credit for the user.
         """
         existing_transaction_count = len(self.transactions_for_learner(lms_user_id)['transactions'])
-        return self.per_learner_enrollment_limit - existing_transaction_count
+        return int(self.per_learner_enrollment_limit or 0) - existing_transaction_count
```
adamstankiewicz, line 733: nit: Is the `int` necessary versus `(self.per_learner_enrollment_limit or 0) - existing_transaction_count`? I believe both would have the same result.

Same question for the other `per_learner_spend_limit` case below.

adamstankiewicz, line 733: In the case where `per_learner_enrollment_limit` is `None` or the limit itself is `0`, I'm wondering if it conceptually makes sense for this method to return a negative number?

A negative number representing "remaining balance" feels a bit odd, when as implemented, the field really becomes the negative of "how many transactions has this learner has" which doesn't really make any semantic sense as "remaining balance", IMHO.

```python
>>> existing_transaction_count = 10
>>> per_learner_enrollment_limit = None
>>> (per_learner_enrollment_limit or 0) - existing_transaction_count
-10
```

I'm wondering if these methods should instead fail early and `return None` if there is no `self.per_learner_enrollment_limit` or `self.per_learner_spend_limit` set for the budget/policy?

Rationale: when there is no per-learner limit set for the budget, there can't really be a notion of a remaining balance specific to the learner, where it'd instead fallback to looking at the overall budget's remaining balance (not per-learner).

Example:

```python
def remaining_balance_per_user:
  if self.per_learner_enrollment_limit is None:
    return None

  if self.per_learner_enrollment_limit <= 0:
    return 0

  existing_transaction_count = len(self.transactions_for_learner(lms_user_id)['transactions'])
  return self.per_learner_enrollment_limit - existing_transaction_count
```

What do you think?

sameenfatima78: That does make a lot of sense. I have changed the logic according to your suggestion.

sameenfatima78:
@adamstankiewicz Thanks for the suggestion, I have made the changes. Can you please review again? 

Also, do you think it would make sense to move the "credit_available" method to the parent class as their definitions in both child classes are identical?

sameenfatima78:
@adamstankiewicz Reminder to please take a look into this on priority. This change is required by this [PR](https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/810) that is already merged and deployed on stage but I am waiting to deploy it on prod.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -721,14 +721,18 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
         return (True, None, existing_redemptions)
 
     def credit_available(self, lms_user_id=None):
-        if self.remaining_balance_per_user(lms_user_id) > 0:
-            return True
-        return False
+        remaining_balance_per_user = self.remaining_balance_per_user(lms_user_id) or 0
```
adamstankiewicz, line 724: Similar to handling `None` versus `0` differently in the `remaining_balance_per_user` methods, the `None` or `0` should likely be handled as semantically different here as well. If `self.remaining_balance_per_user` returns `None` because there is no `self.per_learner_enrollment_limit` set, I don't believe it's semantically correct to fallback to say the user has a remaining balance of `0`, as there is no per-learner limit (this method would return `False` currently, even when there is no limit set).

Related, are there additional checks that are relevant to determine whether these policies are "available" beyond only checking against the per-learner limits? Would cases such as the following be handled by these `credit_available` methods?

* User hasn't exceeded their individual per-learner limit, but the policy has already exceeded its configured `spend_limit` (not redeemable).
* User hasn't exceeded their individual per-learner limit, but the associated subsidy has no remaining balance.
* User hasn't exceeded their individual per-learner limit, but the policy is no longer active.
* User hasn't exceeded their individual per-learner limit, but the associated subsidy is expired.

These are some of the additional checks that the `can_redeem` methods perform that might be worth referring to. Just verifying that the `credits_available` API is taking into account all factors that could deem a policy available or unavailable.

adamstankiewicz (commented):
> Also, do you think it would make sense to move the "credit_available" method to the parent class as their definitions in both child classes are identical?

Hm, good question. Based on the other review comment about the `credit_available` methods, I think there might be an opportunity to abstract some common/generic checks used across both policy types' `credit_available` method (e.g., did the policy exceed its configured `spend_limit`?), but I might recommend keeping the per-learner checks separate as one's "remaining balance" deals with enrollment counts and the other's "remaining balance" deals with spend (money), which are fundamentally two different constructs that might have divergent behavior.

enterprise_access/apps/subsidy_access_policy/models.py
sameenfatima78: > Similar to handling None versus 0 differently in the remaining_balance_per_user methods, the None or 0 should likely be handled as semantically different here as well.
> 
> If self.remaining_balance_per_user returns None because there is no self.per_learner_enrollment_limit set, I don't believe it's semantically correct to fallback to say the user has a remaining balance of 0, as there is no per-learner limit (this method would return False currently, even when there is no limit set).
> 

@adamstankiewicz Agreed. I have handled that part in the new changes. 

> Related, are there additional checks that are relevant to determine whether these policies are "available" beyond only checking against the per-learner limits? Would cases such as the following be handled by these credit_available methods?
> 
> User hasn't exceeded their individual per-learner limit, but the policy has already exceeded its configured spend_limit (not redeemable).
> User hasn't exceeded their individual per-learner limit, but the associated subsidy has no remaining balance.
> User hasn't exceeded their individual per-learner limit, but the policy is no longer active.
> User hasn't exceeded their individual per-learner limit, but the associated subsidy is expired.
> These are some of the additional checks that the can_redeem methods perform that might be worth referring to. Just verifying that the credits_available API is taking into account all factors that could deem a policy available or unavailable.
> 

@adamstankiewicz I believe the existing code doesn't check any of these conditions. Would it be okay if I create a follow up ticket/PR to address this feedback as all of these checks would require adding unit tests..

sameenfatima78:
@adamstankiewicz Gentle reminder to review this PR and if all looks good, please merge it.

enterprise_access/apps/subsidy_access_policy/models.py
adamstankiewicz: > @adamstankiewicz I believe the existing code doesn't check any of these conditions. Would it be okay if I create a follow up ticket/PR to address this feedback as all of these checks would require adding unit tests..

Yep, totally. Just wanted to call out that the current logic may not be fully taking into account all factors.

adamstankiewicz (approved):
LGTM. :shipit:

sameenfatima78:
@adamstankiewicz Thanks for the approval. Can you also please merge it? I don't seem to have the right access.

adamstankiewicz:
> @adamstankiewicz Thanks for the approval. Can you also please merge it? I don't seem to have the right access.

Merged!

PR #264, opened 2023-09-15
docs: ADR on assignment-based policies

iloveagent57:


docs/decisions/0012-assignment-based-policies.rst
```diff
@@ -0,0 +1,133 @@
...
+*******************************
+
+Status
+======
+
+Accepted - September 2023
+
+Context
+=======
+Some enterprise customers want tighter control over the purse strings, particularly for
```
adamstankiewicz, line 11: nit: What does "purse strings" mean? Is there a way to say the same thing without feeling too slangy?

docs/decisions/0012-assignment-based-policies.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Before determining if new assignments can be allocated for a set of learners,
+the assignment-based policy logic must now also take into account, via the sum of:
+
+- The total cost of new assignments to be allocated.
+- The total cost of all existing allocated (but not accepted, failed, or cancelled) assignments related to the policy.
+- Must not exceed the remaining balance of the policy’s subsidy.
+- Must not exceed the remaining spend limit configured for the policy.
+
+Furthermore, we'll have to modify *redemption* logic for assignment-based policies as follows:
```
adamstankiewicz, line 79: While not _redemption_ logic per se, should we mention any implications to the `credits_available` API? For example, given the `credits_available` API returns all Budgets (policies) for a given customer/learner that are redeemable (agnostic to a piece of content), under what situations would the new assignment policy type be returned by `credits_available` API? E.g., only when an assignment(s) have been accepted/redeemed by the learner or when there is an allocated assignment?

docs/decisions/0012-assignment-based-policies.rst
```diff
@@ -0,0 +1,133 @@
...
+We need to support query-ability of assignment records by course *title*.  We'll start
+replicating content metadata into Django models within the ``enterprise-access`` service.
+This replication will initially happen ad-hoc/on read, and there will be some
+index on *recency* to determine if replicated data should be updated from our upstream
+systems of records.  This design will provide us with the flexibility to eventually move
+from a "pull" model (replicate-on-read) to an event-based "pull" model (update-on-upstream-event).
+
+Rejected alternative: requiring that clients of our APIs provide this title,
+which we'd persist on the ``LearnerContentAssignment`` model.  This becomes
+distasteful from a separation-of-concerns perspective - the domain of *asignments*
```
adamstankiewicz, line 120: nit: typo in assignments.

docs/decisions/0012-assignment-based-policies.rst
iloveagent57: Yeah, good point.  I'd say as long as an assignment-based policy is active, current, and has balance, `credits_available` probably should return it (as long as the learner is linked to the enterprise).

adamstankiewicz: Makes sense to me, so long as the product/UX requirements expects the "Learner Credit" subsidy box to appear on the Learner Portal dashboard page and for the catalog tied to the assignment policy to be included in the Learner Portal search results when the learner hasn't been assigned any content yet.

iloveagent57: Actually, maybe that's not the point of `credits_available`?  Is the purpose to inform the learner portal of budgets with which the learner could redeem *right now*?  If so, then maybe it should only describe budgets with allocated assignments, as you suggested.

iloveagent57: - ``credits_available`` This endpoint returns a list of current, active policies
  with credit availble to be redeemed **immediately**.  Therefore, we'll have to conditionally
  include assignment-based policies in this endpoint's response payload - if the requesting
  learner has an allocated assignment record associated with a current, active policy, that
  policy record should be included in the response payload.

macdiesel (approved):

PR #266, opened 2023-09-18
fix: Added support for django42

zubairshakoorarbisoft:
Issue: https://github.com/openedx/enterprise-access/issues/265

### DIFF
--- a/enterprise_access/apps/api/urls.py
+++ b/enterprise_access/apps/api/urls.py
@@ -4,11 +4,11 @@ Root API URLs.
 All API URLs should be versioned, so urlpatterns should only
 contain namespaces for the active versions of the API.
 """
-from django.conf.urls import include, url
+from django.urls import include, path
 
 from enterprise_access.apps.api.v1 import urls as v1_urls
 
 app_name = 'api'
 urlpatterns = [
-    url(r'^v1/', include(v1_urls)),
+    path('v1/', include(v1_urls)),
 ]
--- a/enterprise_access/apps/core/admin.py
+++ b/enterprise_access/apps/core/admin.py
@@ -2,12 +2,13 @@
 
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
 from djangoql.admin import DjangoQLSearchMixin
 
 from enterprise_access.apps.core.models import User
 
 
+@admin.register(User)
 class CustomUserAdmin(DjangoQLSearchMixin, UserAdmin):
     """ Admin configuration for the custom User model. """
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.8"]
+        django-version: ["pinned", "4.2"]
```
UsamaSadiq, line 15: We've implemented `pinned` logic only in edx-platform to make sure we are testing with the same patch version as production. For other IDAs we don't need such strict testing so you don't need to copy the method from edx-platform here. 
Adding `django42` in toxenv below should suffice for the scope of this repo.

.github/workflows/ci.yml
```diff
@@ -21,7 +23,12 @@ jobs:
     - name: Install requirements
       run: make requirements
     - name: Upgrade packages
-      run: pip install -U pip wheel
+      run: |
```
UsamaSadiq, line 26: Change not needed.

enterprise_access/apps/subsidy_request/tests/test_utils.py
```diff
@@ -26,7 +26,7 @@ def test_get_user_from_request_session(self):
 
         # Create request with session
         request = HttpRequest()
-        middleware = SessionMiddleware()
+        middleware = SessionMiddleware(mock.Mock())
```
UsamaSadiq, line 29: Please add documentation link for this change.

.github/workflows/ci.yml
zubairshakoorarbisoft: I don't think so we need to keep only django42 as we have decided to remove `toxenv` @UsamaSadiq

.github/workflows/ci.yml
zubairshakoorarbisoft: oky removed

enterprise_access/apps/subsidy_request/tests/test_utils.py
zubairshakoorarbisoft: done

.github/workflows/ci.yml
```diff
@@ -21,7 +22,12 @@ jobs:
...
+        pip install -U pip wheel codecov
```
iamsobanjaved, line 26: Kindly remove `codecov` from here.

iamsobanjaved (approved):
One change, otherwise 👍

.github/workflows/ci.yml
UsamaSadiq, line 26:
```suggestion
        pip install -U pip wheel
```

UsamaSadiq: updated.

UsamaSadiq (approved):

PR #268, opened 2023-09-18
feat: unique constraints on LearnerContentAssignments

iloveagent57:
Makes the `LearnerContentAssignment` model unique on (assignment_configuration, learner identifier, content identifier).
This means that only one combination of (learner, content) can exist in a given assignment configuration.  So in the lifecycle of a given assignment, state transitions such as:
      `allocated` -> `cancelled` -> `allocated` -> `accepted`
are allowed, and we can use the history table of this model to ascertain when/why such state transitions occurred.


enterprise_access/apps/content_assignments/models.py
```diff
@@ -76,7 +89,6 @@ class LearnerContentAssignment(TimeStampedModel):
         AssignmentConfiguration,
         related_name="assignments",
         on_delete=models.SET_NULL,
-        db_index=True,
```
iloveagent57, line 79: This is just a superfluous line - the FK includes an index by default.

macdiesel (approved):

PR #269, opened 2023-09-19
feat: django admin for content_assignments

iloveagent57:
also adds an index on `LearnerContentAssignment.state`


enterprise_access/apps/content_assignments/models.py
```diff
@@ -126,6 +126,7 @@ class Meta:
         max_length=255,
         blank=False,
         null=False,
+        db_index=True,
```
johnnagro, line 129: i assume this table is currently small in production so indexing wont take long

johnnagro (approved):

enterprise_access/apps/content_assignments/models.py
iloveagent57: I think it is zero records big.

PR #270, opened 2023-09-19
feat: add API client method to create pending enterprise learners

pwnage101:
ENT-7539


enterprise_access/apps/api_client/lms_client.py
```diff
@@ -142,3 +143,43 @@ def enterprise_contains_learner(self, enterprise_customer_uuid, learner_id):
             logger.exception('Incorrect data received from LMS. [%s]', url)
 
         return result
+
+    def create_pending_enterprise_users(self, enterprise_customer_uuid, user_emails):
```
pwnage101, line 147: This is 95% copied from https://github.com/edx/license-manager/blob/f7a32bc1e06333cad2d6644e7938e3c516b1a63c/license_manager/apps/api_client/enterprise.py#L94

iloveagent57 (approved):

PR #272, opened 2023-09-20
feat: policy validation refactor, policy -> assignment dependency strictness

iloveagent57:
### Support validation at model and serializer level with common rule set
In adding another type of SubsidyAccessPolicy, I was finding it harder to write validation logic around what types of fields
certain policy types *must* have and which they *must not* have, in a way that was consistent across the model level and the DRF serialization level.
Background:
1. It's really nice to have some validation of fields on model instances at the point where the instance has been created/updated in-memory, but not yet saved.  This helps validate model instance manipulation across Django Admin and our business logic layers.
2. DRF _really_ doesn't want to do model-level validation in serializers; it strongly prefers to validate inputs *before* doing any in-memory model instance manipulation.  See https://www.django-rest-framework.org/community/3.0-announcement/#differences-between-modelserializer-validation-and-modelform , arguments confused over this: https://github.com/encode/django-rest-framework/issues/3144 and a blog post from DRF-maintainer about motivations for the decision: https://www.dabapps.com/insights/django-models-and-encapsulation/

So, with (1) and (2) at odds, I tried to introduce a tool where we can define constraints around custom policy model fields (think learner spend limit) in a fairly flexible way, and that's by defining the `FIELD_CONSTRAINTS` dict on each policy class.  There's also some model-base logic to make use of these for *model-level* validation, and a utility function in the policy serializers to use the same data structure for *serializer-level* validation.

This approach differs from our previous approach, which was more along the lines of "deny anything not explicitly listed in the allow list for this policy type".  `FIELD_CONSTRAINTS` makes use of python function to be able to say both "you must have this field" and "this field must be null" for the same type of policy.

### Policies strictly depend on assignments
We previously implemented the API for assignment configuration such that they depended on a policy record existing before a configuration could be created in the API.  Modifying the validation logic above highlighted the fact that this dependency direction is actually the _reverse_ of the direction we're trying to maintain of policies -> assignment configs.
This PR updates the assignment configuration views/serializers such that assignment configurations are created independently of subsidy access policies.

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_assignment_configuration_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_assignment_configuration_views.py
@@ -7,6 +7,7 @@ import ddt
 from rest_framework import status
 from rest_framework.reverse import reverse
 
+from enterprise_access.apps.content_assignments.models import AssignmentConfiguration
 from enterprise_access.apps.content_assignments.tests.factories import AssignmentConfigurationFactory
 from enterprise_access.apps.core.constants import (
     SYSTEM_ENTERPRISE_ADMIN_ROLE,
@@ -31,7 +32,7 @@ class CRUDViewTestMixin:
         super().setUpTestData()
 
         cls.enterprise_uuid = TEST_ENTERPRISE_UUID
-        other_enterprise_uuid = uuid4()
+        cls.other_enterprise_uuid = uuid4()
 
         # Create a pair of AssignmentConfiguration + SubsidyAccessPolicy for the main test customer.
         cls.assignment_configuration_existing = AssignmentConfigurationFactory(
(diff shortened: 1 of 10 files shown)


enterprise_access/apps/api/v1/tests/test_assignment_configuration_views.py
```diff
@@ -374,46 +365,27 @@ def test_create(self):
...
-        assert response.status_code == status.HTTP_201_CREATED
-        self.assigned_learner_credit_policy_standalone.refresh_from_db()
-        assert response.json() == {
-            'uuid': str(self.assigned_learner_credit_policy_standalone.assignment_configuration.uuid),
-            'active': True,
-            'enterprise_customer_uuid': str(TEST_ENTERPRISE_UUID),
-            'subsidy_access_policy': str(self.assigned_learner_credit_policy_standalone.uuid),
-        }
 
-    def test_create_idempotent(self):
```
macdiesel, line 396: No concerns over dropping the idempotency test?

enterprise_access/apps/api/v1/views/assignment_configuration.py
```diff
@@ -240,14 +221,5 @@ def destroy(self, request, *args, uuid=None, **kwargs):
         # Custom delete() method should set the active flag to False.
         assignment_config_to_soft_delete.delete(reason=delete_reason)
 
-        # Manually unlink this soft-deleted AssignmentConfig from the associated policy.
```
macdiesel, line 243: We stop doing this as the policy records are now soft deleted correct?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -693,11 +704,14 @@ class PerLearnerEnrollmentCreditAccessPolicy(CreditPolicyMixin, SubsidyAccessPol
 
     .. no_pii: This model has no PII
     """
-
-    REQUIRED_CUSTOM_FIELDS = ['per_learner_enrollment_limit']
-
     objects = PolicyManager()
 
+    # Policies of this type *must not* define a per-learner spend limit or an assignment configuration
+    FIELD_CONSTRAINTS = {
```
macdiesel, line 710: I really like this idea.

macdiesel (approved):
One question, looks good. :shipit:

enterprise_access/apps/api/v1/tests/test_assignment_configuration_views.py
iloveagent57: Nope, because we dropped the idempotency!

enterprise_access/apps/api/v1/views/assignment_configuration.py
iloveagent57: Here's how I'm thinking about it:
1. An assignment-based policy record refers to an AssignmentConfiguration, call it config A.
2. For 👋 business reasons 👋 , we soft-delete config A.
3. The assignment-based policy type has a hard requirement of "my config can never be null".  So we leave the soft-deleted config A reference in the policy record.
4. We create config B, and then update the ref in the policy record from soft-deleted config A to new-and-fresh config B.

pwnage101: From a domain-purity perspective, the assignment is not aware of the existence of policies, which is reason enough to remove this manual unlinking logic.  It's completely up to the policy itself to decide what to do when it discovers that its assignment config has been deleted. :+1:

pwnage101 (approved):
LGTM!

PR #278, opened 2023-09-26
feat: Initial REST API endpoints for assignments.

pwnage101:
ENT-7680


enterprise_access/apps/api/v1/views/content_assignments/assignments.py
```diff
@@ -0,0 +1,87 @@
...
+    viewsets.GenericViewSet,
+):
+    """
+    Viewset supporting all Admin CRUD operations on ``LearnerContentAssignment`` records.
+    """
+    permission_classes = (permissions.IsAuthenticated,)
+    serializer_class = serializers.LearnerContentAssignmentResponseSerializer
+    authentication_classes = (JwtAuthentication, authentication.SessionAuthentication)
+    filter_backends = (filters.NoFilterOnDetailBackend,)
+    filterset_class = filters.AssignmentFilter
```
iloveagent57, line 33: Do we need filtering on the learner version?  And if so, can learners filter on the same fields as admins?

iloveagent57 (commented):
This is great, one question about list-filtering for learners.

enterprise_access/apps/api/v1/views/content_assignments/assignments.py
pwnage101: Right, at least not the same as admins.  The one's the learners might conceivably every want to filter on are content_key and state.  I don't know if the learner portal would ever need to do that filtering though.

pwnage101: Maybe we just not pre-emtively introduce dead code and not allow learners to filter the list results.

iloveagent57: Either one sounds reasonable to me!

iloveagent57: anyway, this is my only concern, I'll approve so you can merge at your convenience

iloveagent57 (approved):

PR #284, opened 2023-10-02
feat: LearnerContentAssignmentActions model

iloveagent57:
Implement a model that tracks a history of actions taken against assignments, which actions are not strictly about the lifecycle of the assignment record..  For the MVP, this includes:
* Learner linked to enterprise
* Learner notified (initial notification)
* Learner reminded (possibly multiple reminders)

Includes some updates to existing django admin classes.  I've tried to setup the admin for actions to be useful, with e.g. inlined action records inside the assignment details page.
http://localhost:18270/admin/content_assignments/learnercontentassignment/


enterprise_access/settings/local.py
```diff
@@ -41,8 +41,6 @@
         'debug_toolbar.middleware.DebugToolbarMiddleware',
     )
 
-    DEBUG_TOOLBAR_PATCH_SETTINGS = False
```
iloveagent57, line 44: this setting is no longer used in debug-toolbar

pwnage101 (approved):

PR #286, opened 2023-10-03
feat: allocate view expects non-negative content price. provide well-serialized non-allocation reasons

iloveagent57:
https://2u-internal.atlassian.net/browse/ENT-7689
Couple of main things here.
* Everything above the model/persistence layer now expects to deal in **non-negative** content prices in cents.  That is, the allocate view layer, serialization/validation, and business logic layers all want positive prices in cents, which are converted to negative `content_quantities` before assignment model instances are created/updated.
* We previously didn't have a good happy path, end-to-end test case for requests to the allocate view all the way through the `can_allocate()` business logic checks in the policy layer.  We now have such test cases.
* Changes in the allocate view to return more structured 422 responses, with reason-for-non-allocation error codes and corresponding user-facing messages; very similar to the response pattern for the redeem view.  Added good e2e tests for each of the possible reasons that something may be un-allocatable.

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -398,7 +398,8 @@ class SubsidyAccessPolicy(TimeStampedModel):
         """
         Returns true if redeeming for this content price would exceed
         the given ``limit_to_check`` after taking into account the amount already
-        spent.  ``spent_amount`` is assumed to be an integer <= 0.
+        spent.  ``spent_amount`` is assumed to be an integer <= 0 and ``content_price``
+        is assumed to be an integer >= 0.
         """
         if spent_amount > 0:
             raise Exception('Expected a sum of transaction quantities <= 0')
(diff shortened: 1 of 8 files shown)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -950,7 +951,15 @@ def can_allocate(self, number_of_learners, content_key, content_price_cents):
         into account to determine if ``number_of_learners`` new assignment
         records can be allocated in this policy for the given ``content_key``
         and it's current ``content_price_cents``.
+
+        Params:
+          number_of_learners: Non-negative integer indicating the number of learners to allocate this content to.
+          content_key: Typically a course key (although theoretically could be *any* content identifier).
+          content_price_cents: A *negative* integer reflecting the current price of the content in USD cents.
         """
+        if content_price_cents > 0:
```
adamstankiewicz, line 960: [curious/clarification] Why is `content_price_cents` expected to be negative here when the `ContentMetadataViewSet` in enterprise-subsidy where we're sourcing the price from returns content price as a positive? Should the sign (positive/negative) of content metadata be consistent throughout?

```json
{
    "content_uuid": "0912928c-00cb-44ca-87a0-e0eb75584267",
    "content_key": "OxfordX+OSMEPS",
    "course_run_uuid": "3477b85d-0b85-449c-a211-84fce307b994",
    "course_run_key": "course-v1:OxfordX+OSMEPS+1T2023",
    "source": "2u",
    "mode": "paid-executive-education",
    "content_price": 726100,
    "geag_variant_id": "bd72ac93-e11b-4707-b4b4-772b3c660abe"
}
```

iloveagent57: Naive answer to this question: I changed it to explicitly expect a negative number here to be consistent with what the allocate view (written like 2 weeks ago) currently expects: a negative number in cents.
Real answer to this question: we should do whatever is most reasonable and obvious from the perspective of a client of the allocate view.  If the most reasonable thing is to pass in whatever price is in the metadata (positive cents), then we can do that.

pwnage101 (commented):
LGTM!  besides whatever changes are coming to the sign of price at the view layer.

pwnage101 (approved):

PR #289, opened 2023-10-06
fix: handle HTTPErrors in SubsidyAccessPolicy.subsidy_record()

iloveagent57:
Noticed this bug in the context of listing policies for a particular customer within the provisioning tool.


pwnage101 (approved):

adamstankiewicz (approved):
These changes seem reasonable to me!

PR #290, opened 2023-10-06
feat: rely on DefaultPagination from `edx_rest_framework_extensions`

adamstankiewicz:
Rather than relying on `from rest_framework.pagination import PageNumberPagination` for the base `PaginationWithPageCount` that exists in enterprise-access today and is consumed by viewsets including browse & request, subsidy access policy, and assignments, this PR proposed we rely on `from edx_rest_framework_extensions.paginators import DefaultPagination` instead, so we don't need to manually set the `num_pages` and `current_page` properties ourselves. Additionally, it gives us a `currentPage` property which is useful for working with the `DataTable` in consuming MFE(s).

Secondly, this PR proposes the introduction of an optional `?no_page` query parameter for any viewset that utilizes a new `OptionalPaginationWithPageCount` pagination such that callers can disable pagination for any given request should the use case deem it necessary. The only viewset that utilizes the new `OptionalPaginationWithPageCount` is assignments.

For example, currently, listing all assignments for a learner on the Dashboard page would only return 10 assignments (unless the frontend naively overrides `?page_size=500`) and force the frontend to traverse the pagination which is not performant for users. In such cases, the `?no_page` query parameter could be used to return all assignments as a single list instead of segmented by page.


enterprise_access/apps/api/v1/views/utils.py
```diff
@@ -1,19 +1,59 @@
...
+        Paginates a queryset, if required. Allows disabling pagination by setting
+        the `no_page` query param, supporting use cases where API callers may not be
+        able to rely on pagination (e.g., if more than 1 page of results could be expected
+        for a use case where the client can't traverse all pages and passing a "sufficiently high"
+        `page_size` query parameter is too naive of a solution).
+        """
+        # Determine whether to paginate the queryset based on the value of the
+        # `no_page` query param. If True, do not paginate the response.
+        if self.has_non_paginated_response(request):
+            return None
```
adamstankiewicz, line 56: [inform] By returning `None` here, DRF never calls `self.get_paginated_response()`, so we wouldn't be able to conditionally modify `self.get_paginated_response()` to behave different with the presence of a `?no_page=1` query parameter.

It feels preferable to rely on this logic living in the paginator than the viewset itself even though it means we have to  return a top-level list, e.g.:

```json
[
    {
        "uuid": "db2372f3-c25c-4c38-8de0-60d4d105e0e3",
        "assignment_configuration": "942410d4-ef20-4e4a-9a5f-fa2da07cbb39",
        "learner_email": "[edx@example.com](mailto:edx@example.com)",
        "lms_user_id": null,
        "content_key": "edX+DemoX",
        "content_quantity": -19900,
        "state": "allocated",
        "transaction_uuid": null,
        "last_notification_at": null
    },
    {
        "uuid": "f9c7c4c8-11a7-45f7-ae3e-d81029ec5541",
        "assignment_configuration": "942410d4-ef20-4e4a-9a5f-fa2da07cbb39",
        "learner_email": "[edx@example.com](mailto:edx@example.com)",
        "lms_user_id": null,
        "content_key": "HarvardX+CS50x",
        "content_quantity": -500000,
        "state": "allocated",
        "transaction_uuid": null,
        "last_notification_at": null
    }
]
```

FWIW, the top-level list for non-paginated responses does seem to be the current convention for several (most?) non-paginated list APIs in Enterprise (e.g., `enterprise_course_enrollments`) and the broader Open edX platform.

enterprise_access/apps/content_assignments/admin.py
```diff
@@ -13,8 +13,8 @@ class AssignmentConfigurationAdmin(DjangoQLSearchMixin, SimpleHistoryAdmin):
     Admin configuration for AssignmentConfigurations.
     """
     list_display = (
-        'enterprise_customer_uuid',
         'uuid',
```
adamstankiewicz, line 16: Opting to treat `uuid` as the clickable cell to go into the detail. It felt a bit odd clicking on an enterprise customer UUID to go into a specific content assignment.

enterprise_access/apps/api/v1/views/utils.py
```diff
@@ -1,19 +1,59 @@
 """
 Utilities for REST API View and Viewset classes.
 """
-from rest_framework.pagination import PageNumberPagination
+from django.conf import settings
+from edx_rest_framework_extensions.paginators import DefaultPagination
 
 
-class PaginationWithPageCount(PageNumberPagination):
+class PaginationWithPageCount(DefaultPagination):
```
adamstankiewicz, line 8: By migrating from `PageNumberPagination` (DRF standard) -> `DefaultPagination` (Open edX standard), list APIs throughout enterprise-access will have additional properties such as `num_pages` and `current_page` without extra effort.

However, it does mean we have to override `DefaultPagination`'s `page_size = 10` to keep the original `page_size` for `PaginationWithPageCount` in place (i.e., relies on the `REST_FRAMEWORK.PAGE_SIZE` Django settings.

pwnage101 (approved):
LGTM!  Not gonna block, but I too am leaning on beggs' side of the fence with regards to wanting pagination even if we just set a large page size.  Why can't we just make unreasonably large requests take >1 request to fetch the full list, and 99% of the time the normal experience is 1 request?

enterprise_access/apps/api/v1/views/utils.py
```diff
@@ -1,19 +1,59 @@
...
+    no_page_query_param = 'no_page'
+
+    def formatted_no_page_query_param(self, request):
+        """
+        Parses the query parameters for `self.no_page_query_param` and transforms
+        its value to a lowercase string, if present.
+        """
+        no_page_query_param = request.query_params.get(self.no_page_query_param)
+        formatted_no_page_query_param = no_page_query_param.lower() if no_page_query_param else no_page_query_param
+        return formatted_no_page_query_param
```
iloveagent57, line 35: code-golf nit: you could default the `get()` to return an empty string, or use `or ''` to do something like e.g.
```python
param_value = request.query_params.get(self.no_page_query_param) or ''
return param_value.lower()
```
or if we believe `no_page` will never be set explicitly as null in the query params, simply
```python
return request.query_params.get(self.no_page_query_param, '').lower()
```

iloveagent57 (approved):

adamstankiewicz:
@pwnage101 @iloveagent57 I've decided to remove the `?no_page` update from this PR (more details why below); the only changes in this PR now are to ensure we rely on the `DefaultPagination` from `edx_rest_framework_extensions` to get the default `num_pages` and `current_page` fields returned.

The rationale to no longer include support for `?no_page` is the realization that the learner-facing assignments list API can't be used as it exists today to support the learner portal dashboard UX, given it requires additional metadata for display on the assignment cards. Additionally, the assignments list API currently assumes a single request per budget as the list API is for a given assignment configuration UUID. As is, the frontend would need to resolve all budgets available to the learner through the `credits_available` API, then make another call to the assignments list API for each assignable budget type. Ideally, we'd have a single API endpoint that returns all assignments and their associated course metadata for display in the UI.

Given the uncertainty around which API endpoint(s) will be used to support the learner-facing UX of content assignments (a different assignments list API? extending `credits_available`? etc.), I've decided to punt on supporting `?no_page` on the existing list API since there's a good chance it won't be used for the learner UX as is anyways.

iloveagent57 (approved):

PR #292, opened 2023-10-12
feat: [assignments] backend support for "State" and "Recent action" columns

pwnage101:
Assignment serializer contains three additional output keys:

* `actions` (list of dict): A list of all serialized actions for the assignment.
* `recent_action` (dict): intended to power the "Recent action" column in the frontend display.
* `learner_state` (str): intended to power the "Status" column in the frontend display.

Assignment list views are now sortable on two keys:

* `learner_state_sort_order`: This sorts based on the internal sort order of dynamic field `learner_state`.
* `recent_action_time`: This is a dynamic field defined as `coalesce(most recent reminder, assignment creation)`

Sample URLs
-----------

Listing assignments sorted by learner_state:
```
/api/v1/assignment-configurations/{assignment_configuration_uuid}/admin/assignments/?ordering=learner_state_sort_order
/api/v1/assignment-configurations/{assignment_configuration_uuid}/admin/assignments/?ordering=-learner_state_sort_order
```

Listing assignments sorted by recent_action_time:
```
/api/v1/assignment-configurations/{assignment_configuration_uuid}/admin/assignments/?ordering=recent_action_time
/api/v1/assignment-configurations/{assignment_configuration_uuid}/admin/assignments/?ordering=-recent_action_time
```

Sample Response Snippets
------------------------

Serialized assignment:
```
{
    [...]
    'actions': [
        {
            'uuid': <action uuid>,
            'action_type': <one of AssignmentActions: ["learner_linked", "notified", "reminded"]>,
            'completed_at': <action completed_at timestamp>,
            'error_reason': <action error message>,
        },
        [...]
    ],
    'recent_action': {
        'action_type': <one of AssignmentRecentActionTypes: ["assigned", "reminded"]>,
        'timestamp': <timestamp of recent action>,
    },
    'learner_state': <one of AssignmentLearnerStates: ["notifying", "waiting", "failed"]>,
}
```

ENT-7780

---

Redoc preview:

image

### DIFF
--- a/enterprise_access/apps/api/serializers/content_assignments/assignment.py
+++ b/enterprise_access/apps/api/serializers/content_assignments/assignment.py
@@ -43,9 +43,11 @@ class LearnerContentAssignmentRecentActionSerializer(serializers.Serializer):
     action_type = serializers.ChoiceField(
         help_text='Type of the recent action.',
         choices=AssignmentRecentActionTypes.CHOICES,
+        source='recent_action',
     )
     timestamp = serializers.DateTimeField(
         help_text='Date and time when the action was taken.',
+        source='recent_action_time',
     )
 
 
@@ -59,18 +61,6 @@ class LearnerContentAssignmentResponseSerializer(serializers.ModelSerializer):
     actions = LearnerContentAssignmentActionSerializer(
         help_text='All actions associated with this assignment.',
         many=True,
-    )
-    recent_action = LearnerContentAssignmentRecentActionSerializer(
-        help_text='Structured data about the most recent action. Meant to power a frontend table column.',
-        source='get_recent_action_data',
-    )
-    learner_state = serializers.ChoiceField(
-        help_text=(
-            'learner_state is an admin-facing dynamic state, not to be confused with `state`. Meant to power a '
-            'frontend table column.'
-        ),
-        choices=AssignmentLearnerStates.CHOICES,
-        source='get_learner_state',
     )
 
     class Meta:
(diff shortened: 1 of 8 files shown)


enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -3,18 +3,23 @@
 """
 import logging
 
+from django.db.models import Case, Exists, F, Max, OuterRef, Prefetch, Q, Value, When
```
pwnage101, line 6: I think this one line exemplifies my concerns over the last week about frontend-backend design compatibility and tech debt.  There's a time and place for ORM magic, this should not have been one of them.

pwnage101:
Right now this doesn't give the frontend an direct field to distinguish between "Notifying" and "Waiting" assignments, but it is sorted correctly.  I'd like to address that in a separate PR tomorrow.

enterprise_access/apps/api/serializers/content_assignments/assignment.py
```diff
@@ -3,20 +3,76 @@
...
     # This causes the related AssignmentConfiguration to be serialized as a UUID (in the response).
     assignment_configuration = serializers.PrimaryKeyRelatedField(read_only=True)
 
+    actions = LearnerContentAssignmentActionSerializer(
+        help_text='All actions associated with this assignment.',
+        many=True,
+    )
+    recent_action = LearnerContentAssignmentRecentActionSerializer(
+        help_text='Structured data about the most recent action. Meant to power a frontend table column.',
+        source='get_recent_action_data',
```
iloveagent57, line 65: Couldn't these use the annotated fields you added to the queryset?  That would save a couple of queries per row of response data.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -62,14 +70,25 @@ def get_queryset(self):
         """
         A base queryset to list or retrieve ``LearnerContentAssignment`` records.
         """
+        queryset = LearnerContentAssignment.objects.all()
```
iloveagent57, line 73: Should we throw a `select_related('actions')` on here?

iloveagent57 (changes requested):
(Experimenting with big red button)
This looks great! Two questions to maybe help with DB query count.

enterprise_access/apps/api/serializers/content_assignments/assignment.py
iloveagent57: ...or it's possible that an upstream `select_related()` would avoid the additional queries, and that using these functions as the source is fine.

PR #295, opened 2023-10-16
fix: nested serialization of assignment config within policy

iloveagent57:
also, fixes policy django admin pages to not display policy serialization when adding a new record.

image


adamstankiewicz (approved):
LGTM! Trying it locally, it resolves the 500 status code issue I was previously seeing when trying to create any policies via Django Admin. I've updated my frontend PR to account for this `assignment_configuration` serialization as well ✅

PR #296, opened 2023-10-19
feat: persist course titles on assignments for UIUX

johnnagro:
## Description

In order to more efficiently display and use Content/Course Titles on various assignments-related experiences, we should persist the title in our DB.

## Next Steps

If we like this approach I can add a management command to back-fill missing course titles.


enterprise_access/apps/content_assignments/models.py
```diff
@@ -142,6 +142,14 @@ class Meta:
             "ContentMetadata.content_key in enterprise-catalog."
         ),
     )
+    content_title = models.CharField(
```
johnnagro, line 145: later on we can develop a process that updates `content_title` and `learner_email` from authoritative sources if need be

pwnage101 (approved):
LGTM!

PR #306, opened 2023-10-25
feat: populate assignment.lms_user_id on core.User.save()

pwnage101:
During the post_save hook for core.User, lookup any existing assignments for the user and populate the lms_user_id into the assignments.  This should happen whenever the user logs in, so it theoretically covers the case where a learner is assigned content before authing with enterprise-access.

The case where a learner is assigned content AFTER authing with enterprise-access would be covered by ENT-7874.

ENT-7875

### DIFF
--- a/enterprise_access/apps/content_assignments/apps.py
+++ b/enterprise_access/apps/content_assignments/apps.py
@@ -4,5 +4,14 @@ from django.apps import AppConfig
 
 
 class ContentAssignmentsConfig(AppConfig):
+    """
+    App config for content_assignments.
+    """
     default_auto_field = 'django.db.models.BigAutoField'
     name = 'enterprise_access.apps.content_assignments'
+
+    def ready(self):
+        super().ready()
+
+        # pylint: disable=unused-import, import-outside-toplevel
+        import enterprise_access.apps.content_assignments.signals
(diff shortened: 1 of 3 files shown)


enterprise_access/apps/core/signals.py
```diff
@@ -0,0 +1,39 @@
+"""
+Signal handlers for core app.
+"""
+import logging
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from enterprise_access.apps.content_assignments.models import LearnerContentAssignment
```
iloveagent57, line 9: What would you think of moving this file into the `content_assignments` app instead?  That dependency direction feels a little more obvious to me.

enterprise_access/apps/core/signals.py
```diff
@@ -0,0 +1,39 @@
...
+            assignment.lms_user_id = user.lms_user_id
+        num_assignments_updated = LearnerContentAssignment.bulk_update(assignments_to_update, ['lms_user_id'])
+
+        # Intentionally not logging PII (email).
+        if len(assignments_to_update) > 0:
+            logger.info(
+                f'Set lms_user_id={user.lms_user_id} on {num_assignments_updated} assignments for User.id={user.id}'
+            )
+        else:
+            logger.info(f'lms_user_id NOT updated on any assignments for User.id={user.id}')
```
iloveagent57, line 39: nit: this could end up being a lot of log lines, esp if `core.User` is saved on every successfully authenticated request to this service.

iloveagent57 (approved):

PR #308, opened 2023-10-27
feat: Try to populate lms_user_id on assignments during creation

pwnage101:
This is an attempt to cover the case where learners are already logged in at the moment of assignment.  This case is a hole left by the work in ENT-7875 which only set out to cover the case where a learner was logged out at the moment of assignment.

ENT-7874

### DIFF
--- a/enterprise_access/apps/content_assignments/api.py
+++ b/enterprise_access/apps/content_assignments/api.py
@@ -12,12 +12,25 @@ from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from opaque_keys.edx.locator import CourseLocator
 
+from enterprise_access.apps.core.models import User
 from enterprise_access.apps.subsidy_access_policy.content_metadata_api import get_and_cache_content_metadata
 
 from .constants import LearnerContentAssignmentStateChoices
 from .models import AssignmentConfiguration, LearnerContentAssignment
+from .utils import chunks
 
 logger = logging.getLogger(__name__)
+
+# The number of emails we are allowed to filter on in a single User request.
+#
+# Batch size derivation inputs:
+#   * The old MySQL client limit was 1 MB for a long time.
+#   * 254 is the maximum number of characters in an email.
+#   * 258 is the length an email plus 4 character delimiter: `', '`
+#   * Divide result by 10 in case we are off by an order of magnitude.
+#
+# Batch size derivation formula: ((1 MB) / (258 B)) / 10 ≈ 350
+USER_EMAIL_READ_BATCH_SIZE = 350
 
 
 class AllocationException(Exception):
@@ -233,6 +246,11 @@ def allocate_assignments(assignment_configuration, learner_emails, content_key, 
     # new assignments for these.
     learner_emails_with_existing_assignments = set()
 
+    # Try to populate lms_user_id field on any existing assignments found.  We already ran this function on these
+    # assignments (when they were created in a prior request), but time has passed since then so the outcome might be
+    # different this time. It's technically possible some learners have registered since the last request.
+    assignments_with_updated_lms_user_id = _try_populate_assignments_lms_user_id(existing_assignments)
+
     # Split up the existing assignment records by state
     for assignment in existing_assignments:
         learner_emails_with_existing_assignments.add(assignment.learner_email)
@@ -244,10 +262,20 @@ def allocate_assignments(assignment_configuration, learner_emails, content_key, 
         else:
             already_allocated_or_accepted.append(assignment)
 
+    # These two sets of updated assignments may contain duplicates when combined. Since the duplicates are just
+    # references to the same assignment object, and django model instances are hashable (on PK), they can be
+    # de-duplicated using set union.
+    existing_assignments_to_update = set(cancelled_or_errored_to_update).union(assignments_with_updated_lms_user_id)
+
     # Bulk update and get a list of refreshed objects
     updated_assignments = _update_and_refresh_assignments(
-        cancelled_or_errored_to_update,
-        ['content_quantity', 'state']
+        existing_assignments_to_update,
+        [
+            # `lms_user_id` is updated via the _try_populate_assignments_lms_user_id() function.
+            'lms_user_id',
+            # `content_quantity` and `state` are updated via the for-loop above.
+            'content_quantity', 'state',
+        ]
     )
 
     # Narrow down creation list of learner emails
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/content_assignments/api.py
```diff
@@ -294,6 +295,24 @@ def _get_content_title(assignment_configuration, content_key):
...
+    Notes:
+    * This function does NOT save() the assignment record.
+    * This is a best-effort only; most of the time a User will not exist, and this function is a no-op.
+    * If multiple User records match based on email, choice is non-deterministic.
+
+    Args:
+        LearnerContentAssignment: the assignment on which to update the lms_user_id field.
+    """
+    if not assignment.lms_user_id:
+        first_user = User.objects.filter(email=assignment.learner_email).first()
```
iloveagent57, line 311: My only concern with this is that we'll end up doing one select against the `User` table for each provided learner email.  Can we prefetch the users in bulk, store them in a dictionary of `{email: User record}`, and use that dictionary to populate the `lms_user_id`?

pwnage101: I was trying to brainstorm how to do this, but what comes to mind aren't ideal (but still better than what's currently written).
1. Prefetch only the users relevant to the requested emails.  I think this would be sensitive to limitations in how long one line of SQL can be, since the statement length is a function of the request size.
2. Prefetch all users in the system to build a mapping including users we don't care about.  I think this would result in a full table scan on a potentially very large table (User).
3. \# 1, but prefetch users in batches.  This would work but the code complexity seems too heavy for such a simple task.
4. Bulk create assignments without lms_user_id, then followup fast with a fancy ORM query to do an update+join (second write) to populate the lms_user_id on the newly created assignments.  The downside here is two writes for an API endpoint which is supposed to be super reliable and atomic.

I'm leaning towards 4 but LMK what you think about introducing a second write to the transaction.

pwnage101: we discussed during standup and landed on option 3 because we have some prior art doing it this way in license-manager already.

enterprise_access/apps/content_assignments/api.py
```diff
@@ -12,13 +12,26 @@
...
 
+# The number of emails we are allowed to filter on in a single User request.
+#
+# Batch size derivation inputs:
+#   * The old MySQL client limit was 1 MB for a long time.
+#   * 254 is the maximum number of characters in an email.
+#   * 258 is the length an email plus 4 character delimiter: `', '`
+#   * Divide result by 10 in case we are off by an order of magnitude.
+#
+# Batch size derivation formula: ((1 MB) / (258 B)) / 10 ≈ 387
```
iloveagent57, line 32: I deeply, deeply appreciate you showing the mathematical derivation of this constant.  Could we knock it down to say, 350, to give us a buffer and help my base-10 brain?

enterprise_access/apps/content_assignments/api.py
```diff
@@ -294,10 +312,43 @@ def _get_content_title(assignment_configuration, content_key):
...
+        emails = [assignment.learner_email for assignment in assignment_chunk]
+        # Construct a list of tuples containing (email, lms_user_id) for every assignment in this chunk.
+        email_lms_user_id = User.objects.filter(
+            email__in=emails,  # this is the part that could exceed max statement length if batch size is too large.
+            lms_user_id__isnull=False,
+        ).values_list('email', 'lms_user_id')
+        # dict() on a list of 2-tuples treats the first elements as keys and second elements as values.
+        lms_user_id_by_email = dict(email_lms_user_id)
+        for assignment in assignment_chunk:
+            assignment.lms_user_id = lms_user_id_by_email.get(assignment.learner_email)
```
iloveagent57, line 344: 👏

enterprise_access/apps/content_assignments/api.py
```diff
@@ -309,9 +360,17 @@ def _create_new_assignments(assignment_configuration, learner_emails, content_ke
             content_quantity=content_quantity,
             state=LearnerContentAssignmentStateChoices.ALLOCATED,
         )
-        assignment.full_clean()
         assignments_to_create.append(assignment)
 
+    # Next, try to populate the lms_user_id field on all assignments to be created (resulting in reads against User).
+    # Note: This covers the case where an admin assigns content to a learner AFTER they register.  For the case where an
+    # admin assigns content to a learner BEFORE they register, see the User post_save hook implemented in ENT-7875.
```
iloveagent57, line 367: Nit: maybe a reference to the filename instead of the ticket number?

iloveagent57 (approved):
Looks great, two non-blocking nits!

enterprise_access/apps/content_assignments/api.py
```diff
@@ -309,9 +380,17 @@ def _create_new_assignments(assignment_configuration, learner_emails, content_ke
             content_quantity=content_quantity,
             state=LearnerContentAssignmentStateChoices.ALLOCATED,
         )
-        assignment.full_clean()
         assignments_to_create.append(assignment)
 
+    # Next, try to populate the lms_user_id field on all assignments to be created (resulting in reads against User).
+    # Note: This covers the case where an admin assigns content to a learner AFTER they register.  For the case where an
+    # admin assigns content to a learner BEFORE they register, see the User post_save hook implemented in signals.py.
+    _try_populate_assignments_lms_user_id(assignments_to_create)
```
iloveagent57, line 388: nit: is it worth storing the result of this call in a variable, and then validating/bulk-creating below?  I know it doesn't functionally matter because `_try_populate_assignments_lms_user_id()` has the side-effect of possibly updating every assignment in `assignments_to_create`, but it's possibly confusing.

enterprise_access/apps/content_assignments/tests/test_api.py
```diff
@@ -366,3 +374,151 @@ def test_cancel_assignments_happy_path(self):
...
+                learner_email=existing_user.email,
+                lms_user_id=None,
+                state=LearnerContentAssignmentStateChoices.ALLOCATED,
+            )
+            for existing_user in existing_users
+        ]
+        # Assemble a list of assignments 2.5 longer than the max batch size.
+        all_assignments = assignments_for_non_existing_users + assignments_for_existing_users
+
+        _try_populate_assignments_lms_user_id(all_assignments)
```
iloveagent57, line 511: minor suggestion: this is one of those rare cases where `assertNumQueries` might actually be valuable. https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries

iloveagent57 (approved):

enterprise_access/apps/content_assignments/api.py
pwnage101: Can you elaborate more?  My understanding is that a non-strict superset of the assignment objects with updated lms_user_ids are already being validated/bulk-created below.

PR #314, opened 2023-11-01
fix: data migration to delete Actions sans Assignments, and use on_delete=CASCADE

adamstankiewicz:
https://enterprise-access-internal.stage.edx.org/admin/content_assignments/learnercontentassignmentaction/ currently throwing a `Server Error (500)` due to Actions that exist without associated Assigments, which makes the Django Admin screen throw an exception due to `return obj.assignment.uuid` where `NoneType` object has no attribute `uuid` (in `get_assignment`).

Includes:
* Data migration to delete any Actions without an associated Assignment. Will leave any existing Actions with Assignments remaining to not lose any valid data.
* Adjust the `assignment` field on the Actions model to include `on_delete=models.CASCADE` such that when Assignments get deleted, any associated Actions will also be deleted.
* Adds `created` & `modified` fields to the `LearnerContentAssignmentActionSerializer` so clients have more visibility into when these actions occurred, notably for when they are not successful (`completed_at=None`). Will also be useful to help debug the actions ordering further to QA the "Status" column in the "Assigned" table.
* Being explicit about the order of Actions returned by the Assignments API (previously seemed to be non-deterministic).
* Serializes an `error_reason` field in the Assignments API so the frontend doesn't have to parse the actions to find the appropriate reason why the assignment is in an errored state).
* Ensures the has_reminded derived field only takes into account successful reminder actions. Similarly, ensures the has_notification derived field only takes into account successful email notifications.


enterprise_access/apps/api/v1/tests/test_allocation_view.py
```diff
@@ -6,7 +6,6 @@
 
 import ddt
 from django.core.cache import cache as django_cache
-from django.core.exceptions import ValidationError
```
adamstankiewicz, line 9: `make quality` was complaining about a duplicate import.

enterprise_access/apps/content_assignments/migrations/0010_alter_learnercontentassignmentaction_assignment.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('content_assignments', '0009_historicallearnercontentassignment_content_title_and_more'),
+    ]
+
+    operations = [
+        migrations.RunPython(delete_actions_without_assignment, migrations.RunPython.noop),
```
adamstankiewicz, line 28: Run data migration before altering the `assignments` field below.

enterprise_access/apps/content_assignments/models.py
```diff
@@ -323,12 +323,16 @@ def annotate_dynamic_fields_onto_queryset(cls, queryset):
...
+        # ``recent_action_time`` is defined as the time of the most recent successful reminder, and falls
+        # back to assignment creation time if there are no successful reminders.
         new_queryset = queryset.annotate(
             recent_action_time=Coalesce(
                 # Time of most recent reminder.
-                Max('actions__completed_at', filter=Q(actions__action_type=AssignmentActions.REMINDED)),
+                Max('actions__completed_at', filter=(
+                    Q(actions__action_type=AssignmentActions.REMINDED) &
+                    Q(actions__error_reason__isnull=True) &
+                    Q(actions__completed_at__isnull=False)
```
adamstankiewicz, line 334: Before, this logic would consider reminder actions even when they have an `error_reason` and/or have not been marked as "completed" via the `completed_at` field.

enterprise_access/apps/content_assignments/models.py
```diff
@@ -342,6 +346,8 @@ def annotate_dynamic_fields_onto_queryset(cls, queryset):
                 LearnerContentAssignmentAction.objects.filter(
                     assignment=OuterRef('uuid'),
                     action_type=AssignmentActions.REMINDED,
+                    completed_at__isnull=False,
+                    error_reason__isnull=True,
```
adamstankiewicz, line 350: Before, this logic would consider reminder actions even when they have an `error_reason` and/or have not been marked as "completed" via the `completed_at` field.

enterprise_access/apps/content_assignments/models.py
```diff
@@ -361,6 +367,8 @@ def annotate_dynamic_fields_onto_queryset(cls, queryset):
                 LearnerContentAssignmentAction.objects.filter(
                     assignment=OuterRef('uuid'),
                     action_type=AssignmentActions.NOTIFIED,
+                    completed_at__isnull=False,
+                    error_reason__isnull=True,
```
adamstankiewicz, line 371: Before, this logic would consider notified actions even when they have an `error_reason` and/or have not been marked as "completed" via the `completed_at` field.

enterprise_access/apps/subsidy_access_policy/migrations/0018_alter_subsidyaccesspolicy_assignment_configuration.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.2.21 on 2023-11-01 20:35
```
adamstankiewicz, line 1: This migration is generated with `make migrations` on the `main` branch without any code changes... Should it be committed?

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -274,7 +274,7 @@ def __new__(cls, *args, **kwargs):
         except Exception:  # pylint: disable=broad-except
             pass
 
-        return super().__new__(proxy_class)  # pylint: disable=lost-exception
+        return super().__new__(proxy_class)
```
adamstankiewicz, line 277: `make quality` suggested this pylint ignore comment wasn't necessary.

enterprise_access/apps/api/serializers/content_assignments/assignment.py
```diff
@@ -100,10 +110,34 @@ class LearnerContentAssignmentAdminResponseSerializer(LearnerContentAssignmentRe
...
 
     class Meta(LearnerContentAssignmentResponseSerializer.Meta):
         fields = LearnerContentAssignmentResponseSerializer.Meta.fields + [
             'recent_action',
             'learner_state',
+            'error_reason',
         ]
         read_only_fields = fields
+
+    def get_error_reason(self, assignment):
```
pwnage101, line 123: In order for this field to show up in redoc/openapi as the correct type, you may need to decorate this method with

```python
@extend_schema_field(serializers.CharField)
```

enterprise_access/apps/subsidy_access_policy/migrations/0018_alter_subsidyaccesspolicy_assignment_configuration.py
pwnage101: I feel like this missing migration removes the field-level unique constraint that was supposed to be removed as part of https://github.com/openedx/enterprise-access/pull/258.  I'd say keeping this migration makes the database constraints more closely match the field definition.

Of course now we also know that mysql doesn't actually support the `unique_together` that was supposed to replace the field-level constraint.  But maybe that's a problem for a separate ticket.

pwnage101 (approved):
just a couple nits, looks great!

enterprise_access/apps/subsidy_access_policy/migrations/0018_alter_subsidyaccesspolicy_assignment_configuration.py
iloveagent57: Oh, hmm, weird.  FWIW, MySQL does support `unique_together`, it just doesn't support conditional unique constraints.

iloveagent57: Here's the current UNIQUE keys on the table in our production environment:
```
| subsidy_access_policy_subsidyaccesspolicy | CREATE TABLE `subsidy_access_policy_subsidyaccesspolicy` (
  `created` datetime(6) NOT NULL,
  `modified` datetime(6) NOT NULL,
  `policy_type` varchar(64) NOT NULL,
  `uuid` char(32) NOT NULL,
  ...
  PRIMARY KEY (`uuid`),
  UNIQUE KEY `assignment_configuration_id` (`assignment_configuration_id`),
  UNIQUE KEY `subsidy_access_policy_su_active_assignment_config_f6b904ce_uniq` (`active`,`assignment_configuration_id`),
  ...
```
So the composite key is there, but so is the single-column unique key, which we don't want.  So yes, follow Troy's suggestion and let this migration go in.

enterprise_access/apps/api/serializers/content_assignments/assignment.py
```diff
@@ -80,6 +79,17 @@ class Meta:
         ]
         read_only_fields = fields
 
+    def get_actions(self, assignment):
+        """
+        Resolves any associated actions to the assignment in chronological order based on created timestamp.
+        """
+        related_actions = assignment.actions.order_by('created').all()
```
iloveagent57, line 86: Tiniest possible nit: it's more conventional to put the `all()` before the `order_by()` (because if you were writing it as raw SQL, the `ORDER BY` would come last).

enterprise_access/apps/content_assignments/models.py
iloveagent57, line 346: Another tiniest possible nit: can we switch this so that the error reason filter comes before the completed_at filter?  That way, it'll match the filter order above.

enterprise_access/apps/api/serializers/content_assignments/assignment.py
```diff
@@ -100,10 +110,34 @@ class LearnerContentAssignmentAdminResponseSerializer(LearnerContentAssignmentRe
...
+        """
+        Resolves the error reason for the assignment, if any, for display purposes based on
+        any associated actions.
+        """
+        # If the assignment is not in an errored state, there should be no error reason.
+        if assignment.state != LearnerContentAssignmentStateChoices.ERRORED:
+            return None
+
+        # Assignment is an errored state, so determine the appropriate error reason so clients don't need to.
+        related_actions_with_error = assignment.actions.filter(error_reason__isnull=False).order_by('-created').all()
```
iloveagent57, line 133: nit: you don't need the `.all()` here, since it's already a `filter()`.

enterprise_access/apps/api/serializers/content_assignments/assignment.py
iloveagent57: BTW, there's a whole other way to do this, which is to add an `ordering` to the Action model's meta class. https://docs.djangoproject.com/en/4.2/ref/models/options/#ordering

adamstankiewicz: Oh, nice. Didn't realize there was a shorthand way to do this via the `Meta` class (TIL)! It's a cleaner solution, will update. Thanks!

enterprise_access/apps/subsidy_access_policy/migrations/0018_alter_subsidyaccesspolicy_assignment_configuration.py
adamstankiewicz: Sounds good, thanks for the sanity check! Will keep this migration.

iloveagent57 (approved):

PR #319, opened 2023-11-03
chore: allowing consumers of  to filter by email and content title

alex-sheehan-edx:


adamstankiewicz (commented):
Code changes LGTM; I'd just recommend adding a test case(s) to account for the added search functionality to assert that the search stuff is wired up properly and works as expected. Without a test explicitly covering whether the API supports search, the `search_fields` could somehow get erroneously removed in a future change and no unit tests would break catching the regression.

adamstankiewicz (approved):
LGTM! 😎

katrinan029 (approved):
LGTM!

PR #320, opened 2023-11-06
feat: link learners during assignment allocation

iloveagent57:
Modifies `allocate_assignments()` to call an async task to link the assignment's user to the relevant enterprise, for all updated or created assignments.
https://2u-internal.atlassian.net/browse/ENT-7778

### DIFF
--- a/enterprise_access/apps/content_assignments/api.py
+++ b/enterprise_access/apps/content_assignments/api.py
@@ -18,6 +18,7 @@ from enterprise_access.apps.subsidy_access_policy.content_metadata_api import ge
 
 from .constants import LearnerContentAssignmentStateChoices
 from .models import AssignmentConfiguration, LearnerContentAssignment
+from .tasks import create_pending_enterprise_learner_for_assignment_task
 from .utils import chunks
 
 logger = logging.getLogger(__name__)
@@ -290,6 +291,10 @@ def allocate_assignments(assignment_configuration, learner_emails, content_key, 
         content_quantity,
     )
 
+    # Enqueue an asynchronous task to link assigned learners to the customer
+    for assignment in updated_assignments + created_assignments:
+        create_pending_enterprise_learner_for_assignment_task.delay(assignment.uuid)
+
     # Return a mapping of the action we took to lists of relevant assignment records.
     return {
         'updated': updated_assignments,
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/content_assignments/api.py
```diff
@@ -307,8 +312,10 @@ def _update_and_refresh_assignments(assignment_records, fields_changed):
     LearnerContentAssignment.bulk_update(assignment_records, fields_changed)
 
     # Get a list of refreshed objects that we just updated
-    return LearnerContentAssignment.objects.filter(
-        uuid__in=[record.uuid for record in assignment_records],
+    return list(
+        LearnerContentAssignment.objects.filter(
+            uuid__in=[record.uuid for record in assignment_records],
+        )
```
pwnage101, line 318: Hm, isn't this also sensitive to MySQL statement length limits?  ((1 megabyte) / (40 bytes)) / 10 = 2500 max assignments.

pwnage101: In practice, the modern MySQL client limit is 16MB per request, so without the order-of-magnitude safety net that should result in about (16 megabytes) / (40 bytes) = 400,000 max assignments.

pwnage101 (approved):
Looks great!  The comment is more significant than a nit, but is about code that pre-existed this PR.

enterprise_access/apps/content_assignments/api.py
```diff
@@ -290,6 +291,10 @@ def allocate_assignments(assignment_configuration, learner_emails, content_key,
         content_quantity,
     )
 
+    # Enqueue and asynchronous task to link assigned learners to the customer
```
adamstankiewicz, line 294: nit: s/and/an?

adamstankiewicz (approved):

enterprise_access/apps/content_assignments/api.py
iloveagent57: Ah good point.  I'm going to leave this as-is for now, since anything over 1,000 assignments in a single allocation request is likely going to require a more "bulk-oriented" UX.  I guess we could've said that about the `core.User()` lookup, too.

pwnage101: the core.User lookup was more of a bottleneck since the "safe" batch size we landed on was <1000.

PR #323, opened 2023-11-07
feat: validate requested allocation price

iloveagent57:
https://2u-internal.atlassian.net/browse/ENT-7793

### DIFF
--- a/pylintrc_tweaks
+++ b/pylintrc_tweaks
@@ -10,3 +10,4 @@ DISABLE+=
     django-not-configured,
     consider-using-with,
     useless-option-value,
+    superfluous-parens,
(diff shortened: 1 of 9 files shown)


pylintrc_tweaks
```diff
@@ -10,3 +10,4 @@ DISABLE+=
     django-not-configured,
     consider-using-with,
     useless-option-value,
+    superfluous-parens,
```
iloveagent57, line 13: the linter complained about 
```python
if not (lower_bound <= requested_price_cents <= upper_bound):
```
but in my not-so-humble opinion, parens actually make this (and other such statements) much easier to read.  I think this violation, in general, is misguided.

pwnage101 (approved):

adamstankiewicz (approved):

PR #317, opened 2023-11-03
feat: cancel pending assignment functionality

katrinan029:
# Description
Create a cancel functionality for pending assignments that the admin of an enterprise learner credit no longer wants the learner to enroll in. The learner will receive a cancellation email notification.

# Solution
Add a celery task to the send out a braze email campaign to the learner when the cancellation is successful.

[JIRA Ticket](https://2u-internal.atlassian.net/browse/ENT-7882)

PR #325, opened 2023-11-09
feat: remind pending assignment functionality 

kiram15:
### Description
Create a remind functionality  that the admin of an enterprise learner credit wants to remind them of their pending assignment. The learner will receive a email with course details of their pending assignment. This PR is similar to Katrina's [here](https://github.com/openedx/enterprise-access/pull/317) but for the remind functionality 

### Solution
Add a celery task to the send out a braze email campaign to the learner to remind them

[JIRA Ticket](https://2u-internal.atlassian.net/browse/ENT-7691)


enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -131,3 +131,74 @@ def send_cancel_email_for_pending_assignment(cancelled_assignment_uuid):
...
+    try:
+        recipient = braze_client_instance.create_recipient(
+            user_email=assignment.learner_email,
+            lms_user_id=assignment.lms_user_id,
+        )
+        braze_trigger_properties["organization"] = enterprise_customer_data['name']
+        braze_trigger_properties["course_title"] = assignment.content_title
+        braze_trigger_properties["enrollment_deadline"] = course_metadata['normalized_metadata']['enroll_by_date']
+        braze_trigger_properties["start_date"] = course_metadata['normalized_metadata']['start_date']
+        braze_trigger_properties["course_partner"] = course_metadata['owners'][0]['name']
```
kiram15, line 185: This is only taking the first owner's name to send, is that an OK default to have? Or should I create a method that loops through if there's multiple and do "Harvard and Stanford" or something?

iloveagent57: I think that's fine for now.  We can circle back on this with relative ease in the future.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -204,3 +204,32 @@ def cancel(self, request, *args, uuid=None, **kwargs):
             return Response(response_serializer.data, status=status.HTTP_200_OK)
         else:
             return Response(status=status.HTTP_422_UNPROCESSABLE_ENTITY)
+
+    def remind(self, request, *args, uuid=None, **kwargs):
```
iloveagent57, line 208: you need the same permission and DRF-action decorators as the `cancel` view has:
```python
@permission_required(CONTENT_ASSIGNMENT_ADMIN_WRITE_PERMISSION, fn=assignment_admin_permission_fn)
@action(detail=True, methods=['post'])
```

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -204,3 +204,32 @@ def cancel(self, request, *args, uuid=None, **kwargs):
...
+            404 if the assignment was not found.
+            422 if the assignment was not able to be reminded.
+        """
+        try:
+            assignment_to_remind = self.get_queryset().get(uuid=uuid)
+        except LearnerContentAssignment.DoesNotExist:
+            return Response(None, status=status.HTTP_404_NOT_FOUND)
+
+        # if the assignment is not remindable, this is a no-op.
+        reminder_info = assignments_api.remind_assignments([assignment_to_remind])
```
iloveagent57, line 223: I think this view (and the cancel view, too) have to support bulk actions from the admin portal UI, meaning they have to support the cancellation and reminder of multiple assignments in a single request.  So we'll want this to be a DRF action with `detail=False` that expects to have a list of assignment UUIDs in the POST payload to remind (or cancel for the `cancel` view).

enterprise_access/apps/content_assignments/constants.py
```diff
@@ -11,19 +11,24 @@ class LearnerContentAssignmentStateChoices:
...
 
     # States which allow reallocation by an admin.
     REALLOCATE_STATES = (CANCELLED, ERRORED)
 
     # States which allow cancellation by an admin.
-    CANCELABLE_STATES = (ALLOCATED, ERRORED)
+    CANCELABLE_STATES = (ALLOCATED, ERRORED, REMINDED)
+
+    # States which allow reminders by an admin.
+    REMINDABLE_STATES = (ALLOCATED)
```
iloveagent57, line 31: you need a trailing comma here, otherwise `REMINDABLE_STATES` will just get set to the value of `ALLOCATED`.
for example:
```python
>>> x = 'hello'
>>> foo = (x)
>>> foo
'hello'
>>> for thing in foo: print(thing)
...
h
e
l
l
o
```

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -131,3 +131,78 @@ def send_cancel_email_for_pending_assignment(cancelled_assignment_uuid):
...
+@shared_task(base=LoggedTaskWithRetry)
+def send_reminder_email_for_pending_assignment(assignment_uuid):
+    """
+    Send email via braze for reminding users of their pending assignment
+    Args:
+        assignment_uuid: (string) the subsidy request uuid
+    """
+    # importing this here to get around a cyclical import error
+    # pylint: disable=import-outside-toplevel
+    from enterprise_access.apps.content_assignments.content_metadata_api import get_content_metadata_for_assignments
```
iloveagent57, line 145: You should be able to yank this back up with the rest of the imports now.

enterprise_access/apps/content_assignments/api.py
```diff
@@ -452,23 +453,50 @@ def cancel_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+        send_reminder_email_for_pending_assignment
+    )
+    remindable_assignments = set(
+        assignment for assignment in assignments
+        if assignment.state in LearnerContentAssignmentStateChoices.REMINDABLE_STATES
+    )
+    already_reminded_assignments = set(
+        assignment for assignment in assignments
+        if assignment.state == LearnerContentAssignmentStateChoices.REMINDED
+    )
```
iloveagent57, line 485: Good idea to copy this pattern from above.  Reminding is different from cancellation, though - it's something that can happen more than once for a single assignment, and we want the learner to know about it each time (via the email you're sending).  So get rid of everything about `already_reminded_assignments` from this function.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
iloveagent57: Don't worry about changing the `cancel` view for your PR - I'll ticket out some additional work to fix that up.

iloveagent57 (changes requested):

enterprise_access/apps/content_assignments/constants.py
iloveagent57: vs. 
```python
>>> foo = (x,)
>>> for thing in foo: print(thing)
...
hello
>>>
```

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -131,3 +131,78 @@ def send_cancel_email_for_pending_assignment(cancelled_assignment_uuid):
...
+            trigger_properties=braze_trigger_properties,
+        )
+        learner_content_assignment_action.completed_at = datetime.now()
+        learner_content_assignment_action.save()
+        return
+    except Exception as exc:  # pylint: disable=broad-exception-caught
+        logger.error(f"Unable to send email for {lms_user_id} due to exception: {exc}")
+        learner_content_assignment_action.error_reason = AssignmentActionErrors.EMAIL_ERROR
+        learner_content_assignment_action.traceback = exc
+        learner_content_assignment_action.save()
```
adamstankiewicz, line 208: [question/clarification] If the email fails to send, the associated `learner_content_assignment_action` is updated accordingly, but is the `LearnerContentAssignment` record's `state` updated to `errored`? If the assignment record's `state` isn't updated when an error occurs, the UI won't know to display the "Failed: Bad email" status chip.

enterprise_access/apps/content_assignments/api.py
```diff
@@ -452,23 +453,50 @@ def cancel_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+        }
     """
-    content_keys = sorted({assignment.content_key for assignment in assignments})
-    content_metadata_list = get_and_cache_catalog_content_metadata(enterprise_catalog_uuid, content_keys)
-    metadata_by_key = {
-        record['key']: record for record in content_metadata_list
-    }
+    from enterprise_access.apps.content_assignments.tasks import (  # pylint: disable=import-outside-toplevel
+        send_reminder_email_for_pending_assignment
+    )
```
adamstankiewicz, line 477: nit: could this be moved to the top of the file?

enterprise_access/apps/content_assignments/api.py
```diff
@@ -452,23 +453,50 @@ def cancel_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+    remindable_assignments = set(
+        assignment for assignment in assignments
+        if assignment.state in LearnerContentAssignmentStateChoices.REMINDABLE_STATES
+    )
+    already_reminded_assignments = set(
+        assignment for assignment in assignments
+        if assignment.state == LearnerContentAssignmentStateChoices.REMINDED
```
adamstankiewicz, line 484: In a somewhat similar vein to Alex's comment below, I believe the assignment `state` should remain `allocated` even when there has been a reminder. Whether a reminder has been sent or not can be deduced from any associated successful reminder actions. The frontend "Assigned" table currently only intends to pull in assignment states of `allocated` or `errored` 😄 Additionally, the only column in the "Assigned" table that needs to know the latest reminder action is the "Recent action" column, which should already be set up to handle the display of the "Reminded: <last reminder date>" in that table cell.

enterprise_access/apps/content_assignments/tasks.py
iloveagent57: Ah, that's a good point!

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -168,39 +206,63 @@ def list(self, request, *args, **kwargs):
...
-        # if the assignment is not cancelable, this is a no-op.
-        cancellation_info = assignments_api.cancel_assignments([assignment_to_cancel])
+        uuids = serializer.data['assignment_uuids']
+        for uuid in uuids:
+            response = self.cancel_remind_helper(uuid, True)
+            uuid_response_dict[uuid] = response.status_code
+        for uuid in uuid_response_dict.keys():
+            if uuid_response_dict[uuid] != status.HTTP_200_OK:
+                return Response(uuid_response_dict, status=status.HTTP_422_UNPROCESSABLE_ENTITY)
+        return Response(uuid_response_dict, status=status.HTTP_200_OK)
```
iloveagent57, line 236: The `api.py` functions operate on multiple records for us, so we don't want to loop through one record at a time in here.  I would try to switch this around to a pattern like so (forgive the shorthand/psuedocode):
```python
assignments = self.get_queryset().filter(assignment_configuration__uuid=self.requested_assignment_configuration_uuid, uuid__in=serializer.data['assignment_uuids'])
try:
    results = assignments_api.cancel_assignments(assignments)
    response_serializer = serializers.LearnerContentAssignmentAdminResponseSerializer(results['cancelled'])
    return Response(serializer.data, 200)
except:
    return 422
```

iloveagent57 (changes requested):

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
iloveagent57: You can assume that the `cancel_assignments()` call will either succeed for everything or fail for everything - we don't have to worry about ascertaining status of each assignment one-by-one.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -168,39 +171,71 @@ def list(self, request, *args, **kwargs):
...
-        if cancellation_succeeded:
-            # Serialize the assignment object obtained via get_queryset() instead of the one from the assignments_api.
-            # Only the former has the additional dynamic fields annotated, and those are required for serialization.
-            assignment_to_cancel.refresh_from_db()
-            response_serializer = serializers.LearnerContentAssignmentAdminResponseSerializer(assignment_to_cancel)
-            return Response(response_serializer.data, status=status.HTTP_200_OK)
-        else:
+            response = assignments_api.cancel_assignments(assignments)
+            if len(assignments) < len(request.data['assignment_uuids']):
+                return Response(status=status.HTTP_404_NOT_FOUND)
```
iloveagent57, line 200: This seems like a check we'd want to do before calling `cancel_assignments()`, right?  If this condition were true, it'd mean we couldn't find all of the assignments given the assignment uuids provided by the caller in our `filter()` above.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -168,39 +171,71 @@ def list(self, request, *args, **kwargs):
...
+            if response.get('non_cancelable'):
+                return Response(status=status.HTTP_422_UNPROCESSABLE_ENTITY)
```
iloveagent57, line 202: The assignments in `response['non_cancelable']` aren't related to errors, really - they are just the assignments that are not in a cancellable state (e.g. they're already cancelled, or they're already accepted).  I might change this around a little to be like
```python
if not response.get('cancelled'):
    return Response(status=422)
```

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -168,39 +171,71 @@ def list(self, request, *args, **kwargs):
...
+        serializer.is_valid(raise_exception=True)
+        assignments = self.get_queryset().filter(
+            assignment_configuration__uuid=self.requested_assignment_configuration_uuid,
+            uuid__in=serializer.data['assignment_uuids'])
+        try:
+            response = assignments_api.remind_assignments(assignments)
+            if len(assignments) < len(request.data['assignment_uuids']):
+                return Response(status=status.HTTP_404_NOT_FOUND)
+            if response.get('non_remindable_assignments'):
+                return Response(status=status.HTTP_422_UNPROCESSABLE_ENTITY)
```
iloveagent57, line 238: Same comments here as above in the `cancel` view.

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -125,9 +125,87 @@ def send_cancel_email_for_pending_assignment(cancelled_assignment_uuid):
...
+@shared_task(base=LoggedTaskWithRetry)
+def send_reminder_email_for_pending_assignment(assignment_uuid):
+    """
+    Send email via braze for reminding users of their pending assignment
+    Args:
+        assignment_uuid: (string) the subsidy request uuid
+    """
+    # importing this here to get around a cyclical import error
+    # pylint: disable=import-outside-toplevel
+    from enterprise_access.apps.content_assignments.content_metadata_api import get_content_metadata_for_assignments
```
iloveagent57, line 146: This should be able to be imported at the top of the file now.

iloveagent57 (commented):
Good changes!  I left a couple minor suggestions, then this should be good to merge.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
kiram15: No, because if we call it before then if any of the assignments are not found, it will just return before it hit the cancel rather than canceling the remaining ones.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
kiram15: With this I know they're not related to errors, but wouldn't it mean that some of the uuids were unprocessable?

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
iloveagent57: for posterity: we decided offline to leave the condition as-is, but change the response status to 422 here.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
iloveagent57: for posterity: we decided offline to leave this condition and response as-is.  If we need more fine-grained detail about un-cancellable assignments later, we can modify.

iloveagent57 (approved):

iloveagent57:
closing in favor of the other PR with commits all squashed down

PR #327, opened 2023-11-13
docs: ADR on client-provided allocation price validation

iloveagent57:
This ADR describes an approach to validating client-provided allocation prices.  The implementation way previously merged in https://github.com/openedx/enterprise-access/pull/323/files
https://2u-internal.atlassian.net/browse/ENT-7953


adamstankiewicz (approved):
Very nice!

PR #328, opened 2023-11-14
feat: automatically expire assignments

muhammad-ammar:
**Description:** Find the learner content assignments for which the deadline has passed and then send emails

**JIRA:** https://2u-internal.atlassian.net/browse/ENT-7903


enterprise_access/apps/content_assignments/management/commands/send_assigment_deadline_passed_emails.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # TODO: Check contract expiration date has passed
+        # Not sure from where to fetch this?
+
+        return queryset.values_list(
+            'uuid',
+            flat=True,
+        )
+
+    def handle(self, *args, **options):
```
iloveagent57, line 77: 1. I would iterate over `AssignmentConfiguration` records at the top-level, and then iterate over the allocated assignments for that configuration.  An `AssignmentConfiguration` relates to a `SubsidyAccessPolicy`, which means we can iterate through one budget at a time - probably useful for thinking about logging the progress of this command.
2. We need to make sure that the assignment record has its state updated to `CANCELLED` when it is expired.
3. You can use `get_content_metadata_for_assignments()` function to fetch metadata about an assignment's content.  The place where that lives is being refactored a little right now: https://github.com/openedx/enterprise-access/blob/b17e09e485571ea8efa6f17bd3dc3bd564eac7da/enterprise_access/apps/content_assignments/content_metadata_api.py but the function signature and functionality should remain unchanged.
4. "Contract expiration" really just means that the related subsidy record has expired, which you can check via https://github.com/openedx/enterprise-access/blob/ddd745088f58e97c51175f40836b105d8f66757f/enterprise_access/apps/subsidy_access_policy/models.py#L197
Here's some pseudocode for how'd I would generally go about executing this.
```python
for each active assignment_configuration:
  assignments = assignment_configuration.assignments.filter(state=ALLOCATED)
  for each assignment in assignments:
    if current_date() > assignment.created + 90 days:
      assignment.expire() # this method doesn't exist, but we might want it.  make sure to update the state to cancelled
      send_assignment_deadline_passed_email.delay(assignment.uuid)
    metadata_dict = get_your_metadata_lookup_by_content_key()
    if metadata_dict[assignment.content_key].get('enroll_by_date') < current_date():
        # expire and invoke email task
    if assignment_configuration.policy.subsidy_expiration_date() < current_date():
        # expire and invoke email task
```

One more thing that's in the weeds: make sure to not give `get_content_metadata_for_assignments()` more than 100 assignments as input.  Under the covers, it relies on our enterprise-catalog metadata fetching endpoint, which will return up to 100 metadata results in a single request.

enterprise_access/apps/content_assignments/constants.py
```diff
@@ -34,12 +34,14 @@ class AssignmentActions:
     NOTIFIED = 'notified'
     REMINDED = 'reminded'
     CANCELLED_NOTIFICATION = 'cancelled'
+    DEADLINE_PASSED_NOTIFICATION = 'deadline_passed'
```
iloveagent57, line 37: Good idea.  I might generalize this name/string a little bit to be something like "automatic_cancellation_notification"

iloveagent57 (commented):
I might name the command something like "automatically_expire_assignments" because it's doing more than just sending an email - it also updates the state of assignment records.

enterprise_access/apps/content_assignments/management/commands/automatically_expire_assignments.py
```diff
@@ -0,0 +1,131 @@
...
+                        assignment.created + timezone.timedelta(days=90),
+                        enrollment_end_date,
+                        subsidy_expiration_datetime
+                    )
+
+                    expired_assignment_uuid = None
+                    # just for logging
+                    assignment_expiry_reason = None
+
+                    if current_date > (assignment.created + timezone.timedelta(days=90)):
```
iloveagent57, line 108: I would move your `current_date = timezone.now()` definition to the line right before this - a non-trivial amount of time may pass during this for loops.

iloveagent57, line 108: nit: consider adding a constant like `NUM_DAYS_BEFORE_AUTO_CANCELLATION = 90` somewhere.

enterprise_access/apps/content_assignments/management/commands/automatically_expire_assignments.py
```diff
@@ -0,0 +1,131 @@
...
+                    elif enrollment_end_date and enrollment_end_date < current_date:
+                        expired_assignment_uuid = assignment.uuid
+                        assignment_expiry_reason = AssignmentAutomaticExpiredReason.ENROLLMENT_DATE_PASSED
+                    elif subsidy_expiration_datetime and subsidy_expiration_datetime < current_date:
+                        expired_assignment_uuid = assignment.uuid
+                        assignment_expiry_reason = AssignmentAutomaticExpiredReason.SUBSIDY_EXPIRED
+
+                    if expired_assignment_uuid:
+                        logger.info(
+                            '%s Assignment Expired. UUID: [%s], Reason: [%s]',
```
iloveagent57, line 120: nit: consider adding the assignment configuration uuid in this log info, too.

enterprise_access/apps/content_assignments/management/commands/automatically_expire_assignments.py
```diff
@@ -0,0 +1,131 @@
...
+
+            admin_emails = self.fetch_enterprise_customer_admin_emails(
+                assignment_configuration.enterprise_customer_uuid
+            )
+
+            allocated_assignments = assignment_configuration.assignments.filter(
+                state=LearnerContentAssignmentStateChoices.ALLOCATED
+            )
+
+            paginator = Paginator(allocated_assignments, 100)
```
iloveagent57, line 81: neat idea! I've never thought of this before.

enterprise_access/apps/content_assignments/management/commands/automatically_expire_assignments.py
```diff
@@ -0,0 +1,131 @@
...
+                            expired_assignment_uuid,
+                            assignment_expiry_reason
+                        )
+
+                        if not dry_run:
+                            assignment.state = LearnerContentAssignmentStateChoices.CANCELLED
+                            assignment.save()
+                            send_assignment_automatically_expired_email.delay(expired_assignment_uuid, admin_emails)
+
+                sleep(sleep_duration)
```
iloveagent57, line 131: just curious - what's the `sleep()` for?

iloveagent57 (commented):
Nice job on this!  Couple small suggestions, then this is good to go!

enterprise_access/apps/content_assignments/management/commands/automatically_expire_assignments.py
muhammad-ammar: My understanding was to wait for some time before processing the next batch of assignments? The whole purpose was to let as many celery tasks as possible to complete their work? Thoughts?

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -209,3 +210,50 @@ def send_reminder_email_for_pending_assignment(assignment_uuid):
...
+        learner_content_assignment_action.completed_at = timezone.now()
+        learner_content_assignment_action.save()
+        logger.info(f'Sending braze campaign message for automatically expired assignment {assignment}')
+        return
+    except Exception as exc:
+        logger.error(f"Unable to send email for {lms_user_id} due to exception: {exc}")
+        learner_content_assignment_action.error_reason = AssignmentActionErrors.EMAIL_ERROR
+        learner_content_assignment_action.traceback = exc
+        learner_content_assignment_action.save()
+        raise
```
muhammad-ammar, line 259: Should we update the assignment state to `LearnerContentAssignmentStateChoices.ERRORED`?

iloveagent57: I think we can leave the assignment record's state alone in this case - there's a notion of "learner state" that's derived from the state of the assignment plus the state of most recent actions.  If we need to represent to an admin that we cancelled the assignment, but there was an error sending the email, I think that learner state representation is the right place to do it.  But we don't need to address that now.

enterprise_access/apps/content_assignments/management/commands/automatically_expire_assignments.py
iloveagent57: Ah, I see.  The celery broker can handle the queueing of tasks for us, I don't think we need to `sleep()` here.

iloveagent57 (approved):
Nice job.

PR #333, opened 2023-11-20
perf: improve allocate view latency

iloveagent57:
1. `full_clean()` actually runs queries to validate uniqueness, and we don't need that here. switch to `clean()` instead. https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.full_clean
2. We only need to fetch the content title once, outside of the loop to create assignments.
3. Prefetch updated/created assignments at the tail end of allocation, so that any downstream serializers don't need to do N queries for actions.


brobro10000 (approved):
Doing some homework on my end to learn what `prefetch_related` actually does, but phenomenal job!
Just some breaking tests to fix up 👍🏽

adamstankiewicz (approved):
LGTM, once CI is passing.

PR #334, opened 2023-11-20
feat!: allow in expr filtering for assignment view fields

iloveagent57:
ENT-7809 | Allow all fields that we filter assignments by to accept either exact filtering or "in" filtering.  This slightly changes the current usage of filtering on "learner_state" to remain consistent. Callers must now request filters on this field as "learner_state__in=state1,state2" instead of "learner_state=state1,state2".

### DIFF
--- a/enterprise_access/apps/api/filters/content_assignments.py
+++ b/enterprise_access/apps/api/filters/content_assignments.py
@@ -1,6 +1,8 @@
 """
 API Filters for resources defined in the ``assignment_policy`` app.
 """
+from django_filters import CharFilter
+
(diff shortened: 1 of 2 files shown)


macdiesel (approved):

adamstankiewicz (approved):

adamstankiewicz:
@iloveagent57 Definitely in support of this, for consistency for callers between fields. Just curious if you're also planning to open the PR for the associated frontend change 😄 Fortunately we can live with temporarily breaking the "Assigned" table's learner state filtering because its behind the feature flag.

iloveagent57:
Yup, just opened the PR! https://github.com/openedx/frontend-app-admin-portal/pull/1096/files

PR #335, opened 2023-11-21
feat: create assigned policy configuration on save

iloveagent57:
ENT-7988 | Create an assigned policy's `assignment_configuration` on save() if none is present. Clean up django admin form for assigned policies.

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -996,11 +996,11 @@ class AssignedLearnerCreditAccessPolicy(CreditPolicyMixin, SubsidyAccessPolicy):
     """
     objects = PolicyManager()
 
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -1012,8 +1012,14 @@ class Meta:
     def save(self, *args, **kwargs):
         """
         This type of policy must always have an access_method of "assigned".
+        Additionally, if this record is being created, and no ``assignment_configuration``
+        is present, create one and associate it with this record.
         """
         self.access_method = AccessMethods.ASSIGNED
+        if self._state.adding and not self.assignment_configuration:
```
pwnage101, line 1019: Nit: if self._state.adding is ORM magic, please add a comment explaining what it does.

pwnage101 (approved):

PR #338, opened 2023-11-22
feat: adding error messages to content assignments

kiram15:
Adding granularity of error messaging to use on the front end.

### DIFF
--- a/enterprise_access/apps/content_assignments/tests/factories.py
+++ b/enterprise_access/apps/content_assignments/tests/factories.py
@@ -47,4 +47,4 @@ class LearnerContentAssignmentFactory(factory.django.DjangoModelFactory):
     lms_user_id = factory.LazyAttribute(lambda _: FAKER.pyint())
     content_key = factory.LazyAttribute(lambda _: random_content_key())
     content_title = factory.LazyAttribute(lambda _: f'{FAKER.word()}: a master class')
-    content_quantity = factory.LazyAttribute(lambda _: FAKER.pyint())
+    content_quantity = factory.LazyAttribute(lambda _: FAKER.pyfloat(positive=False, right_digits=0))
--- a/enterprise_access/apps/api/serializers/content_assignments/assignment.py
+++ b/enterprise_access/apps/api/serializers/content_assignments/assignment.py
@@ -7,6 +7,7 @@ from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
 from enterprise_access.apps.content_assignments.constants import (
+    AssignmentActionErrors,
     AssignmentLearnerStates,
     AssignmentRecentActionTypes,
     LearnerContentAssignmentStateChoices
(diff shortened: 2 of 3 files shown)


enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -131,7 +131,7 @@ def send_cancel_email_for_pending_assignment(cancelled_assignment_uuid):
         return
     except Exception as exc:  # pylint: disable=broad-exception-caught
         logger.error(f"Unable to send email for {lms_user_id} due to exception: {exc}")
-        learner_content_assignment_action.error_reason = AssignmentActionErrors.EMAIL_ERROR
+        learner_content_assignment_action.error_reason = AssignmentActionErrors.CANCEL_ERROR
```
katrinan029, line 134: [question]: curious if it would be a better approach to keep the two existing error reasons as is but add an action type.

```
learner_content_assignment_action.error_reason = { 
   error_reason: AssignmentActionErrors.CANCEL_ERROR,
   action_type: AssignmentActions.CANCELLED,
}
```
so the data could look something like this similar to what Adam suggested [here] (https://github.com/openedx/frontend-app-admin-portal/pull/1099/files#r1400722639) and we wouldn't need to add additional error types. 
```
"error_reason":  { 
   "action_type": "cancelled", 
   "error_reason": "internal_api_error" 
}
```

katrinan029 (approved):

enterprise_access/apps/content_assignments/tests/factories.py
```diff
@@ -47,4 +47,4 @@ class Meta:
     lms_user_id = factory.LazyAttribute(lambda _: FAKER.pyint())
     content_key = factory.LazyAttribute(lambda _: random_content_key())
     content_title = factory.LazyAttribute(lambda _: f'{FAKER.word()}: a master class')
-    content_quantity = factory.LazyAttribute(lambda _: FAKER.pyint())
+    content_quantity = -1
```
adamstankiewicz, line 50: nit: I wonder if its worth keeping the random int, but change it up to generate negatives?

enterprise_access/settings/base.py
```diff
@@ -466,7 +466,8 @@ def root(*path_fragments):
 BRAZE_APPROVE_NOTIFICATION_CAMPAIGN = ''
 BRAZE_DECLINE_NOTIFICATION_CAMPAIGN = ''
 BRAZE_AUTO_DECLINE_NOTIFICATION_CAMPAIGN = ''
-BRAZE_ASSIGNMENT_REMINDER_NOTIFICATION_CAMPAIGN = ''
+BRAZE_ASSIGNMENT_REMINDER_LOGISTRATE_CAMPAIGN = 'a5fbad5c-73b2-4aa0-a5be-03045f1043ce'
+BRAZE_ASSIGNMENT_REMINDER_NOTIFICATION_CAMPAIGN = '2009d64a-9980-4fb6-a6a3-005b9afed416'
```
adamstankiewicz, line 470: Are these the stage/prod specific campaign identifiers? If so, they should not be committed in the repo, similar to the others that are default to `''`. The UUIDs will come from our 2U/edX specific config.

adamstankiewicz (approved):
LGTM, seems reasonable. Would just recommend removing the Braze campaign ids from the `base.py` file before it merges.

iloveagent57 (approved):

iloveagent57:
to fix the failing `test_cancel` test, you have to mock the tast that sends the email, like so: 
```
@mock.patch('enterprise_access.apps.content_assignments.tasks.send_cancel_email_for_pending_assignment')
```

enterprise_access/apps/content_assignments/tasks.py
iloveagent57: FYI, the `LearnerContentAssignmentAction.action_type` field already exists, and `cancelled` is one of the choices.

iloveagent57 (changes requested):
Upon further reflection, I don't think this change is needed.  There's already an `action_type` field on the action model, along with an `error_reason` field.  We may just want to ensure that both fields are surfaced via the API as Katrina suggested.

iloveagent57:
Also, sorry for being too quick on the "approve" trigger on Friday, trying to multi-task without really internalizing the code I was reading!

iloveagent57 (approved):
follow-up unit tests forthcoming

PR #340, opened 2023-11-28
fix: prevent drift in existing assignment config customer uuids

iloveagent57:
Ensures that when an assignable budget/policy’s enterprise customer UUID is updated, the associated AssignmentConfiguration's enterprise customer UUID is also updated (these should always match).
ENT-8036

https://github.com/openedx/enterprise-access/assets/2307986/5f22afd8-cfc4-4aac-83bc-79090bba0113


pwnage101 (approved):
I assume you removed the self._state.adding check because we just always want to create an assignment configuration whenever one doesn't exist?

otherwise, LGTM!

adamstankiewicz (approved):

iloveagent57:
> I assume you removed the self._state.adding check because we just always want to create an assignment configuration whenever one doesn't exist?
Yup, exactly!

PR #344, opened 2023-11-30
fix: refactor assignment braze tasks, and so forth

iloveagent57:
Refactor to send consistent, standardized braze trigger properties for all of our assignment campaigns.
1. Produces human-readable date strings for start/enroll-by dates, e.g. "Jan 01, 2024" instead of "2024-01-01T...".
2. Produces well format course partner strings.
3. Safely fetches card image urls.
4. Introduces a utility class for sending braze campaigns with standard trigger properties from all of the braze email tasks.


adamstankiewicz (approved):
Seems reasonable!

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -12,17 +12,142 @@
...
+        self._course_metadata = None
+
+    def send_campaign_message(self, braze_trigger_properties, campaign_identifier):
+        """
+        Creates a recipient and sends a braze campaign message.
+        """
+        recipient = self.braze_client.create_recipient(
+            user_email=self.assignment.learner_email,
+            lms_user_id=self.assignment.lms_user_id,
+        )
```
irfanuddinahmad, line 76: @iloveagent57 Do we need to check if the lms_user_id has been created in Braze before triggering the campaign?

PR #348, opened 2023-12-01
feat: subsidy access policy description field allowed to be blank

pwnage101:
Before this change, editing learner credit plans from the LC provisioning tool in frontend-app-support-tools had a weird UX where blanking the description field of a budget/policy would silently re-populate the field with a calculated value.  This allows the frontend to just blank the field and not subvert expectations of the enterprise customer support person, and helps kill some frontend code.

ENT-7922


enterprise_access/apps/content_assignments/migrations/0012_assignmentaction_add_automatic_cancellation_option.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.6 on 2023-12-01 08:50
```
pwnage101, line 1: Throwing this in here as a separate commit.  It was a missed migration from 7242525a034bcded3d9d59230ebafaac1030401d.

enterprise_access/apps/subsidy_access_policy/migrations/0019_subsidyaccesspolicy_description_allow_blank.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.6 on 2023-12-01 08:50
```
pwnage101, line 1: Here's the migration actually for this PR.

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -214,7 +214,7 @@ class Meta:
                 'trim_whitespace': True,
             },
             'description': {
-                'allow_blank': False,
+                'allow_blank': True,
```
adamstankiewicz, line 217: [curious] Should the `description` also now include `null=True` to store `NULL` in the database when left blank?

pwnage101: Django docs discourages allowing both blank (empty string) and null values for any char/text fields to avoid confusion on how to represent (and filter for) "no content".  They specifically recommend always using `null=False`, and `blank` can be anything depending on what's needed.

enterprise_access/apps/content_assignments/migrations/0012_assignmentaction_add_automatic_cancellation_option.py
pwnage101: FYI @muhammad-ammar

macdiesel (approved):

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -119,7 +119,11 @@ class Meta:
         verbose_name='Display name',
         help_text='User-facing display name for this policy record',
     )
-    description = models.TextField(help_text="Brief description about a specific policy.")
+    description = models.TextField(
+        blank=True,
+        null=False,
```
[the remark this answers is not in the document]
pwnage101: oh, that makes sense.  Not sure what the default default was, but explicit is better than implicit so I'll add `default=''`.

PR #349, opened 2023-12-01
fix: braze alias stuff and other fixes

iloveagent57:
* Send aliased recipient when no lms_user_id on assignment (Braze API won't accept request otherwise).
* Task retries should result in only 1 errored action.
* Even safer date parsing (account for possible microseconds).
* Fetch course metadata record inside assignment tasks properly.
ENT-8066

### DIFF
--- a/enterprise_access/apps/api_client/braze_client.py
+++ b/enterprise_access/apps/api_client/braze_client.py
@@ -76,3 +76,15 @@ class BrazeApiClient(BrazeClient):
             'send_to_existing_only': False,
             'trigger_properties': trigger_properties or {},
         }
+
+    def create_recipient_no_external_id(self, user_email):
+        """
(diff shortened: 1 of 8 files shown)


enterprise_access/apps/api_client/braze_client.py
```diff
@@ -76,3 +76,15 @@ def create_recipient(
...
+        """
+        Create a Braze recipient dict identified only by an alias based on their email.
+        """
+        return {
+            'attributes': {'email': user_email},
+            'user_alias': {
+                'alias_label': ENTERPRISE_BRAZE_ALIAS_LABEL,
+                'alias_name': user_email,
+            },
+        }
```
iloveagent57, line 90: We have to use this braze recipient structure to send campaign messages for anonymous users - that is, users for whom we don't yet have an `lms_user_id` (which maps to the "external id" in braze).

PR #354, opened 2023-12-05
feat: get_list_price refactoring for assigned price.

iloveagent57:
ENT-8040 | Computes the list price dict returned by the `can-redeem` view from the value of the `LearnerContentAssignment.quantity` for assignment-based policies.


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -398,14 +404,12 @@ def catalog_contains_content_key(self, content_key):
             content_key,
         )
 
+    @memoize_instance_method
```
iloveagent57, line 407: Subtle change, but take note - the result of this method is now memoized for the lifetime of the policy instance. This is what necessitated some of the additional `tearDown()` logic in the model tests below.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -421,6 +425,18 @@ def get_content_price(self, content_key, content_metadata=None):
...
+    def get_list_price(self, lms_user_id, content_key):  # pylint: disable=unused-argument
+        """
+        Determine the price for content for display purposes only.
+        We likely have content metadata prefetched on this policy record instance at the time
+        of invocation, so we do that prefetch here via ``self.get_content_metadata()``.
+        """
+        return get_list_price_for_content(
+            self.enterprise_customer_uuid,
+            content_key,
+            self.get_content_metadata(content_key),
```
iloveagent57, line 437: The memoization makes this cheap, as long as `get_list_price()` is called _after_ `can_redeem()` was called on this policy.

enterprise_access/apps/subsidy_access_policy/tests/test_models.py
```diff
@@ -76,8 +76,8 @@ def setUp(self):
         )
         self.mock_catalog_contains_content_key = catalog_contains_content_key_patcher.start()
 
-        get_content_metadata_patcher = patch.object(
-            SubsidyAccessPolicy, 'get_content_metadata'
+        get_content_metadata_patcher = patch(
+            'enterprise_access.apps.subsidy_access_policy.models.get_and_cache_content_metadata'
```
iloveagent57, line 80: Better to patch the thing underlying the policy model's `get_content_metadata` method.

enterprise_access/apps/subsidy_access_policy/content_metadata_api.py
```diff
@@ -79,3 +81,46 @@ def get_and_cache_catalog_contains_content(enterprise_catalog_uuid, content_key,
...
+    return list_price_dict_from_usd_cents(content_metadata['content_price'])
+
+
+def list_price_dict_from_usd_cents(list_price_integer_cents):
+    """
+    Helper to compute a list price dictionary given the non-negative price of the content in USD cents.
+    """
+    list_price_decimal_dollars = None
+    if list_price_integer_cents is not None:
+        list_price_decimal_dollars = float(list_price_integer_cents) / 100
```
macdiesel, line 121: The float cast here concerns me since we are working with currency. Why not `Decimal`?

iloveagent57: This is how I found it, Chief! ;)
We could work with a decimal, here, but it'll eventually get `float`-ed when it's serialized as JSON for view resonses.  And the list price is generally used for display purposes only.  I'll try it out and see how it looks.

iloveagent57: hmm, ok `s/float/Decimal/` seems to "just work", so we'll go with that.

enterprise_access/apps/subsidy_access_policy/models.py
pwnage101: Can you explain the reasoning for memoizing this method?  If the purpose is to avoid redundant network calls to memcache during the lifecycle of this request, then my understanding is that TieredCache already handles that under the hood:

> With a tiered cache, multiple calls for the same value within a request will not require additional network calls to Memcached.

https://docs.openedx.org/projects/openedx-proposals/en/latest/best-practices/oep-0022-bp-django-caches.html

enterprise_access/apps/subsidy_access_policy/content_metadata_api.py
```diff
@@ -79,3 +82,46 @@ def get_and_cache_catalog_contains_content(enterprise_catalog_uuid, content_key,
...
+            logger.warning(
+                f'{exc} when checking content metadata for {enterprise_customer_uuid} and {content_key}'
+            )
+            raise ContentPriceNullException(f'Could not determine list price for {content_key}') from exc
+
+    # Note that the "content_price" key is guaranteed to exist, but the value may be None.
+    return list_price_dict_from_usd_cents(content_metadata['content_price'])
+
+
+def list_price_dict_from_usd_cents(list_price_integer_cents):
```
pwnage101, line 116: thanks for splitting this up 👍

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -1060,6 +1076,28 @@ def spend_available(self):
...
+
+    def get_list_price(self, lms_user_id, content_key):
+        """
+        Uses the relevant assignment for this policy instance to determine the appropriate
+        list price of the requested ``content_key``.
+        """
+        found_assignment = self.get_assignment(lms_user_id, content_key)
+        if not found_assignment:
+            return super().get_list_price(lms_user_id, content_key)
+        return list_price_dict_from_usd_cents(found_assignment.content_quantity * -1)
```
pwnage101, line 1099: Please document the expected sign of found_assignment.content_quantity, and the desired sign of get_list_price return value in the docstring.  Thx!

pwnage101 (changes requested):
minor nits and questions, otherwise LGTM!

enterprise_access/apps/subsidy_access_policy/models.py
iloveagent57: Ah, good point, that's not necessary, wasn't thinking about the fact that this was already Tiered-cached.

iloveagent57: And on second though, it seems dangerous to introduce that memoization without some notion of a cache key (for indexing on the input of the function). I'll rip that out and use a RequestCache.

enterprise_access/apps/subsidy_access_policy/models.py
iloveagent57: strike that - it's really request caching here.  There's no longer memoization going on (outside of `RequestCache`).

enterprise_access/apps/subsidy_access_policy/models.py
pwnage101: when I brought up TieredCache was referring to what's already being used by get_and_cache_content_metadata() https://github.com/openedx/enterprise-access/blob/cc341c005bc9b3ca9cdff9fa9a252b1bc7800204/enterprise_access/apps/subsidy_access_policy/content_metadata_api.py#L31

pwnage101 (approved):

PR #355, opened 2023-12-05
fix: more logging for assignment email tasks

iloveagent57:
Some logging to help us understand
1. What trigger properties we're sending for assignment braze campaigns.
2. when we're sending an empty `course_card_image` property, if ever.


brobro10000 (approved):
👍🏽

PR #361, opened 2023-12-11
feat: can-redeem user message for no/cancelled/errored assignments

iloveagent57:
ENT-8114 | The can-redeem endpoint now returns a reason and user message related to policies that are unredeemable due to a lack of allocated `LearnerContentAssignment` record for the requesting user, or the existence of an assignment in a `CANCELLED` or `ERRORED` state.

Sample response payload for my local policy that does *not* have an assignment for the requesting test user:
```
[
    {
        "content_key": "course-v1:edX+DemoX+Demo_Course",
        "list_price": null,
        "redemptions": [],
        "redeemable_subsidy_access_policy": null,
        "can_redeem": false,
        "reasons": [
            {
                "reason": "reason_learner_not_assigned_content",
                "user_message": "You can't enroll right now because this course is not assigned to you.",
                "policy_uuids": [
                    "3cc0ad2c-ede1-476c-8271-97683855e41d"
                ],
                "metadata": {
                        {
                            "email": "enterprise_admin_test-enterprise@example.com",
                            "lms_user_id": 145
                        }
                    ]
                }
            }
        ]
    }
]
```

### Assignment does not exist:
image

### Assignment exists but canceled
image

### Errored assignment
I re-used the "no assignment exists" message here, since there was no pre-defined spec for the message to show in this case.  This is still better than the current state of things, where we'd show a message about a missing code if the requesting learner had an errored assignment for the requested content.
image

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
@@ -24,6 +24,9 @@ from enterprise_access.apps.core.constants import (
     SYSTEM_ENTERPRISE_OPERATOR_ROLE
 )
 from enterprise_access.apps.subsidy_access_policy.constants import (
+    REASON_LEARNER_ASSIGNMENT_CANCELLED,
+    REASON_LEARNER_ASSIGNMENT_FAILED,
+    REASON_LEARNER_NOT_ASSIGNED_CONTENT,
     REASON_NOT_ENOUGH_VALUE_IN_SUBSIDY,
     AccessMethods,
     MissingSubsidyAccessReasonUserMessages,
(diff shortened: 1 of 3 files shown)


enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -2076,3 +2099,90 @@ def test_can_redeem_assigned_policy(self, mock_transactions_cache_for_learner):
             str(self.assigned_learner_credit_policy.uuid)
         assert response_list[0]["can_redeem"] is True
         assert len(response_list[0]["reasons"]) == 0
+
+    @mock.patch('enterprise_access.apps.api.v1.views.subsidy_access_policy.LmsApiClient')
+    @mock.patch('enterprise_access.apps.subsidy_access_policy.subsidy_api.get_and_cache_transactions_for_learner')
+    @ddt.data(
+        # Only a cancelled assignment exists.
+        {'has_cancelled_assignment': True, 'has_failed_assignment': False},
+        # Only an erroed assignment exists.
```
adamstankiewicz, line 2108: nit/typo: 

```suggestion
        # Only an errored assignment exists.
```

adamstankiewicz (approved):

adamstankiewicz:
@iloveagent57 Just to verify/confirm, there were no associated changes necessary in frontend-app-learner-portal-enterprise to ensure the message appears and/or the "Contact administrator" CTA appears?

iloveagent57:
Correct, no local changes to the learner-portal required to get those screenshots as above.

PR #362, opened 2023-12-12
fix: compare emails case-insensitively when setting assignment lms_user_ids

iloveagent57:
ENT-8129 | The email stored on assignment records might be only a case-insensitive match to the corresponding email in the core.User records, so do case-insensitive matching during the assignment flow and during the post-user-save signal that updates assignment ``lms_user_ids``.

### DIFF
--- a/enterprise_access/apps/content_assignments/api.py
+++ b/enterprise_access/apps/content_assignments/api.py
@@ -8,7 +8,8 @@ import logging
 from typing import Iterable
 
 from django.db import transaction
-from django.db.models import Sum
+from django.db.models import Q, Sum
+from django.db.models.functions import Lower
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from opaque_keys.edx.locator import CourseLocator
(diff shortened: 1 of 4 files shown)


macdiesel (approved):

johnnagro (approved):

enterprise_access/apps/content_assignments/api.py
```diff
@@ -392,14 +393,22 @@ def _try_populate_assignments_lms_user_id(assignments):
...
-        email_lms_user_id = User.objects.filter(
-            email__in=emails,  # this is the part that could exceed max statement length if batch size is too large.
-            lms_user_id__isnull=False,
-        ).values_list('email', 'lms_user_id')
+        # this is the part that could exceed max statement length if batch size is too large.
+        # There's no case-insensitive IN query in Django, so we have to build up a big
+        # OR type of query.
+        email_filter = Q()
+        for assignment_email in emails:
+            email_filter |= Q(email__iexact=assignment_email)
```
pwnage101, line 401: FWIW, I think this would increase the statement length (adding a string of OR email checks instead of a flat list of emails).  Might be worth halving USER_EMAIL_READ_BATCH_SIZE to 175 just to be safe.

pwnage101 (changes requested):
The rest LGTM!

enterprise_access/apps/content_assignments/api.py
iloveagent57: Tuned it down to a nice even 100.

PR #363, opened 2023-12-12
fix: case-insensitive email matching for existing assignments

iloveagent57:
ENT-8129 | Also do filters for existing assignments against a batch of emails to allocate in a case-insensitive manner.  Helps in the case where a client allocates to `foo@bar.com` at time t1 and against attempts to allocate to `FOO@bar.com` at time t2.  No second assignment record will be written for `FOO@bar.com`.

### DIFF
--- a/enterprise_access/apps/content_assignments/api.py
+++ b/enterprise_access/apps/content_assignments/api.py
@@ -44,6 +44,19 @@ class AllocationException(Exception):
     user_message = 'An error occurred during allocation'
 
 
+def _inexact_email_filter(emails, field_name='email'):
+    """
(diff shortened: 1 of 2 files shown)


enterprise_access/apps/content_assignments/api.py
```diff
@@ -44,6 +44,19 @@ class AllocationException(Exception):
     user_message = 'An error occurred during allocation'
 
 
+def _inexact_email_filter(emails, field_name='email'):
```
johnnagro, line 47: 🙌

johnnagro (approved):

pwnage101 (approved):

PR #364, opened 2023-12-12
feat: add a "redeemed" assignment action

pwnage101:
This is the backend support for a new "redeemed" assignment action to support storing the error_reason code (and traceback) on failed assignment redemptions (especially from failed calls to the subsidy creation API).

ENT-8105


iloveagent57 (approved):

enterprise_access/apps/content_assignments/models.py
```diff
@@ -372,6 +372,25 @@ def add_errored_expiration_action(self, exc):
             traceback=format_traceback(exc),
         )
 
+    def add_successful_redeemed_action(self):
+        """
+        Adds a successful expiration LearnerContentAssignmentAction for this assignment record.
```
adamstankiewicz, line 377: nit: docstring reference expiration, but this action deals with redemption, not expiration.

enterprise_access/apps/content_assignments/models.py
```diff
@@ -372,6 +372,25 @@ def add_errored_expiration_action(self, exc):
...
+        """
+        return self.actions.create(
+            action_type=AssignmentActions.REDEEMED,
+            completed_at=timezone.now(),
+        )
+
+    def add_errored_redeemed_action(self, exc):
+        """
+        Adds an errored expiration LearnerContentAssignmentAction for this assignment record.
```
adamstankiewicz, line 386: nit: docstring reference expiration, but this action deals with redemption, not expiration.

adamstankiewicz (approved):
LGTM, with 2 tiny nits. 

Just might be also worth sanity checking with Product/UX what, if anything, we expect admins to see in the UI for this error case. That is, will we need to explicitly handle this case in the admin portal "Assigned" table?

As is, I believe it will return the fallback failed `Chip`, which differs from all the other Failed chip.

### Fallback failed chip, which i think would rendered in this case

Note, it wouldn't include a modal popup explaining _why_ it failed with helpful resolution steps / link to the Help Center. It would also only say "Failed" instead of something like "Failed: System"; the error icon on the `Chip` is missing in this fallback case as well.

image

### Explicitly handled failure reason chip

It'd probably be good to show explicit messaging that the assignment failed due to a redemption error so admins have that visibility, similar to the other failure chips, e.g.:

image

PR #366, opened 2023-12-13
fix: add has_dismissed field

katrinan029:


iloveagent57:
I was chatting with Adam about this offline, here's a proposal for a slightly different approach:
1. IMO, I think `has_dismissed` would be better modeled as a new type of `LearnerContentAssignmentAction` rather than a field on the assignment itself, with a new action type like `learner_acknowledged_cancellation`.  
2. You could have the `credits-available` view filter out cancelled assignments based on the presence of such an action.
3. We would then add a new view to the `LearnerContentAssignmentViewSet` with the DRF `action()` decorator, like `def acknowledge_cancellation(...)`.  That view would add the ack'd cancellation action for the assignment.

Let me know what you think!

PR #369, opened 2023-12-19
docs: proposed ADR for assignment expiration improvement

iloveagent57:


docs/decisions/0015-expiration-improvements.rst
```diff
@@ -0,0 +1,94 @@
...
+Decision
+========
+
+``EXPIRED`` should be a distinct Assignment State
+-------------------------------------------------
+This will help disambiguate behavior and simplify/remove business logic
+from the learner-portal frontend.  We may wish to also store the expiration reason
+for an assignment (i.e. which of the 3 cases above caused the assignment to become expired).
+
+Persist the allocation, cancellation, and expiration dates
```
adamstankiewicz, line 43: [clarification] If we were to persist these dates, in the case where an assignment record is updated/re-used (i.e., the assignment allocated -> canceled -> re-allocated use case), I'm assuming the relevant, persisted date fields would be updated to reflect the re-allocation?

docs/decisions/0015-expiration-improvements.rst
```diff
@@ -0,0 +1,94 @@
...
+----------------------------------------------------------
+This also helps simplify and clarify the true state of an assignment for
+consumption by the learner-portal frontend.
+
+Cancellation/Expiration acknowledgement should be an action
+-----------------------------------------------------------
+There should be an action type to support this, and an endpoint scoped to the assignment identifier
+to create such action records.  This will greatly simplify logic in the learner-portal frontend
```
adamstankiewicz, line 51: The dismiss action deals with one _or more_ assignments, across all assignment configurations for the customer. Ideally, the frontend would have a single API call (or least one API call per assignment configuration) in order to dismiss relevant canceled/expired assignments. As is, this implies the frontend would need to make an API call per assignment, which adds complexity and frontend performance concerns (e.g., all N API requests would need to be resolved in parallel, have a reasonable interaction for error handling should one or more of those requests fail, etc. resolve before the alert or assignments can be dismissed.

Edit: I see you mentioned this below:

```
It's ok to not have a bulk API view to acknowledge multiple assignment expiration in a single request -
these acknowledgements are one-time activities for most assignments, and we can live with the cost
of making multiple requests from the frontend for now.  In return, we get to deal with a simpler API view
and logic.
```

What are your concerns around at least making it one API request per assignment configuration (given the API schema today for assignments is specific to assignment configurations)? Given we don't really have any inclination yet for how cancelled/expired assignments learners will have in practice, I'm still a bit uncomfortable about making 1 API request per assignment.

docs/decisions/0015-expiration-improvements.rst
```diff
@@ -0,0 +1,94 @@
...
+and logic.
+
+Modify existing expiration command
+----------------------------------
+* Compare against the allocation date of assignments to determine 90 day expiration condition.
+* Set the state to ``EXPIRED`` and set the expiration date on the expired record.
+
+Consequences
+============
+* The difference between admin-directed cancellation vs. automatic expiration becomes obvious
```
adamstankiewicz, line 73: [note] This should actually get the behavior more in line with the intended UX 🎉

docs/decisions/0015-expiration-improvements.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Alternatives Considered
+=======================
+
+Expiration acknowledgement as an Assignment field
+-------------------------------------------------
+Rejected because it doesn't align with the pattern we've already
+introduced around assignment action records.
+
+Others
```
adamstankiewicz, line 92: Is it worth documenting the (WIP) short-term alternative from ENT-8131 here? That is, the alternative to keep the short-term approach relying on localStorage. Rejected because it allows learners to clear the dismissal history and see canceled/expired assignments again.

docs/decisions/0015-expiration-improvements.rst
adamstankiewicz, line 43: I'm wondering whether the complexity to manage these persisted dates is worth taking on versus only serializing the `earliest-expiration` field via the `credits-available` API as discussed below. That is, will any consumers of the API need the persisted model fields, or only the serialized `earliest-expiration`?

Though, I suppose persisting the model fields would simplify the serialization of `earliest-expiration`?

docs/decisions/0015-expiration-improvements.rst
```diff
@@ -0,0 +1,94 @@
...
+about the state of expiration/cancellation notifications and whether they have been acknowledged/dismissed
+by the learner.
+
+Serialize the expiration deadlines with assignments
+---------------------------------------------------
+Serialize the subsidy expiration date, enrollment deadline, and 90 day deadline in the serialized
+response of assignments inside the ``credits-available`` response payload (perhaps in the
+assignments list view response payload in the future).  Additionally, return a key like ``earliest-expiration``
+that is the minimum of these three dates.
+It's ok to not have a bulk API view to acknowledge multiple assignment expiration in a single request -
```
adamstankiewicz, line 61: See other comment regarding this sentence.

docs/decisions/0015-expiration-improvements.rst
```diff
@@ -0,0 +1,94 @@
...
+======
+Proposed - December 2023
+
+Context
+=======
+There are subtleties around the Assignment cancellation and expiration user
+experience in the enterprise learner-portal frontend that require a more nuanced
+and descriptive approach to the data we persist around these states, and
+how that data is represented in API response payloads for consumption
+by the learner-portal frontend.
```
adamstankiewicz, line 14: nit: might mention the repo name `frontend-app-learner-portal-enterprise` whenever "learner-portal" is referred (e.g., someone might get confused with `frontend-app-learner-portal-programs`?)

docs/decisions/0015-expiration-improvements.rst
```diff
@@ -0,0 +1,94 @@
...
+   date as a proxy for the date at which an assignment was allocated, which is problematic under
+   the edge case where an assignment is allocated, cancelled, and re-allocated later.
+2. The current date is greater than the inferred enrollment deadline for the assigned course.
+3. The current date is greater than the expiration date of the subsidy associated with the Assignment's
+   access policy record.
+
+Learners are notified of expiration and cancellation via the learner-portal frontend
+------------------------------------------------------------------------------------
+The learner-portal notifies learners of assignment cancellation and expiration.  These notifications
+are required to be dismissable, which means we need to persist state about dismissal
```
adamstankiewicz, line 31: nit: Is it worth defining "dismissable" a bit more explicitly? That is, _permanently_ dismissable where after seeing/dismissing them in the UI, the learner would never seen them again (removes any ambiguity around whether "dismiss" means for the page view, for the session, forever, etc.).

docs/decisions/0015-expiration-improvements.rst
adamstankiewicz, line 43: [clarification] At what point would these fields be persisted/updated? For example, if the subsidy expiration date changes, what is the trigger to update the associated assignment records' persisted expiration date, if needed?

macdiesel (approved):

docs/decisions/0015-expiration-improvements.rst
pwnage101: My interpretation is that we would persist all the lifecycle state change dates in-model (allocation date, cancellation date, and expiration date), but the other 4 dates (subsidy expiration date, enrollment deadline 90 day deadline, and earliest-expiration) would all be serialized dynamically and not persisted in-model.

That said, I do feel like out of all the latter 4 dynamic dates, only the last needs to be serialized.  Thoughts?

iloveagent57: Yup, great suggestion, I'll update this section to clarify what to do with these dates on state transition.

docs/decisions/0015-expiration-improvements.rst
iloveagent57: > 1 API request per assignment configuration
That's reasonable, I'll update!

docs/decisions/0015-expiration-improvements.rst
iloveagent57: Yeah, Troy's interpretation is what I was going for, I can clarfiy that better in the doc.  I think only serializing the earliest possible expiration date (which is determined dynamically) makes sense to start.  Can clarify that as well.

iloveagent57: I should clarify those field names and intented semantics.  The fields should really correspond to the datetime at which the assignment record's `state` changed.  For example, when the record is born/allocated, we'd have `state='allocated'` and `allocated_at='2023-12-19T14:28:00Z'`.  When the assignment moves to `expired`, we'd have `state='expired'` and `expired_at='2023-12-25T00:00:00'`.

pwnage101 (approved):

adamstankiewicz (approved):

PR #377, opened 2024-01-08
build: no more mysql 5.7

iloveagent57:
Removes the mysql 5.7 db container from devstack environment.


alex-sheehan-edx (approved):
:shipit: 🔥 🔥

PR #379, opened 2024-01-08
refactor: assignment expiration

iloveagent57:
Part of ENT-7557, refactors expiration business logic into the `content_assignments.api` module, so its
clearer where to find the logic that determines whether and when an assignment is expired.


pwnage101 (approved):
LGTM!

adamstankiewicz (approved):

PR #380, opened 2024-01-09
feat: scrub assignment pii on auto-expiration

iloveagent57:
ENT-7557 | Removes PII on assignment records (and related history) when the record has been in an expirable state for long enough.

# How expiration currently works
There's currently a management command (run via a cron) named ``automatically_exire_assignments.py``
that toggles the state of ``LearnerContentAssignment`` records ("Assignments")
to ``CANCELLED`` under any of the following conditions:

1. The current state of the assignment is ``ALLOCATED``.
2. The current date is more than 90 days after the most recent notification action - this action
   time is used as a proxy for understanding *when* the assignment *was last allocated*. This helps
   deal with the edge case where an assignment is allocated, cancelled, and re-allocated later.
3. The current date is greater than the inferred enrollment deadline for the assigned course.
4. The current date is greater than the expiration date of the subsidy associated with the Assignment's
   access policy record.

# Decision
The above management command will now remove Personally-Identifiable Information ("PII") from assignments
that are automatically moved from ``ALLOCATED`` to ``CANCELLED`` under condition (1) above - that is, only
for such assignments whose last notification time was more than 90 days ago.
This PII includes the learner email address and their ``lms_user_id``.

## Scrubbing the learner email
Note that, to remove learner email PII, we change the value to a "tombstone" - ``retired_user@retired.invalid``.
This is done so that the ``learner_email`` database column can continue to have a non-null constraint.

# Consequences
* Assignments that were cancelled by the admin or that fell into an ``ERRORED`` state will not
  currently have PII cleared by this cron-based management command.
* The Assignment data schema does not yet fully support the improvements proposed in
  `<0015-expiration-improvements.rst>`_. The implementation of these improvements will
  allow us to also take a more nuanced approach about automatically expiring assignment records
  in non-allocated states, or to retire PII fields in other ways.

# Alternatives Considered

## Hook into the edX User Retirement Pipeline
We have not yet rejected, but not yet committed to, integrating ``LearnerContentAssignment``
record retirement with the edX User Retirement Pipeline. This would involve exposing
some API view to scrub PII from certain assignment records associated with a
registered edX user who has requested that their account be retired.

(Above copied from ADR)

### DIFF
--- a/docs/decisions/0016_automatic_expiration.rst
+++ b/docs/decisions/0016_automatic_expiration.rst
@@ -1,0 +1,57 @@
+0016 Automatic Cancellation (Expiration)
+****************************************
+
+Status
+======
+Accepted - January 2024
+
+Context
+=======
+
+How expiration currently works
+------------------------------
+There's currently a management command (run via a cron) named ``automatically_exire_assignments.py``
+that toggles the state of ``LearnerContentAssignment`` records ("Assignments")
+to ``CANCELLED`` under any of the following conditions, as long as the current state
+of the assignment is ``ALLOCATED``:
+
+1. The current date is more than 90 days after the most recent notification action - this action
+   time is used as a proxy for understanding *when* the assignment *was last allocated*. This helps
+   deal with the edge case where an assignment is allocated, cancelled, and re-allocated later. Note
+   that a reminder action on an assignment record *does not* reset this specific expiry time.
+2. The current date is greater than the inferred enrollment deadline for the assigned course.
+3. The current date is greater than the expiration date of the subsidy associated with the Assignment's
+   access policy record.
+
+Decision
+========
+The above management command will now remove Personally-Identifiable Information ("PII") from assignments
+that are automatically moved from ``ALLOCATED`` to ``CANCELLED`` under condition (1) above - that is, only
+for such assignments whose last notification time was more than 90 days ago.
+This PII includes the learner email address.
+
+Scrubbing the learner email
+---------------------------
+Note that, to remove learner email PII, we change the value to a "tombstone" - ``retired_user@retired.invalid``.
+This is done so that the ``learner_email`` database column can continue to have a non-null constraint.
+
+Consequences
+============
+* Assignments that were cancelled by the admin or that fell into an ``ERRORED`` state will not
+  currently have PII cleared by this cron-based management command.
+* The Assignment data schema does not yet fully support the improvements proposed in
+  `<0015-expiration-improvements.rst>`_. The implementation of these improvements will
+  allow us to also take a more nuanced approach about automatically expiring assignment records
+  in non-allocated states, or to retire PII fields in other ways.
+
+Alternatives Considered
+=======================
+
+Hook into the edX User Retirement Pipeline
+------------------------------------------
+We have not yet rejected, but not yet committed to, integrating ``LearnerContentAssignment``
+record retirement with the edX User Retirement Pipeline. This would involve exposing
+some API view to scrub PII from certain assignment records associated with a
+registered edX user who has requested that their account be retired. This is mostly relevant
+in the case of ``ACCEPTED`` assignments, or assignments that have fallen into an ``ERRORED``
+state prior to being automatically expired.
--- a/enterprise_access/apps/content_assignments/constants.py
+++ b/enterprise_access/apps/content_assignments/constants.py
@@ -28,6 +28,9 @@ class LearnerContentAssignmentStateChoices:
     # States which allow reminders by an admin.
     REMINDABLE_STATES = (ALLOCATED,)
 
+    # States from which an assignment can be expired
+    EXPIRABLE_STATES = (ALLOCATED,)
+
 
 class AssignmentActions:
     """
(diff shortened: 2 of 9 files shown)


enterprise_access/apps/content_assignments/models.py
```diff
@@ -391,6 +398,19 @@ def add_errored_redeemed_action(self, exc):
...
+        """
+        self.learner_email = RETIRED_EMAIL_ADDRESS
+        self.lms_user_id = None
+
+    def clear_historical_pii(self):
+        """
+        Removes PII values from this assignment's historical records.
+        """
+        self.history.update(learner_email=RETIRED_EMAIL_ADDRESS, lms_user_id=None)  # pylint: disable=no-member
+
```
iloveagent57, line 413: This mostly copies what we do in [license-manager](https://github.com/openedx/license-manager/blob/cfca89f003707db1d2d2a083cd4f7526808e1aae/license_manager/apps/subscriptions/models.py#L1039-L1053), except here we set to a tombstone-style email address.

enterprise_access/apps/content_assignments/constants.py
```diff
@@ -107,3 +110,5 @@ class AssignmentAutomaticExpiredReason:
 
 
 NUM_DAYS_BEFORE_AUTO_CANCELLATION = 90
+
+RETIRED_EMAIL_ADDRESS = 'retired@edx.org'
```
pwnage101, line 114: Hm, within the retirement pipeline we've gone to moderate lengths to craft a specific format for the email address.  In this case, we cannot use the exact format because we don't always even have a real LMS user to reference, but at least can we inherit the `@retired.invalid` domain? https://github.com/edx/edx-platform/blob/c233327ea07cd3a3869df532e37e7aadca13673d/lms/envs/common.py#L4853

enterprise_access/apps/content_assignments/models.py
```diff
@@ -234,30 +237,36 @@ def bulk_update(cls, assignment_records, updated_field_names):
...
+    def get_auto_expiration_date(self):
         """
-        Returns the first successful "linked" LearnerContentAssignmentActions for this assignment,
+        Returns the date at which this assignment expires due to
+        waiting too long to move into the "accepted" state.
+        """
+        last_notification = self.get_last_successful_notified_action()
+        if not last_notification:
+            return None
+        return last_notification.completed_at + timezone.timedelta(days=NUM_DAYS_BEFORE_AUTO_CANCELLATION)
```
pwnage101, line 248: Can you document that triggering a reminder on an assignment does not re-set the auto-expiration timer?

pwnage101 (commented):
couple comments, the rest LGTM!

pwnage101 (approved):

docs/decisions/0016_automatic_expiration.rst
```diff
@@ -0,0 +1,55 @@
...
+to ``CANCELLED`` under any of the following conditions:
+
+1. The current state of the assignment is ``ALLOCATED``.
+2. The current date is more than 90 days after the most recent notification action - this action
+   time is used as a proxy for understanding *when* the assignment *was last allocated*. This helps
+   deal with the edge case where an assignment is allocated, cancelled, and re-allocated later. Note
+   that a reminder action on an assignment record *does not* reset this specific expiry time.
+3. The current date is greater than the inferred enrollment deadline for the assigned course.
+4. The current date is greater than the expiration date of the subsidy associated with the Assignment's
+   access policy record.
```
pwnage101, line 24: Nit: these list elements can't be ORed together as the previous paragraph suggests, it's more like 1 AND (2 OR 3 OR 4).  Maybe move 1 out of the list?

docs/decisions/0016_automatic_expiration.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Decision
+========
+The above management command will now remove Personally-Identifiable Information ("PII") from assignments
+that are automatically moved from ``ALLOCATED`` to ``CANCELLED`` under condition (1) above - that is, only
+for such assignments whose last notification time was more than 90 days ago.
+This PII includes the learner email address and their ``lms_user_id``.
```
pwnage101, line 31: "lms_user_id is not PII" was my operating assumption for the last 5 years

iloveagent57: Oh hmm, maybe we got it wrong in license-manager?  That existing prior art is the only reason I decided to clear it here.

docs/decisions/0016_automatic_expiration.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Alternatives Considered
+=======================
+
+Hook into the edX User Retirement Pipeline
+------------------------------------------
+We have not yet rejected, but not yet committed to, integrating ``LearnerContentAssignment``
+record retirement with the edX User Retirement Pipeline. This would involve exposing
+some API view to scrub PII from certain assignment records associated with a
+registered edX user who has requested that their account be retired.
```
pwnage101, line 55: I might add that the retirement pipeline may only be required for PII that isn't automatically scrubbed.  In a way, self-destructive PII is safer and more preferable anyway.

pwnage101: That said, accepted assignments are a different story that you're aware of: https://www.cityofsacramento.org/-/media/Corporate/Files/Public-Works/Transportation/Special-Projects/Truxel-Bridge-Info-Sheet82523.pdf?la=en

Maybe just clarify that we may not need the pipeline for expired assignments, but may still need it for accepted ones?

pwnage101 (approved):

PR #382, opened 2024-01-10
Aed/action required by

iloveagent57:
ENT-8146 | Send an `action_required_by` property for assignment braze notification and reminder campaigns.  The value is the min of the enrollment deadline, subsidy expiration, and auto-cancellation date.


pwnage101 (approved):
looks great!

PR #386, opened 2024-01-12
feat: new "retired" field on policy models

pwnage101:
Implements ADR 0016-policy-retirement, and also updates it to reflect
the decision to rename the field to "retired" and change it to a
boolean.

Note this also adds extra visibility into the historical values of
various policy fields via the django admin record history interface.
Previously, changes were only noted (e.g. "active field changed"), but
now the actual historical value of the field is visible.

ENT-8206

Screenshots
===
What the checkboxes look like in django admin edit page for a policy.
image

Example of the history page for a policy that I changed to retired:
Screenshot 2024-01-17 at 15 53 35


enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -1288,6 +1297,11 @@ def test_credits_available_endpoint(
             enterprise_customer_uuid=self.enterprise_uuid,
             active=False,
         )
+        # The following policy should never be returned as it has redeemability disabled.
+        PerLearnerEnrollmentCapLearnerCreditAccessPolicyFactory(
+            enterprise_customer_uuid=self.enterprise_uuid,
+            redeemability_disabled_at=datetime.utcnow() - timedelta(days=1),
+        )
```
pwnage101, line 1304: As discussed at standup on Friday, when redeemability is disabled (was: policy is "retired"), we should make the policy effectively disappear from a learner's perspective.  I.e. it should not be returned by the credits_available endpoint, tested here.

brobro10000 (approved):
Looks good to me. I don't see any usages in the frontend for `policy_is_active` so it looks like its safe to change it without any FE regressions.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -388,12 +388,11 @@ def get_permission_object(self):
 
     def get_queryset(self):
         """
-        Base queryset that returns all active policies associated
+        Base queryset that returns all active & redeemable policies associated
         with the customer uuid requested by the client.
         """
-        return SubsidyAccessPolicy.objects.filter(
+        return SubsidyAccessPolicy.policies_with_redemption_enabled().filter(
```
iloveagent57, line 394: 💯 nice

iloveagent57 (approved):

PR #388, opened 2024-01-18
feat: add nudge braze email using commands

brobro10000:
WIP


enterprise_access/apps/content_assignments/constants.py
```diff
@@ -104,7 +104,7 @@ class AssignmentAutomaticExpiredReason:
     """
     Reason for assignment automatic expiry.
     """
-    NIENTY_DAYS_PASSED = 'NIENTY_DAYS_PASSED'
+    NINETY_DAYS_PASSED = 'NINETY_DAYS_PASSED'
```
brobro10000, line 107: Noticed a typo in this constant, refactored all cases where its used with the correct spelling for the value and variable

enterprise_access/apps/content_assignments/management/commands/automatically_nudge_assignments.py
```diff
@@ -0,0 +1,135 @@
...
+                    start_date = content_metadata.get('normalized_metadata', {}).get('start_date')
+                    course_type = content_metadata.get('course_type')
+
+                    is_executive_education_course_type = course_type in (
+                        'executive-education-2u', 'executive-education')
+
+                    # Determine if the date from today + days_before_course_state_date is
+                    # equal to the date of the start date
+                    # If they are equal, then send the nudge email, otherwise continue
+                    datetime_start_date = self.to_datetime(start_date)
```
iloveagent57, line 111: There's an existing `parse_datetime_string` in here, it's designed for reading several different possible date formats from content metadata records.

iloveagent57: ...if you want to be a good citizen, you could add an optional kwarg to that function like `set_to_utc` that, if true, will return a datetime object with the `.replace(tzinfo=datetime.timezone.utc)` tacked-on, like you have above.

enterprise_access/apps/content_assignments/utils.py
```diff
@@ -15,3 +17,15 @@ def chunks(a_list, chunk_size):
 def format_traceback(exception):
     trace = ''.join(traceback.format_tb(exception.__traceback__))
     return f'{exception}\n{trace}'
+
+
+def are_dates_matching_with_day_offset(days_offset, target_date, date_to_offset):
```
iloveagent57, line 22: Just a suggestion, you could change the name/signature of this function to clarify its intent a little.  The question you're trying to answer here is like "is this date 30 days from today?"  The function could be more precise in trying to answer that question, like
```python
def is_date_n_days_from_now(target_datetime, num_days):
    future_datetime = timezone.now() + timezone.timedelta(days=num_days)
    return target_datetime.date() == future_datetime.date()
```
Also, this function might be better off (for now) being in that `content_metadata_api.py` file.

enterprise_access/apps/content_assignments/management/commands/automatically_nudge_assignments.py
```diff
@@ -0,0 +1,135 @@
...
+                    assignments
+                )
+
+                for assignment in assignments:
+                    content_metadata = content_metadata_for_assignments.get(assignment.content_key, {})
+                    start_date = content_metadata.get('normalized_metadata', {}).get('start_date')
+                    course_type = content_metadata.get('course_type')
+
+                    is_executive_education_course_type = course_type in (
+                        'executive-education-2u', 'executive-education')
```
iloveagent57, line 106: (you already know this) you only need to check against that first string, here.

enterprise_access/apps/content_assignments/management/commands/automatically_nudge_assignments.py
```diff
@@ -0,0 +1,135 @@
...
+                        )
+                        logger.info(message,
+                                    assignment_configuration.uuid,
+                                    start_date,
+                                    datetime_start_date,
+                                    days_before_course_start_date,
+                                    can_send_nudge_notification_in_advance,
+                                    course_type,
+                                    dry_run,
+                                    )
```
iloveagent57, line 133: style nit: we typically try to newline then indent 4 spaces in from the opening line in cases like this
```python
logger.info(
    'my-long-message',
    arg1,
    arg2,
    ...
)
```
https://docs.openedx.org/en/latest/developers/references/developer_guide/style_guides/python-guidelines.html#breaking-long-lines

enterprise_access/apps/content_assignments/management/commands/tests/test_automatically_nudge_assignments.py
```diff
@@ -0,0 +1,587 @@
...
+    def test_command_dry_run(
+            self,
+            mock_subsidy_client,
+            mock_send_reminder_email_for_pending_assignment_task,
+            mock_content_metadata_for_assignments,
+    ):
+        """
+        Verify that management command work as expected in dry run mode.
+        """
+        enrollment_end = timezone.now() - timezone.timedelta(days=5)
```
iloveagent57, line 116: nit: you could tack the `.replace()` right after the `.now()` call to avoid all this reassignment. e.g. `enrollment_end = timezone.now().replace(microsecond=0) - ...`

enterprise_access/apps/content_assignments/management/commands/automatically_nudge_assignments.py
```diff
@@ -0,0 +1,135 @@
...
+                message,
+                assignment_configuration.uuid,
+                subsidy_access_policy.uuid,
+                enterprise_catalog_uuid,
+                assignment_configuration.enterprise_customer_uuid,
+                dry_run,
+            )
+
+            accepted_assignments = assignment_configuration.assignments.filter(
+                state=LearnerContentAssignmentStateChoices.ACCEPTED
```
iloveagent57, line 88: You want to filter for only allocated assignments. `accepted` means they've already redeemed/enrolled.

enterprise_access/apps/content_assignments/management/commands/tests/test_automatically_nudge_assignments.py
```diff
@@ -0,0 +1,587 @@
...
+                },
+                'course_type': 'executive-education-2u',
+            },
+        }
+
+        call_command(self.command, days_before_course_start_date=14)
+
+        mock_send_reminder_email_for_pending_assignment_task.assert_has_calls([
+            call(self.alice_assignment.uuid),
+            call(self.bob_assignment.uuid),
```
iloveagent57, line 288: This assertion will change when you change the filter from accepted to allocated.

iloveagent57 (changes requested):

iloveagent57:
> You want to filter for only allocated assignments. accepted means they've already redeemed/enrolled.
This is the main thing to fix (plus tests) and the reason for requesting changes.

iloveagent57:
Ah, I misunderstood the intent here.  The fact that we're sending reminders made me think you wanted allocated assignments.  But you actually do want accepted ones.  Let's take this conv off github for now, I think there's some nuance here.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -347,3 +349,39 @@ def remind_all(self, request, *args, **kwargs):
...
+        serializer = LearnerContentAssignmentNudgeRequestSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        assignments = self.get_queryset().filter(
+            assignment_configuration__uuid=self.requested_assignment_configuration_uuid,
+            uuid__in=serializer.data['assignment_uuids'],
+        )
+        days_before_course_start_date = serializer.data['days_before_course_start_date']
+        try:
+            response = assignments_api.nudge_assignments(assignments, days_before_course_start_date)
+            return Response(data=response, status=status.HTTP_200_OK)
```
iloveagent57, line 385: nit: you could take advantage of your response serializer here, something like
```python
result = assignments_api.nudge_assignments(assignments, days_before_course_start_date)
response_serializer = LearnerContentAssignmentNudgeResponseSerializer(result)
response_serializer.is_valid()
return Response(data=response_serializer.data, status=status.HTTP_200_OK)
```

enterprise_access/apps/content_assignments/api.py
```diff
@@ -570,6 +578,97 @@ def remind_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+def nudge_assignments(assignments, days_before_course_start_date):
+    """
+    Nudge assignments.
+
+    This is a no-op for assignments in the following state: [allocated, errored, canceled, expired]. We only allow
+    assignments which are in the accepted state.
+
+
+    Args:
+        assignment: An assignment to nudge
```
iloveagent57, line 590: nit: an iterable of assignments to nudge

enterprise_access/apps/content_assignments/api.py
```diff
@@ -570,6 +578,97 @@ def remind_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+        days_before_course_start_date: Number of days prior to start date to nudge individual assignment
+        content_metadata: Content metadata of the assigned course to be nudged
```
iloveagent57, line 592: nit: this function doesn't take this arg.

enterprise_access/apps/content_assignments/api.py
```diff
@@ -570,6 +578,97 @@ def remind_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+    """
+    nudged_assignment_uuids = []
+    unnudged_assignment_uuids = []
+    
+    for assignment in assignments:
+        if assignment.state == LearnerContentAssignmentStateChoices.ACCEPTED:
+            assignment_configuration = AssignmentConfiguration.objects.filter(uuid=assignment.assignment_configuration)
```
iloveagent57, line 599: You should be able to just do `assignment_configuration = assignment.assignment_configuration`

enterprise_access/apps/content_assignments/api.py
```diff
@@ -570,6 +578,97 @@ def remind_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+            subsidy_access_policy = assignment_configuration.subsidy_access_policy
+            enterprise_catalog_uuid = subsidy_access_policy.catalog_uuid
+
+            message = (
+                '[API_BRAZE_EMAIL_CAMPAIGN_NUDGING_1] '
+                'Assignment Configuration. UUID: [%s], '
+                'Policy: [%s], Catalog: [%s], Enterprise: [%s]',
+            )
```
iloveagent57, line 607: nit: you might want to log the assignment uuid in here - without it, you're going to end up logging identical output for each assignment configuration you end up processing.

enterprise_access/apps/content_assignments/api.py
```diff
@@ -570,6 +578,97 @@ def remind_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+                num_days=days_before_course_start_date
+            )
+
+            if is_executive_education_course_type and can_send_nudge_notification_in_advance:
+                message = (
+                    '[API_BRAZE_EMAIL_CAMPAIGN_NUDGING_2] ',
+                    'assignment_configuration_uuid: [%s], start_date: [%s], datetime_start_date: [%s], '
+                    'days_before_course_start_date: [%s], can_send_nudge_notification_in_advance: [%s], '
+                    'course_type: [%s]'
+                )
```
iloveagent57, line 641: add the assignment uuid in here, too.

enterprise_access/apps/content_assignments/api.py
```diff
@@ -570,6 +578,97 @@ def remind_assignments(assignments: Iterable[LearnerContentAssignment]) -> dict:
...
+    Args:
+        assignment: An assignment to nudge
+        days_before_course_start_date: Number of days prior to start date to nudge individual assignment
+        content_metadata: Content metadata of the assigned course to be nudged
+    """
+    nudged_assignment_uuids = []
+    unnudged_assignment_uuids = []
+    
+    for assignment in assignments:
+        if assignment.state == LearnerContentAssignmentStateChoices.ACCEPTED:
```
iloveagent57, line 598: nit: you can get back a whole level of indentation by early-returning here:
```python
if assignment.state != LearnerContentAssignmentStateChoices.ACCEPTED:
    continue
```

enterprise_access/apps/content_assignments/content_metadata_api.py
```diff
@@ -97,3 +102,14 @@ def get_course_partners(course_metadata):
     if len(names) == 2:
         return ' and '.join(names)
     return ', '.join(names[:-1]) + ', and ' + names[-1]
+
+
+def is_date_n_days_from_now(target_datetime, num_days):
+    """
+    Takes an integer number of days to offset from the date_to_offset to determine if
+    the target_date matches the date_to_offset + days_offset date
```
iloveagent57, line 110: nit: update docstring to reflect new parameters.

enterprise_access/apps/content_assignments/management/commands/automatically_nudge_assignments.py
```diff
@@ -0,0 +1,136 @@
...
+                    # Determine if the date from today + days_before_course_state_date is
+                    # equal to the date of the start date
+                    # If they are equal, then send the nudge email, otherwise continue
+                    datetime_start_date = parse_datetime_string(start_date, set_to_utc=True)
+                    can_send_nudge_notification_in_advance = is_date_n_days_from_now(
+                        target_datetime=datetime_start_date,
+                        num_days=days_before_course_start_date
+                    )
+
+                    if is_executive_education_course_type and can_send_nudge_notification_in_advance:
```
iloveagent57, line 118: this code is largely identical to what you have in `api.py` - you should extract that to a single function that you can use in both places, maybe something like
```python
def should_nudge_assignment(assignment, content_metadata, days_before_course_start_date):
```

enterprise_access/apps/content_assignments/management/commands/tests/test_automatically_nudge_assignments.py
```diff
@@ -0,0 +1,569 @@
...
+COMMAND_PATH = 'enterprise_access.apps.content_assignments.management.commands.automatically_nudge_assignments'
+
+
+@pytest.mark.django_db
+class TestAutomaticallyNudgeAssignmentCommand(TestCase):
+    """
+    Tests `automatically_nudge_assignments` management command.
+    """
+
+    def setUp(self):
```
iloveagent57, line 30: if you're not going to modify the test objects in the body of this method, consider using the classmethod `setUpClass()` instead.  `setUpClass()` is run only once per class, whereas `setUp()` is called before the start of each test function.  Doing that helps save a little bit of time during test runs.

iloveagent57 (commented):
This is looking really good!

enterprise_access/apps/content_assignments/management/commands/automatically_nudge_assignments.py
brobro10000: As part of standup discussions, its TBD where functions like these should exist which are dependent on task related imports until a ticket is written up.

enterprise_access/apps/api/v1/views/content_assignments/assignments_admin.py
```diff
@@ -347,3 +349,46 @@ def remind_all(self, request, *args, **kwargs):
...
+            result = assignments_api.nudge_assignments(
+                assignments,
+                assignment_configuration_uuid,
+                days_before_course_start_date
+            )
+            response_serializer = LearnerContentAssignmentNudgeResponseSerializer(data=result)
+            response_serializer.is_valid(raise_exception=True)
+            return Response(data=response_serializer.data, status=status.HTTP_200_OK)
+        except Exception:  # pylint: disable=broad-except
+            return Response(status=status.HTTP_422_UNPROCESSABLE_ENTITY)
```
iloveagent57, line 394: This is a really, really well-factored and well-documented view, with really clear validation and abstractions around our business logic.  Nice work, hang this on your fridge!

enterprise_access/apps/content_assignments/content_metadata_api.py
```diff
@@ -60,7 +62,7 @@ def get_human_readable_date(datetime_string, output_pattern=DEFAULT_STRFTIME_PAT
     return None
 
 
-def parse_datetime_string(datetime_string):
+def parse_datetime_string(datetime_string, **set_to_utc):
```
iloveagent57, line 65: You'll want the signature to look like this: `def parse_datetime_string(datetime_string, set_to_utc=False):...`

`**set_to_utc` is doing something else - it's made it so that all optional/extrac kwargs passed to this function end up in a dict stored in the `set_to_utc` argument.
Here's a useful example: https://book.pythontips.com/en/latest/args_and_kwargs.html#usage-of-kwargs

iloveagent57 (approved):
Approved, but please address that `**kwargs` thing before merging. 👍

PR #390, opened 2024-01-24
feat: introduce earliest_expiry_date on assignments, expired assignment state, API to acknowledge expired/cancelled assignments, and more

adamstankiewicz:
# Description

Ticket: https://2u-internal.atlassian.net/browse/ENT-8162

Related ADR: https://github.com/openedx/enterprise-access/blob/main/docs/decisions/0015-expiration-improvements.rst

Related PR: https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/932

## Changelog

* Introduces new `expired` state for `LearnerContentAssigment` records.
* Introduces the following model fields on `LearnerContentAssignment` records:
  * `allocated_at`
  * `accepted_at`
  * `errored_at`
  * `cancelled_at`
  * `expired_at`
* Removes unused `last_notification_at` field.
* Replaces new `automatic_cancellation` assignment actions with an `expired` assignment actions instead.
* Introduces two new actions related to assignment acknowledgement/dismissal:
  * `cancelled_acknowledged`: "Learner assignment cancellation acknowledged by learner"
  * `expired_acknowledged`: "Learner assignment expiration acknowledged by learner"
* Implements an `acknowledge-assignments` POST API endpoint, that accepts a list of `assignment_uuids` to acknowledge. An appropriate action is created depending on whether the assignment `state` is `expired` or `cancelled`.
  * Utilizes new permission specific to assignment acknowledgement, available to learners/admins/operators for assignments they are linked to through their `lms_user_id`.
* Writes a data migration to migrate "automatic cancellation" actions to "expired" actions and change the associated assignment record's state to `expired` (if currently marked as cancelled).
* Updates the job that automatically expires assignments to set the expired assignments' states to `expired`.
* Introduces a new `learner_state="expired"` field for assignments in an `expired` state.
* Serializes `earliest_possible_expiration` on assignment records (impacts both `credits_available` and the `admin/assignments/` list endpoints).
* Refactors to consolidate business logic around triple-date expiration into one method, usable in multiple existing places where duplicate logic lived. Introduced the consolidated logic to avoid introducing a third duplicate place.

```json
{
    "earliest_possible_expiration": {
        "date": "2023-06-22T00:00:00Z",
        "reason": "SUBSIDY_EXPIRED"
    }
}
```

* Serializes boolean `learner_acknowledged` on assignments and associated assignment `actions`.
* Ensures already acknowledged assignments are NOT returned in the `credits_available` API, such that they no longer appear for learners once acknowledged/dismissed.
* Expired assignments will NOT appear in admin portal Learner Credit Management as the "Assigned" table only looks at `state="allocated"` or `state="errored"`.

## Examples and screenshots

New POST API endpoint `/api/v1/assignment-configurations/<uuid>/acknowledge-assignments/`, accepting the following JSON payload:

```json
{
    "assignment_uuids": ["38c4269e-f902-4066-b628-860493613138"]
}
```

Screenshot of an assignment that has been expired, and then acknowledged through the API. 

image

Screenshot of an assignment that has been cancelled, and then acknowledged through the API.

image

Example API response where provided assignment UUID was successfully acknowledged:

```json
{
    "acknowledged_assignments": [
        "38c4269e-f902-4066-b628-860493613138"
    ],
    "already_acknowledged_assignments": [],
    "unacknowledged_assignments": []
}
```

Example API response where provided assignment UUID has already been acknowledged previously:

```json
{
    "acknowledged_assignments": [],
    "already_acknowledged_assignments": [
        "38c4269e-f902-4066-b628-860493613138"
    ],
    "unacknowledged_assignments": []
}
```

## Testing instructions

Note: the below instructions expire assignment(s) with expired subsidy; you may also expire assignment via 90 days rule or past content enrollment deadline as well.

1. Provision (or re-use) an assignable Policy that is associated with an non-expired Subsidy.
1. Assign assignment(s) to at least 1 user.
1. Edit the Subsidy to be expired (expiration date in the past).
1. Run the `./manage.py automatically_expire_assignments`.
1. Ensure the associated [Learner Portal PR](https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/932) is running.
1. Observe now-expired assignment on Learner Portal dashboard page route with related `Alert` messaging.
1. Dismissing the  related `Alert` messaging should make API call to `acknowledge-assignments` for the expired assignment, and then make another call to `credits_available` to refresh the query cache, thus removing the now-acknowledged assignments from the dashboard UI.


enterprise_access/apps/api/serializers/content_assignments/assignment.py
```diff
@@ -92,11 +96,17 @@ class Meta:
             'content_quantity',
             'state',
             'transaction_uuid',
-            'last_notification_at',
```
adamstankiewicz, line 95: [inform] this field was never used by anything. Opting to remove it.

enterprise_access/apps/content_assignments/tests/test_api.py
```diff
@@ -601,11 +602,16 @@ def setUpClass(cls):
             spend_limit=1000000,
         )
 
-    def mock_content_metadata(self, enroll_by_date):
+    def tearDown(self):
+        # super().tearDown()
```
brobro10000, line 606: [nit] leftover commented code

brobro10000 (commented):
Couple of nits

enterprise_access/apps/content_assignments/models.py
```diff
@@ -246,7 +400,66 @@ def bulk_update(cls, assignment_records, updated_field_names):
...
+                    return datetime_obj.replace(tzinfo=UTC)
+            except ValueError:
+                logger.warning(
+                    'Bad datetime format for %s, value: %s',
+                    content_metadata.get('key'),
+                    enrollment_end_date_str,
+                )
+        return None
+
+    def get_subsidy_expiration(self):
```
iloveagent57, line 451: I have the same sort of question here about which file this code and `get_automatic_expiration_date_and_reason()` really belongs in.  This stuff touches a couple of different external domains (content metadata, subsidy expiration), I'd lean toward moving this to `api.py` (I don't know how much of a headache this would cause in the serializers, though).
The goal would be to help keep the models layer "thin", so that we don't long-term end up with a tangled web of dependencies in one of the core areas, and let `api.py` be kind of the "octopus" of business logic that has its hands in lots of places. I'm open to a differing view on that, though.

enterprise_access/apps/content_assignments/models.py
```diff
@@ -255,13 +468,79 @@ def get_auto_expiration_date(self):
...
         if not last_notification:
             starting_point = self.created
         else:
             starting_point = last_notification.completed_at
-        return starting_point + timezone.timedelta(days=NUM_DAYS_BEFORE_AUTO_CANCELLATION)
+        allocation_timeout_expiration = starting_point + timezone.timedelta(days=NUM_DAYS_BEFORE_AUTO_EXPIRATION)
+        allocation_timeout_expiration = allocation_timeout_expiration.replace(tzinfo=UTC)
+        return allocation_timeout_expiration
+
+    def get_automatic_expiration_date_and_reason(self):
```
iloveagent57, line 481: Is it at all possible to make this method accept `content_metadata` as a parameter.  The mgmt command and the braze task were in charge of managing their content metadata fetches external to this logic, so they could do things like fetch and cache a batch of content metadata in a single call to enterprise-catalog.  I think having the serializer "inject" the metadata would be the only tricky part, but I feel like we have some prior art for that in some other serializer.

enterprise_access/apps/api/serializers/content_assignments/assignment.py
```diff
@@ -258,3 +307,31 @@ def get_content_metadata(self, obj):
...
+
+class LearnerContentAssignmentWithLearnerAcknowledgedResponseSerializer(
+    LearnerContentAssignmentWithContentMetadataResponseSerializer
+):
+    """
+    Read-only serializer for LearnerContentAssignment records that also includes whether or not the learner has
+    acknowledged the assignment.
+    """
+    learner_acknowledged = serializers.SerializerMethodField(
+        help_text='Whether or not the learner has acknowledged the assignment.',
```
iloveagent57, line 320: nit: is this really "the learner has acknowledged expiration or cancelation of the assignment"?

iloveagent57 (commented):
Took a first pass at about 2/3 of the code, here's some initial thoughts!  Nice job gluing all of this together.

enterprise_access/apps/content_assignments/models.py
iloveagent57: ah, serializer `context` is what I was thinking of, here's an example of where we use it for this very reason: https://github.com/openedx/enterprise-access/blob/f328d1990d519ff5b77ccdb3e7088773df070e4b/enterprise_access/apps/api/serializers/content_assignments/assignment.py#L253-L260

enterprise_access/apps/content_assignments/models.py
```diff
@@ -85,6 +92,137 @@ def policy(self):
...
+            )
+
+        # Check whether expiration has ever been acknowledged; if not, acknowledge it.
+        if not expiration_last_acknowledged:
+            return True, False
+
+        # If it has been acknowledged before, check whether last expiration action is newer
+        # than the last acknowledged expiration action.
+        if last_expiration and last_expiration.completed_at > expiration_last_acknowledged.completed_at:
+            return True, True
```
pwnage101, line 119: I have a sneaking suspicion that this case can be lumped into the previous one (return True, False), and this whole method would only have 3 possible outputs: (False, False), (True, False), and (False, True).  Logically, it makes sense to me that if there's been a more recent expiration than the latest expiration acknowledgement, then that really means the latest acknowledgement has been invalidated and there's effectively no existing acknowledgement.

Later in acknowledge_assignments() there is no logic which distinguishes between (True, False) and (True, True) anyway, and the absence of the (True, True) case can simplify the definitions of `has_already_acknowledged_expiration`/`has_already_acknowledged_cancellation`.

pwnage101 (commented):
The rest LGTM!

enterprise_access/apps/api/serializers/content_assignments/assignment.py
pwnage101: +1

enterprise_access/apps/content_assignments/models.py
iloveagent57: At least for the management command it would - this line does batched fetches of metadata from enterprise-catalog, rather than one call per content key. https://github.com/openedx/enterprise-access/blob/f328d1990d519ff5b77ccdb3e7088773df070e4b/enterprise_access/apps/content_assignments/management/commands/automatically_expire_assignments.py#L84-L87 
And also this line in the `learner_assignments` view fetches all the metadata for a list of assignments in one shot: https://github.com/openedx/enterprise-access/blob/f328d1990d519ff5b77ccdb3e7088773df070e4b/enterprise_access/apps/api/serializers/subsidy_access_policy.py#L555
I have to read through the changes to the serializers in this PR again, but I think you can thread that through the context - the new `...LearnerAcknowledgedSerializer` inherits from this existing `...WithContentMetadataSerializer`.

enterprise_access/apps/content_assignments/models.py
```diff
@@ -85,6 +92,137 @@ def policy(self):
...
+
+        # If it has been acknowledged before, check whether last cancellation action is newer
+        # than the last acknowledged cancellation action.
+        if last_cancellation and last_cancellation.completed_at > cancellation_last_acknowledged.completed_at:
+            return True, True
+
+        # Otherwise, the cancellation has already been acknowledged and should not be acknowledged again.
+        return False, True
+
+    def acknowledge_assignments(self, assignment_uuids, lms_user_id):
```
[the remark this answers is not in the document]
iloveagent57: Thinking about this more, I think it's fairly justifiable for this method to exist in `models.py`.  It's basically saying "hey parent object, go update these specific child objects of yours".  The more complex business logic that crosses domains in `get_subsidy_expiration()` and `get_automatic_expiration_date_and_reason()` feels (to me, at least) as if it belongs in api.py.

enterprise_access/apps/core/constants.py
```diff
@@ -26,6 +26,7 @@
 CONTENT_ASSIGNMENT_ADMIN_READ_PERMISSION = 'content_assignment.has_admin_read_access'
 CONTENT_ASSIGNMENT_ADMIN_WRITE_PERMISSION = 'content_assignment.has_admin_write_access'
 CONTENT_ASSIGNMENT_LEARNER_READ_PERMISSION = 'content_assignment.has_learner_read_access'
+CONTENT_ASSIGNMENTS_ACKNOWLEDGE_PERMISSION = 'content_assignment_configuration.has_acknowledge_assignments_access'
```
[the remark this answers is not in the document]
adamstankiewicz: Ah, I had opted for the variable name to be about assignments, not the assignment configuration given the POST API acts on assignments, not assignment configuration objects. If we continue with that, I could update the permission as follows?:

```
CONTENT_ASSIGNMENT_ACKNOWLEDGE_PERMISSION = 'content_assignment.has_acknowledge_access'
```

Otherwise, I can pull in your suggestion. Curious what you think about the options :)

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -1230,11 +1258,16 @@ def redeem(self, lms_user_id, content_key, all_transactions, metadata=None, **kw
...
+            found_assignment.errored_at = localized_utcnow()
             found_assignment.save()
             found_assignment.add_errored_redeemed_action(exc)
             raise
         # Migrate assignment to accepted.
         found_assignment.state = LearnerContentAssignmentStateChoices.ACCEPTED
+        found_assignment.accepted_at = localized_utcnow()
+        found_assignment.errored_at = None
+        found_assignment.cancelled_at = None
+        found_assignment.expired_at = None
```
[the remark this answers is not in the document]
adamstankiewicz: Yeah, I believe the intent of these fields is to represent the timestamp of the most recent state transition/change on the model fields. Any prior state transitions associated with an assignment action (most, if not all, of them?) would still be related to the assignment, whereby one could deduce the prior successful actions' timestamps.

If we don't "clean" these fields on certain state transitions, there will be timestamps on the assignment not relevant to the active/current state which could be misleading for consumers of the model/APIs.

See the [ADR](https://github.com/openedx/enterprise-access/blob/main/docs/decisions/0015-expiration-improvements.rst) for more rationale on the resetting of these fields.

enterprise_access/apps/content_assignments/models.py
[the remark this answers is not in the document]
adamstankiewicz: I agree it would be preferable for `get_automatic_expiration_date_and_reason` and `get_subsidy_expiration` to live within `api.py`. I originally had it there, but then moved it to the models. When trying to replicate the same to see if there was a particular reason, I believe it may have been due to an issue with a circular import:

```shell
  File "/edx/app/enterprise-access/enterprise_access/apps/subsidy_access_policy/models.py", line 18, in <module>
    from enterprise_access.apps.content_assignments import api as assignments_api
  File "/edx/app/enterprise-access/enterprise_access/apps/content_assignments/api.py", line 16, in <module>
    from enterprise_access.apps.content_assignments.tasks import send_reminder_email_for_pending_assignment
  File "/edx/app/enterprise-access/enterprise_access/apps/content_assignments/tasks.py", line 14, in <module>
    from enterprise_access.apps.content_assignments.api import get_automatic_expiration_date_and_reason
ImportError: cannot import name 'get_automatic_expiration_date_and_reason' from partially initialized module 'enterprise_access.apps.content_assignments.api' (most likely due to a circular import) (/edx/app/enterprise-access/enterprise_access/apps/content_assignments/api.py)
```

Maybe we can rubber duck on this for a bit to determine if there's a workaround I'm not seeing to get past this circular import?

enterprise_access/apps/core/constants.py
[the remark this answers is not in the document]
pwnage101: hm, I thought the POST API location under the assignment configuration viewset and URL route?  I know it doesn't write to the assignment configuration, but in my mind the logical placement of the view feels like it takes precedence over the action it actually takes.

enterprise_access/apps/subsidy_access_policy/models.py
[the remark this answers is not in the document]
pwnage101: Gotcha, that makes sense.  Note to self (or to you): double check that we also nullify the other date fields when we re-allocate an pre-existing assignment.

enterprise_access/apps/core/constants.py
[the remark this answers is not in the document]
adamstankiewicz: >  I thought the POST API location was under the assignment configuration viewset and URL route? 

@pwnage101 It is under the assignment configuration viewset/URL route, but so are the other endpoints related to these `CONTENT_ASSIGNMENT_*_PERMISSION` constants. For example, the `LearnerContentAssignmentViewSet` relies on `CONTENT_ASSIGNMENT_LEARNER_READ_PERMISSION`, but the API route is a URL route for the assignment configuration, in a somewhat similar fashion as the `acknowledge-assignments` action endpoint is, e.g.:

```python
router.register(
    r'assignment-configurations/(?P<assignment_configuration_uuid>[\S]+)/assignments',
    views.LearnerContentAssignmentViewSet,
    'assignments',
)
```

enterprise_access/apps/subsidy_access_policy/models.py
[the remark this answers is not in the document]
adamstankiewicz: [confirmation] In the `allocate_assignments` method from `enterprise_access.apps.content_assignments.api`, the date fields are nullified other than `allocated_at` ([source](https://github.com/openedx/enterprise-access/pull/390/files#diff-3fc1b92222fc79cd89b1a91825dd14aa7f78fea14a256bf9ff7d4f3472dfb04bR286-R290)).

enterprise_access/apps/content_assignments/models.py
adamstankiewicz: Updated. Ended up with other circular import issues related to other methods after moving `get_automatic_expiration_date_and_reason` to the `enterprise_access.apps.api.utils` file. I decided to go up one more level to `enterprise_access.utils` instead for now, which does get around the issue as well.

enterprise_access/apps/content_assignments/models.py
adamstankiewicz: Ah, great catch. I've refactored to ensure `get_automatic_expiration_date_and_reason` accepts a `content_metadata` dict. `get_earliest_possible_expiration` in `LearnerContentAssignmentResponseSerializer` and `get_content_metadata` in `LearnerContentAssignmentWithContentMetadataResponseSerializer` both now rely on the Serializer context `content_metadata`.

enterprise_access/apps/core/constants.py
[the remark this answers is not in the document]
pwnage101: In my mind, `/api/v1/assignment-configurations/<uuid>/acknowledge-assignments/` is arguably an action of an assignment configuration, whereas `/api/v1/assignment-configurations/<uuid>/assignments` is a nested route for a different assignment-specific viewset with assignment-specific permissions.  I lean towards calling it an assignment configuration permission rather than an assignment permission because I personally value the route hierarchy more than what the endpoint actually does in this case, but above all I don't want this to be a blocker either!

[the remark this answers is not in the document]
adamstankiewicz: Fair enough! I don't feel strongly one way or another; I think your rationale is sound 😄 I've updated to make this an assignment configuration permission!

pwnage101 (approved):
great job!

iloveagent57 (approved):

PR #398, opened 2024-02-05
feat: ADR for groups implementation

alex-sheehan-edx:
Description
ADR for the groups implementation into LC

https://2u-internal.atlassian.net/browse/ENT-8336

### DIFF
--- a/docs/decisions/0014-access-policy-grouping.rst
+++ b/docs/decisions/0014-access-policy-grouping.rst
@@ -1,0 +1,41 @@
+0014 Access Policy Grouping
+###########################
+
+Status
+******
+**Accepted** Feb 2024
+
+Context
+*******
+The enterprise-access service needs to allow for the flexibility to control and distribute subsidy access policies to
+custom subsets of learners. The subdivision of access policies will improve user management and budgeting from an admin
+perspective, as well as add support for personalization to the learner experience down the line. While it is not
+the responsibility of the service to house the grouping of learners, an access policy must take into account the 
+existence of related subsets when determining if a subsidy is redeemable by an individual learner.
+
+Decision
+********
+The ``SubsidyAccessPolicy`` model will adopt a new, nullable UUID field: `enterprise_group_uuid`. This field will be
+read and factored in at time of calculating redeemability of an individual policy for a learner. The intention is for
+every new access policy to have an assigned group.
+
+How the systems will interact with one another
+++++++++++++++++++++++++++++++++++++++++++++++
+Upon provisioning a new access policy budget for a customer, the service will make a POST request to edx-platform to
+create a new ``EnterpriseGroup`` record. On successful response, the enterprise-access service will write the returned
+UUID of the newly created group to the access policy `enterprise_group_uuid` field.
+
+``SubsidyAccessPolicy``'s `can_redeem()` method already makes a request to edx-platform for
+`enterprise_contains_learner()` in which `lms_user_id` and `enterprise_customer_uuid` are provided to confirm
+a learner's membership with the associated organization. Now, `enterprise_group_uuid` will be optionally supplied by
+`can_redeem()` if it exists on the individual policy record. This value will further filter down the list of learners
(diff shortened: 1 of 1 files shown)


docs/decisions/0014-access-policy-grouping.rst
```diff
@@ -0,0 +1,40 @@
...
+*******
+The enterprise-access service needs to allow for the flexibility to control and distribute subsidy access policies to
+custom subsets of learners. The subdivision of access policies will improve user management and budgeting from an admin
+perspective, as well as add support for personalization to the learner experience down the line. While it is not
+the responsibility of the service to house the grouping of learners, an access policy must take into account the 
+existence of related subsets when determining if a subsidy is redeemable by an individual learner.
+
+Decision
+********
+The ``SubsidyAccessPolicy`` model will adopt a new, nullable UUID field: `enterprise_group_uuid`. This field will be
```
kiram15, line 18: Do we want this to be a nullable field? With this change, by default, every single policy will have a group, and all the learners will be under this group.

docs/decisions/0014-access-policy-grouping.rst
```diff
@@ -0,0 +1,40 @@
...
+
+How the systems will interact with one another
+++++++++++++++++++++++++++++++++++++++++++++++
+Upon provisioning a new access policy budget for a customer, the service will make a POST request to edx-platform to
+create a new ``EnterpriseGroup`` record. On successful response, the enterprise-access service will write the returned
+UUID of the newly created group to the access policy `enterprise_group_uuid` field.
+
+``SubsidyAccessPolicy``'s `can_redeem()` method already makes a request to edx-platform for
+`enterprise_contains_learner()` in which `lms_user_id`` and `enterprise_customer_uuid` are provided to confirm
+a learner's membership with the associated organization. Now, `enterprise_group_uuid` will be optionally supplied by
```
kiram15, line 29: Again, wouldn't this not be optional? We would have to know if the user's group permissions have access to this course

docs/decisions/0014-access-policy-grouping.rst
alex-sheehan-edx: my thought was backwards compatibility- this way we can have a chance to at a minimum backfill all the existing policies with new groups

alex-sheehan-edx: + there's some order of operations stuff going on, ie we need to create the policy then the group then update the policy with the group uuid

docs/decisions/0014-access-policy-grouping.rst
alex-sheehan-edx: optional in that I think it should be backwards compatible

alex-sheehan-edx: so if a policy doesn't have a group it won't blow up

docs/decisions/0014-access-policy-grouping.rst
kiram15: Yeah that makes sense, but I also think we should call out that it really shouldn't be null in practice

alex-sheehan-edx: Sure, I made an addition

johnnagro: yeah i think for backwards compatibility we'll want this to be nullable. We can also discuss different ways to implement the "all users in the enterprise" group which is an implicit group today. One approach is the "blank" or "no group specified" means "all users in the enterprise". Alternatively you can have a magic/dynamic group which is all users (if you feel the need to be explicit with a UUID for "all users" of a given enterprise customer).

johnnagro (approved):

docs/decisions/0014-access-policy-grouping.rst
alex-sheehan-edx: Agreed and I feel that leaving the field nullable keeps us open to either implementation

PR #402, opened 2024-02-09
feat: updated specs for the groups adr to account for multiple groups per policy

alex-sheehan-edx:
We want the flexibility for policies to have multiple groups assigned, adjusted wording and specs of the ADR accordingly


docs/decisions/0018-access-policy-grouping.rst
```diff
@@ -15,21 +15,33 @@ existence of related subsets when determining if a subsidy is redeemable by an i
...
+will define two fields: a non nullable FK `subsidy_access_policy` with related name `groups` and a non nullable UUID
+char field: `enterprise_group_uuid`. A ``SubsidyAccessPolicy``'s `groups` will be read and factored in at time of
+calculating redeemability of an individual policy for a learner. The intention is for every new access policy to have
+at least one assigned group.
+
+PolicyGroupAssociation
+*********************
+**Model properties**
+------
+- uuid, created, modified, history (boilerplate)
```
johnnagro, line 28: we probably dont need history. technically speaking we also don't need a uuid since the policy uuid and group uuid should suffice as a PK

johnnagro (approved):

PR #404, opened 2024-02-09
feat: remove some timing logs, add allocation logs

iloveagent57:
[new PR b/c of weird failing test in github but not local]

* Strips out some logging that was manually timing the allocate endpoint (the endpoint is now performant enough, and we have better tools for this anyway)
* Adds logging around the main function that allocates assignments.


iloveagent57:
I cannot get the tests to fail locally like they are in the actions on this PR.  They all pass. Even when I run them in random order (using `pytest-random-order`).

pwnage101 (approved):

PR #411, opened 2024-02-23
feat: update enterprise_contains_learner return value

katrinan029:
https://2u-internal.atlassian.net/browse/ENT-8416

### DIFF
--- a/enterprise_access/apps/api_client/lms_client.py
+++ b/enterprise_access/apps/api_client/lms_client.py
@@ -113,7 +113,7 @@ class LmsApiClient(BaseOAuthClient):
             logger.exception(msg, enterprise_customer_uuid)
             raise
 
-    def enterprise_contains_learner(self, enterprise_customer_uuid, learner_id):
+    def get_enterprise_user(self, enterprise_customer_uuid, learner_id):
         """
         Verify if `learner_id` is a part of an enterprise represented by `enterprise_customer_uuid`.
 
@@ -122,10 +122,8 @@ class LmsApiClient(BaseOAuthClient):
             learner_id (int): LMS user id of a learner.
 
         Returns:
-            bool: True if learner is linked with enterprise else False
+            None or the enterprise customer user record
         """
-
-        result = False
         ec_uuid = str(enterprise_customer_uuid)
         query_params = {'enterprise_customer_uuid': ec_uuid, 'user_ids': learner_id}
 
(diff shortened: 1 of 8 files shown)


enterprise_access/apps/api_client/lms_client.py
```diff
@@ -122,10 +122,10 @@ def enterprise_contains_learner(self, enterprise_customer_uuid, learner_id):
             learner_id (int): LMS user id of a learner.
 
         Returns:
-            bool: True if learner is linked with enterprise else False
```
kiram15, line 125: I would maybe change the method name to be fetch_enterprise_user_record or something like that cause the name still kind of implies that you're getting a boolean back

kiram15 (approved):
One non blocking nit!

enterprise_access/apps/api_client/lms_client.py
katrinan029: Thanks for the feedback. I'm leaning towards keeping the function name as is because the existing consumers are still expecting a boolean. @alex-sheehan-edx would love for you to weigh in on this.

kiram15: No that does make sense, more just a nit not a blocking thing

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -137,7 +137,7 @@ def enterprise_contains_learner(self, enterprise_customer_uuid, learner_id):
             results = json_response.get('results')
             results = results and results[0]
             if results and results['enterprise_customer']['uuid'] == ec_uuid and results['user']['id'] == learner_id:
-                result = True
+                result = results
```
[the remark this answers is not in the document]
alex-sheehan-edx: there's some weird assumptions here about multiple learner records which doesn't make a lot of sense to me since we're filtering by user ID _**and**_  customer... so when could there ever be more than one? Unless it's an empty array sort of situation? But I would have to look closer

alex-sheehan-edx (commented):
Great start, I think maybe we could rip out some more of the truthy based logic in the api client. LMK if you have any questions!

enterprise_access/apps/api_client/lms_client.py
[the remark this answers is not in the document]
kiram15: @alex-sheehan-edx There could be a situation where the person is a member of multiple groups in the same enterprise

alex-sheehan-edx (approved):
One small suggestion but otherwise lgtm

PR #413, opened 2024-02-27
feat: adding enterprise group checks in can_redeem

kiram15:
Updating the can_redeem function with an additional check to ensure that if the enterprise has groups, the learner is a member of one of those groups. 

[Jira Ticket ](https://2u-internal.atlassian.net/browse/ENT-8417)

PR #419, opened 2024-03-04
feat: adding enterprise group checks in can_redeem

kiram15:
Re-merging in the [changes](https://github.com/openedx/enterprise-access/pull/413) to the can_redeem function that were [reverted](https://github.com/openedx/enterprise-access/pull/418) because of an error. 

Original [Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-8417)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -591,10 +592,29 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
...
         # learner not associated to enterprise
         if not skip_customer_user_check:
-            if self.lms_api_client.get_enterprise_user(self.enterprise_customer_uuid, lms_user_id) is None:
+            learner_record = self.lms_api_client.get_enterprise_user(self.enterprise_customer_uuid, lms_user_id)
+            if not learner_record:
                 self._log_redeemability(False, REASON_LEARNER_NOT_IN_ENTERPRISE, lms_user_id, content_key)
                 return (False, REASON_LEARNER_NOT_IN_ENTERPRISE, [])
 
+            associated_group_uuids = set(  # pylint: disable=consider-using-set-comprehension
+                [str(group.get('uuid')) for group in learner_record.get('enterprise_group')]
```
alex-sheehan-edx, line 601: I would add a default value for the learner_record getter- otherwise a None value returned may be a problem:
```
>>> yo = [foo for foo in ['b', 'a', 'r']]
>>> yo
['b', 'a', 'r']
>>> yo_2 = [foo for foo in None]
Traceback (most recent call last):
  File "<console>", line 1, in <module>
TypeError: 'NoneType' object is not iterable
```

alex-sheehan-edx (approved):
one question/nit otherwise looks good

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -1066,11 +1067,16 @@ def setup_mocks(self):
         self.addCleanup(contains_key_patcher.stop)
         self.addCleanup(get_content_metadata_patcher.stop)
 
+    @mock.patch('enterprise_access.apps.api_client.base_oauth.OAuthAPIClient')
```
alex-sheehan-edx, line 1070: I get why/how this mock would be used but I'm curious why it was working before this change?

enterprise_access/apps/subsidy_access_policy/models.py
alex-sheehan-edx, line 601: With the release of https://github.com/openedx/edx-enterprise/pull/2040 we'll have to update this logic to account for the new group data structure in the enterprise customer serializer

alex-sheehan-edx (changes requested):
updates needed with new enterprise code

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -591,10 +592,27 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
...
+
+            # if there are groups associated with the policy, if not we want to be backwards compatible with
+            # policies which haven't implemented groups yet
+            if len(policy_groups) > 0:
+                uuid_overlap = associated_group_uuids.intersection(policy_groups_uuids)
+                # if any of the user's associated groups are connected to the policy
+                if len(uuid_overlap) == 0:
+                    self._log_redeemability(False, REASON_LEARNER_NOT_IN_ENTERPRISE_GROUP, lms_user_id, content_key)
+                    return (False, REASON_LEARNER_NOT_IN_ENTERPRISE_GROUP, [])
+
```
iloveagent57, line 615: It would be neat to extract this entire chunk of code into a method like `SubsidyAccessPolicy.includes_learner(lms_user_id)` - then we could reuse this logic down in the `credit_available()` method of this same model.  You could also simplify this a little bit by relying purely on the Django ORM to do the intersection for you
```python
def includes_learner(self, lms_user_id):
    learner_record = self.lms_api_client.get_enterprise_user(self.enterprise_customer_uuid, lms_user_id)
    if not learner_record:
        return False, REASON_LEARNER_NOT_IN_ENTERPRISE

    associated_group_uuids = set(learner_record.get('enterprise_group', []))
    # if there are no policy groups, return early
    if not PolicyGroupAssociation.objects.filter(subsidy_access_policy=self).exists():
        return True, None

    # if no association for this learner's group(s), return false
    if not PolicyGroupAssociation.objects.filter(
        subsidy_access_policy=self, 
        enterprise_group_id__in=associated_group_uuids,
    ).exists():
        return False, REASON_LEARNER_NOT_IN_ENTERPRISE_GROUP

    # otherwise, return true
    return True, None
```
then up here in `can_redeem()`:
```python
if not skip_customer_user_check:
    included_in_policy, reason = self.includes_learner(lms_user_id):
    if not included_in_policy:
        self._log_redeemability(False, reason, lms_user_id, content_key)
        return (False, reason, [])
```

iloveagent57 (commented):
A couple other things you might want to do in addition to the suggested refactoring:
1. `skip_customer_user_check` is passed with a value of (edit: `True`) ~False~ from the [can-redeem view](https://github.com/openedx/enterprise-access/blob/056f2ef4ca38abd57643f01df7fed32e82c0b659/enterprise_access/apps/api/v1/views/subsidy_access_policy.py#L428). With groups in play, I think we'll have to toggle that to `True` (or maybe remove the notion of `skip_customer_user_check` entirely. There's some performance implications to that, which I've captured a potential resolution for in https://2u-internal.atlassian.net/browse/ENT-8603.
2. You probably want to do this same group inclusion check in the `credit_available` method in this same file - it's currently checking only that the user is associated with the customer. That method is used by the [credits_available view](https://github.com/openedx/enterprise-access/blob/056f2ef4ca38abd57643f01df7fed32e82c0b659/enterprise_access/apps/api/v1/views/subsidy_access_policy.py#L468), which the learner-portal frontend relies on for determining which policies/budgets are available to a learner.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -694,6 +727,14 @@ def credit_available(
                     self.enterprise_customer_uuid
                 )
                 return False
+            included_in_policy, reason = self.includes_learner(lms_user_id)
+            if not included_in_policy:
+                logger.info(
+                    '[credit_available] learner %s encountered error %s',
```
alex-sheehan-edx, line 733: Smallest of nits- I've been trying to replace/use fstrings over formatting

alex-sheehan-edx (approved):
Tiny syntax nit but looks great

PR #427, opened 2024-03-27
docs: proposed ADR for forced redemption

iloveagent57:
There is frequently a need to force through a redemption (and related enrollment/fulfillment)of a particular learner, covered by a particular subsidy access policy, into some specific course run. This ADR proposes a model and methodology for achieving such a redemption.

### DIFF
--- a/docs/decisions/0019-forced-redemption.rst
+++ b/docs/decisions/0019-forced-redemption.rst
@@ -1,0 +1,54 @@
+0019 Forced Redemption
+**********************
+
+Status
+======
+
+Accepted - April 2024
+
+Context
+=======
+There is frequently a need to force through a redemption (and related enrollment/fulfillment)
+of a particular learner, covered by a particular subsidy access policy, into
+some specific course run. This need exists for reasons related to upstream
+business constraints, notably in cases where a course is included in a policy's catalog,
+but the desired course *run* is not discoverable due to the current state of its metadata
+(e.g. the ``is_enrollable`` value for that run's record is false).
+
+Some recent work has provided an ``allow_late_enrollment`` directive to be included
+in the ``metadata`` of the subsidy redeem payload.  This solution will utilize that directive,
+because it causes enrollment and/or external fulfillment to be forced through in
+downstream systems.
+
+Decision
+========
+* We'll expose a model, ``ForcedPolicyRedemption``, via Django Admin that allows
+  staff users to force redemption for a given learner, into a particular course,
+  under a particular subsidy access policy.
+* The model should describe which learner is being enrolled, via which policy, and into
+  which specific course **run**.  It should also record which staff user wrote the record
+  when it was created/modified, and when the redemption succeeded or failed.
+  Furthermore, it should store a reference to the created
+  transaction on success, and store a reference to any error information on failure.
+* It should be possible to write a record for this model *without* the forced redemption immediately
+  taking place, although perhaps the default behavior should cause the redemption to
+  occur on save.
+* All the standard redeemability logic will still hold, that is, ``can_redeem()`` must return ``True``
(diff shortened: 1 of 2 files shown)


docs/decisions/0019-forced-redemption.rst
```diff
@@ -0,0 +1,48 @@
...
+  transaction on success, and store a reference to any error information on failure.
+* It should be possible to write a record for this model *without* the forced redemption immediately
+  taking place, although perhaps the default behavior should cause the redemption to
+  occur on save.
+* All the standard redeemability logic will still hold, that is, ``can_redeem()`` must return ``True``
+  for the redemption to take place.  This entails the automated creation of a ``LearnerContentAssignment``
+  record for assignment-based policies.
+
+Consequences
+============
```
pwnage101, line 39: Should this clarify that the tool will not work to force Exec Ed redemptions due to the T's&C's flow requiring learner intervention?

pwnage101: Or is there more to this plan specifically around Exec Ed courses not yet written?  E.g. I can imagine the new force redemption model could indicate intent to force a specific learner-policy-run redemption, then an accompanying UI component treats it like an assignment (even if the policy type is auto-apply).  But this is completely made up and I'm just using my imagination to fill in the gaps.

iloveagent57: Updated to reflect that this won't work for EE content.

docs/decisions/0019-forced-redemption.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Accepted - April 2024
+
+Context
+=======
+There is frequently a need to force through a redemption (and related enrollment/fulfillment)
+of a particular learner, covered by a particular subsidy access policy, into
+some specific course run. This needs exists for reasons related to upstream
+business constraints, notably in cases where a course is included in a policy's catalog,
+but the desired course *run* is not discoverable due to the current state of its metadata.
```
adamstankiewicz, line 15: nit: would it be helpful to include a mention of what specific metadata is of concern here (e.g., `is_enrollable`)?

docs/decisions/0019-forced-redemption.rst
```diff
@@ -0,0 +1,52 @@
...
+This tool will **not** suffice to force redemption for Exec Ed content, or any other
+content type for which the enterprise-subsidy service requires additional context
+in the ``metadata`` element of the redeem payload.
+
+Alternatives Considered
+=======================
+There is ongoing work to allow late policy-redemption under some circumstances
+in a way that's available to the learner-facing UX.  This proposal builds on that
+(particularly the ``allow_late_enrollment`` directive), but is useful in scenarios
+where the desired content is not discoverable to the learner (and therefor not enrollable).
```
adamstankiewicz, line 52: nit: typo 

```suggestion
where the desired content is not discoverable to the learner (and therefore not enrollable).
```

docs/decisions/0019-forced-redemption.rst
```diff
@@ -0,0 +1,52 @@
...
+but the desired course *run* is not discoverable due to the current state of its metadata.
+
+Some recent work has provided an ``allow_late_enrollment`` directive to be included
+in the ``metadata`` of the subsidy redeem payload.  This solution will utilize that directive,
+because it causes enrollment and/or external fulfillment to be forced through in
+downstream systems.
+
+Decision
+========
+* We'll expose a model via Django Admin that allows staff users to force redemption
```
adamstankiewicz, line 24: nit: Should this include the name of the proposed model `ForcedPolicyRedemption`? The updated README below mentions the creation of the new model `ForcedPolicyRedemption`, but the ADR itself doesn't explicitly mention the model name.

adamstankiewicz (approved):
LGTM, with a handful minor nits/suggestions.

PR #435, opened 2024-04-05
feat: usability improvements to forced redemption admin experience

iloveagent57:
ENT-8703

http://localhost:18270/admin/subsidy_access_policy/forcedpolicyredemption/
* Actually _does_ the forced redemption on admin model save (under certain conditions)
* Provides some messaging to the admin user about success/failure/warnings around forced-redemption


enterprise_access/apps/subsidy_access_policy/admin/__init__.py
```diff
@@ -375,4 +376,35 @@ class ForcedPolicyRedemptionAdmin(DjangoQLSearchMixin, SimpleHistoryAdmin):
...
+        If this record has not been successfully redeemed yet,
+        and if ``wait_to_redeem`` is false, then call ``force_redeem()`` on
+        the record.
+        """
+        super().save_model(request, obj, form, change)
+        obj.refresh_from_db()
+
+        if obj.transaction_uuid:
+            message = gettext_lazy("{} has already been redeemed".format(obj))
+            self.message_user(request, message, messages.SUCCESS)
```
adamstankiewicz, line 394: Nice uses of `self.message_user`! 😄

enterprise_access/apps/subsidy_access_policy/admin/__init__.py
```diff
@@ -375,4 +376,35 @@ class ForcedPolicyRedemptionAdmin(DjangoQLSearchMixin, SimpleHistoryAdmin):
...
+            return
+
+        if obj.wait_to_redeem:
+            message = gettext_lazy(
+                "{} has wait_to_redeem set to true, redemption will not occur "
```
adamstankiewicz, line 399: [question/clarification] Unrelated to the specific changes, just curious if you can shed some addtl light on `wait_to_redeem`? What is the expected behavior when `wait_to_redeem` is true?

adamstankiewicz (approved):

enterprise_access/apps/subsidy_access_policy/admin/__init__.py
iloveagent57: Just that `force_redeem()` won't be called - it's a way to create a record without actually executing the forced redemption.

adamstankiewicz: Right. I suppose I'm asking what the use case is to create a record without executing the forced redemption straight away?

PR #443, opened 2024-04-22
fix: uniquify retired email addresses

iloveagent57:
Prior to this change, the `LearnerContentAssignment.clear_pii()` method set a retired email address to the same static value.  This means that if two different learners (by email) in the same assignment configuration (i.e. budget) have the same course assigned, and then those records are expired, the second record to be expired will fail on the unique constraint violation of the model.
This change adds a random hex-string to each retired email address value to help resolve this issue. ENT-8574

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/models.py
+++ b/enterprise_access/apps/content_assignments/models.py
@@ -1,8 +1,8 @@
 """
 Models for content_assignments
 """
-
 import logging
+from os import urandom
 from uuid import UUID, uuid4
 
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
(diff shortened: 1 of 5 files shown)


enterprise_access/apps/content_assignments/models.py
```diff
@@ -675,17 +675,30 @@ def add_successful_acknowledged_expired_action(self):
...
         """
-        self.learner_email = RETIRED_EMAIL_ADDRESS
+        self.learner_email = self._unique_retired_email()
 
     def clear_historical_pii(self):
         """
         Removes PII values from this assignment's historical records.
         """
-        self.history.update(learner_email=RETIRED_EMAIL_ADDRESS)  # pylint: disable=no-member
+        self.history.update(learner_email=self._unique_retired_email())  # pylint: disable=no-member
```
adamstankiewicz, line 701: [sanity check, question] Are there any concerns around the history tables' redacted email addresses likely not matching the non-history tables' redacted email for the same assignment due to the random hex string?

adamstankiewicz (approved):

enterprise_access/apps/content_assignments/models.py
iloveagent57: Good suggestion, I just modified `clear_pii()` to retire the main record and the historical records with the same value.

PR #444, opened 2024-04-23
fix: assignment nudge emails are fixed to initial run

pwnage101:
Assignments now record a "preferred" course run key, defined as the advertised run at the time of allocation. This preferred run is then used by nudge email logic to determine the start date, thereby preventing subsequent runs from re-triggering nudge emails on lingering assignments.

ENT-8789


enterprise_access/apps/content_assignments/models.py
```diff
@@ -304,6 +304,16 @@ class Meta:
         blank=False,
         help_text="Cost of the content in USD Cents.",
     )
+    preferred_course_run_key = models.CharField(
+        max_length=255,
+        blank=False,
```
iloveagent57, line 309: you'll want `blank=True` on this

iloveagent57 (commented):
Good idea!

enterprise_access/apps/content_assignments/models.py
```diff
@@ -304,6 +304,17 @@ class Meta:
...
     )
+    preferred_course_run_key = models.CharField(
+        max_length=255,
+        blank=True,
+        null=True,
+        db_index=True,
+        help_text=(
+            "The preferred course run that the admin primarily intends for the learner, and the one used to control "
+            "nudge emails. This is automatically set at assignment creation time, and is never changed again (even "
+            "during re-allocation)."
```
iloveagent57, line 315: _Should_ we change it during re-allocation?

iloveagent57: The downside is that, if we don't change it during re-allocation, no nudges will go out for the assignment, right? I don't have strong feelings either way - maybe just add a note to the mgmt command's docstring.

pwnage101: yea, this can happen:

* day 0: run A is the latest run, and admin assigns course.
* day 1: admin cancels assignment.
* day 2: run B is published.
* day 3: admin re-assigns the same learner to the same course.

Result: nudge emails will correspond to run A despite run B being the newest run at the time of re-assignment.

pwnage101: after discussing with beggs, I'll just go ahead and add this feature in.

iloveagent57 (approved):

PR #445, opened 2024-04-24
feat: Add TieredCaching for enterprise_contains_learner

brobro10000:
**Description:**

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/lms_client_api.py
+++ b/enterprise_access/apps/subsidy_access_policy/lms_client_api.py
@@ -1,0 +1,51 @@
+"""
+Python API for interacting with the lms client
+for use in the domain of SubsidyAccessPolicies.
+"""
+import logging
+
+from django.conf import settings
+from edx_django_utils.cache import TieredCache
+from requests.exceptions import HTTPError
+
+from enterprise_access.cache_utils import versioned_cache_key
+
+from enterprise_access.apps.api_client.lms_client import LmsApiClient
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_CACHE_TIMEOUT = getattr(settings, 'LMS_CLIENT_CACHE_TIMEOUT', 60 * 5)
+
+
+def get_and_cache_enterprise_contains_learner(enterprise_customer_uuid, learner_id, timeout=None):
+    """
+    Determines if the learner identified by the 'learner_id is linked
+    to a specified enterprise identified by the 'enterprise_customer_uuid
+
+    Returns: A bool to specify if the learner is associated to the enterprise
+    """
+
+    cache_key = versioned_cache_key('get_enterprise_contains_learner', enterprise_customer_uuid, learner_id)
+    cached_response = TieredCache.get_cached_response(cache_key)
+    if cached_response.is_found:
+        logger.info(f'Cache hit for customer {enterprise_customer_uuid} and learner id {learner_id}')
+        return cached_response.value
+
+    lms_client = LmsApiClient()
+    try:
+        is_learner_linked_to_enterprise = lms_client.enterprise_contains_learner(
+            enterprise_customer_uuid,
+            learner_id
+        )
+    except HTTPError as exc:
+        raise exc
+
+    logger.info(
+        'Fetched enterprise contains learner for customer %s and learner_id %s',
+        enterprise_customer_uuid,
+        learner_id
+    )
+    TieredCache.set_all_tiers(cache_key, is_learner_linked_to_enterprise, timeout or DEFAULT_CACHE_TIMEOUT)
+    return is_learner_linked_to_enterprise
+
+
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -43,6 +43,9 @@ from .content_metadata_api import (
     get_list_price_for_content,
     list_price_dict_from_usd_cents
 )
+from .lms_client_api import (
+    get_and_cache_enterprise_contains_learner
+)
 from .exceptions import (
     ContentPriceNullException,
     MissingAssignment,
@@ -362,6 +365,16 @@ class SubsidyAccessPolicy(TimeStampedModel):
         set_tiered_cache_subsidy_record(record, *cache_key_args)
         return record
 
+    def includes_user(self, lms_user_id):
+        """
+        Determines if the user is included in the SubsidyAccessPolicy's customer.
+        Retrieves it from TieredCache if available, otherwise, it will retrieve and initialize the cache.
+        """
+        return get_and_cache_enterprise_contains_learner(
+            enterprise_customer_uuid=self.enterprise_customer_uuid,
+            learner_id=lms_user_id,
+        )
+
     def subsidy_balance(self):
         """
         Returns total remaining balance for the associated subsidy ledger.
@@ -576,7 +589,7 @@ class SubsidyAccessPolicy(TimeStampedModel):
 
         # learner not associated to enterprise
         if not skip_customer_user_check:
-            if not self.lms_api_client.enterprise_contains_learner(self.enterprise_customer_uuid, lms_user_id):
+            if not self.includes_user(lms_user_id):
                 return (False, REASON_LEARNER_NOT_IN_ENTERPRISE, [])
 
         # no content key in catalog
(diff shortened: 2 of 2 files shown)


enterprise_access/apps/subsidy_access_policy/lms_client_api.py
```diff
@@ -0,0 +1,51 @@
...
+from enterprise_access.cache_utils import versioned_cache_key
+
+from enterprise_access.apps.api_client.lms_client import LmsApiClient
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_CACHE_TIMEOUT = getattr(settings, 'LMS_CLIENT_CACHE_TIMEOUT', 60 * 5)
+
+
+def get_and_cache_enterprise_contains_learner(enterprise_customer_uuid, learner_id, timeout=None):
```
iloveagent57, line 20: nit: maybe do `timeout=DEFAULT_CACHE_TIMEOUT`?

enterprise_access/apps/subsidy_access_policy/lms_client_api.py
```diff
@@ -0,0 +1,51 @@
...
+
+    cache_key = versioned_cache_key('get_enterprise_contains_learner', enterprise_customer_uuid, learner_id)
+    cached_response = TieredCache.get_cached_response(cache_key)
+    if cached_response.is_found:
+        logger.info(f'Cache hit for customer {enterprise_customer_uuid} and learner id {learner_id}')
+        return cached_response.value
+
+    lms_client = LmsApiClient()
+    try:
+        is_learner_linked_to_enterprise = lms_client.enterprise_contains_learner(
```
iloveagent57, line 36: note: `enterprise_contains_learner()` doesn't actually exist (I think) - use the `get_enterprise_user()` method.

iloveagent57 (commented):
Here's how I would tweak this a little:
1. rename `lms_client_api.py` to `customer_api.py`.
2. Change `get_and_cache_enterprise_contains_learner()` to `get_and_cache_enterprise_learner_record(...)`. This method should always return either a dictionary of learner data (for the enterprise), or null.
3. Let your `SubsidyAccessPolicy.includes_user()` method make use of that function and return a boolean.

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -1091,10 +1091,16 @@ def setup_mocks(self):
...
+        includes_user_patcher = patch.object(
+            SubsidyAccessPolicy, 'includes_user'
+        )
+        self.mock_includes_user = includes_user_patcher.start()
+
         self.addCleanup(lms_client_patcher.stop)
         self.addCleanup(subsidy_client_patcher.stop)
         self.addCleanup(contains_key_patcher.stop)
         self.addCleanup(get_content_metadata_patcher.stop)
+        self.addCleanup(includes_user_patcher.stop)
```
pwnage101, line 1103: Seems odd to me that we'd be patching this `includes_user` function which is supposed to always return a boolean, but then not setting up a mock return value. To me, it's arbitrary what the return value of a mock function call would be before assigning `.return_value`....is it truthy or falsey?  It seems nondeterministic that it should be one or the other, and makes our unit test sensitive to upstream changes to the unit test framework libraries.

My gut is that we go ahead and toss this in:

```python
        self.mock_includes_user = includes_user_patcher.start()
        self.mock_includes_user.return_value = True
```

enterprise_access/apps/subsidy_access_policy/tests/mixins.py
```diff
@@ -41,11 +41,19 @@ def setUp(self):
...
             SubsidyAccessPolicy, 'lms_api_client'
         )
+
         self.mock_lms_api_client = lms_api_client_patcher.start()
 
+        includes_user_patcher = patch.object(
+            SubsidyAccessPolicy, 'includes_user'
+        )
+
+        self.mock_includes_user = includes_user_patcher.start()
```
pwnage101, line 51: I see here that you're setting up the patcher within the mixin, then in a test that consumes this mixin you're assigning a mock return value.  That's a good pattern.  However, again I'm noticing that not all unit tests which call can_redeem() + includes_user() actually has this mock configured with a return value.  For convenience, I recommend also setting the default return_value to `True` here, like the other place this is patched.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -384,6 +385,18 @@ def subsidy_record_from_tiered_cache(self, *cache_key_args):
...
+        """
+        Determines if the user is included in the SubsidyAccessPolicy's customer.
+        Retrieves it from TieredCache if available, otherwise, it will retrieve and initialize the cache.
+        """
+        if get_and_cache_enterprise_learner_record(
+            enterprise_customer_uuid=self.enterprise_customer_uuid,
+            learner_id=lms_user_id,
+        ) is None:
+            return False
+        return True
```
pwnage101, line 398:
```suggestion
        return get_and_cache_enterprise_learner_record(
            enterprise_customer_uuid=self.enterprise_customer_uuid,
            learner_id=lms_user_id,
        ) is not None
```

enterprise_access/apps/subsidy_access_policy/customer_api.py
```diff
@@ -0,0 +1,48 @@
...
+from django.conf import settings
+from edx_django_utils.cache import TieredCache
+from requests.exceptions import HTTPError
+
+from enterprise_access.apps.api_client.lms_client import LmsApiClient
+from enterprise_access.cache_utils import versioned_cache_key
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_CACHE_TIMEOUT = getattr(settings, 'LMS_CLIENT_CACHE_TIMEOUT', 60 * 5)
```
iloveagent57, line 16: minor, non-blocking suggestions:
1. Set a default value for this setting in `settings/base.py`. 5 minutes is fine.  There's already a [section for cache timeouts](https://github.com/openedx/enterprise-access/blob/75363c6683e7e09f5dfb9ffd3ba5909584a2416a/enterprise_access/settings/base.py#L497).
2. Consider making the setting name more specific, e.g. `ENTERPRISE_USER_RECORD_CACHE_TIMEOUT`.

enterprise_access/apps/subsidy_access_policy/customer_api.py
```diff
@@ -0,0 +1,48 @@
...
+
+    cache_key = versioned_cache_key('get_enterprise_contains_learner', enterprise_customer_uuid, learner_id)
+    cached_response = TieredCache.get_cached_response(cache_key)
+    if cached_response.is_found:
+        logger.info(f'Cache hit for customer {enterprise_customer_uuid} and learner id {learner_id}')
+        return cached_response.value
+
+    lms_client = LmsApiClient()
+    try:
+        is_learner_linked_to_enterprise = lms_client.get_enterprise_user(
```
iloveagent57, line 35: Nit: this variable name could be a little confusing, `get_enterprise_user` returns either a dict of customer-user data, or `None`. And you want this method we're in right now to return the same.

enterprise_access/apps/subsidy_access_policy/models.py
iloveagent57: or if you like depending on truthiness, you could do it as:
```
learner_record = get_and_cache_enterprise_learner_record(
    learner_id=lms_user_id,
)
return bool(learner_record)
```

enterprise_access/apps/subsidy_access_policy/customer_api.py
```diff
@@ -0,0 +1,48 @@
...
+            enterprise_customer_uuid=enterprise_customer_uuid,
+            learner_id=learner_id
+        )
+    except HTTPError as exc:
+        raise exc
+
+    logger.info(
+        'Fetched enterprise contains learner for customer %s and learner_id %s',
```
iloveagent57, line 43: Nit: really something like "fetched enterprise customer learner record for ..."

enterprise_access/apps/subsidy_access_policy/customer_api.py
```diff
@@ -0,0 +1,48 @@
...
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_CACHE_TIMEOUT = getattr(settings, 'LMS_CLIENT_CACHE_TIMEOUT', 60 * 5)
+
+
+def get_and_cache_enterprise_learner_record(enterprise_customer_uuid, learner_id, timeout=DEFAULT_CACHE_TIMEOUT):
+    """
+    Determines if the learner identified by the 'learner_id is linked
+    to a specified enterprise identified by the 'enterprise_customer_uuid
```
iloveagent57, line 22: nit: it's doing even more than that, it's fetching the learner record, if it exists.

iloveagent57 (approved):
Looks great, left a few suggestions to enhance clarity.

PR #446, opened 2024-04-24
feat: new subsidy access policy api endpoint to fetch group members data hydrated with aggregated enrollment count

alex-sheehan-edx:
**Description:**
Adds new endpoint to surface subsidy aggregate data hydrated group membership records 

**Jira:**
https://2u-internal.atlassian.net/browse/ENT-8760

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


requirements/base.txt
```diff
@@ -167,7 +168,7 @@ edx-drf-extensions==10.3.0
     # via
     #   -r requirements/base.in
     #   edx-rbac
-edx-enterprise-subsidy-client==0.4.2
+edx-enterprise-subsidy-client==0.4.3
```
alex-sheehan-edx, line 171: Not this change is in a [PR](https://github.com/openedx/edx-enterprise-subsidy-client/pull/83) and not currently deployed

alex-sheehan-edx: things are released now!

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -115,6 +127,52 @@ def unlink_users_from_enterprise(self, enterprise_customer_uuid, user_emails, is
...
+        if page:
+            params['page'] = page
+
+        response = self.client.get(group_members_url, params=params)
+        response.raise_for_status()
+        response_json = response.json()
+        results = response_json.get('results', [])
+        if not page:
+            next_page = response.json().get("next")
+            while next_page:
```
iloveagent57, line 166: ah, I see. I would make this more explicit by allowing an optional `traverse_pagination` parameter of this method, and maybe the view. Presumably our frontends won't actually want to kick off a loop of requests via the backend.

enterprise_access/apps/subsidy_access_policy/subsidy_api.py
```diff
@@ -30,6 +30,38 @@ class TransactionPolicyMismatchError(Exception):
...
+def subsidy_learner_aggregate_data_cache_key(subsidy_uuid, policy_uuid=None):
+    return versioned_cache_key('get_subsidy_learners_aggregate_data', subsidy_uuid, policy_uuid)
+
+
+def get_and_cache_subsidy_learners_aggregate_data(subsidy_uuid, policy_uuid=None):
+    """
+    Get aggregated learner data for a given subsidy. This can be optionally further filtered
+    """
+    cache_key = subsidy_learner_aggregate_data_cache_key(subsidy_uuid, policy_uuid)
+    cached_response = request_cache(namespace=REQUEST_CACHE_NAMESPACE).get_cached_response(cache_key)
```
iloveagent57, line 42: The request cache will only cache results for the lifetime of a single request - you'll want to use tiered cache to store this data via memcached for some amount of time, primarily to help performance when multiple pages of members data exists.

enterprise_access/apps/subsidy_access_policy/subsidy_api.py
```diff
@@ -30,6 +30,38 @@ class TransactionPolicyMismatchError(Exception):
...
+    if cached_response.is_found:
+        logger.info(
+            f'subsidy_learners_aggregate_data cache hit for subsidy {subsidy_uuid} and policy {policy_uuid}'
+        )
+        return cached_response.value
+
+    client = get_versioned_subsidy_client(version=1)
```
iloveagent57, line 49: Could we just expose this new endpoint in version 2 of the API as well?

iloveagent57: scratch that, that's a different chunk of work to address

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -712,3 +712,79 @@ class SubsidyAccessPolicyAllocationResponseSerializer(serializers.Serializer):
...
+
+class SubsidyHydratedGroupMemberDataResponseSerializer(serializers.Serializer):
+    """
+    A read-only serializer for responding to requests to the ``get_subsidy_hydrated_group_member_data`` endpoint.
+
+    For view: SubsidyAccessPolicyGroupViewset.get_subsidy_hydrated_group_member_data
+    """
+    learner_id = serializers.IntegerField(
+        help_text=('LMS enterprise user ID associated with the membership.'),
+    )
```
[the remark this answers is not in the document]
alex-sheehan-edx: this is ECU ID currently... I agree it should be more clear what the value is reading, it already lead to me crossing wires just setting everything up.

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -712,3 +712,79 @@ class SubsidyAccessPolicyAllocationResponseSerializer(serializers.Serializer):
...
+    recent_action = serializers.CharField(
+        help_text='String representation of the most recent action associated with the creation of the membership.',
+    )
+    status = serializers.CharField(
+        help_text='String representation of the current membership status.',
+    )
+    activated_at = serializers.DateTimeField(
+        help_text='Datetime at which the membership and user record when from pending to realized.',
+    )
```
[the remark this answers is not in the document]
alex-sheehan-edx: each field represents a column of the datatable - member details columns contains both parts of members details field so I kept them as one nested field

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -712,3 +712,79 @@ class SubsidyAccessPolicyAllocationResponseSerializer(serializers.Serializer):
...
+    enrollment_count = serializers.IntegerField(
+        help_text=(
+            'Number of enrollments associated with the subsidy connected to the group made by the group member'
```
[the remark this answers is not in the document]
alex-sheehan-edx: correct!

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -851,3 +856,145 @@ def allocate(self, request, *args, **kwargs):
                 }
             ]
             raise AllocationRequestException(detail=error_detail) from exc
+
+
+class SubsidyHydratedGroupMembersCsvRenderer (CSVRenderer):
```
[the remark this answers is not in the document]
alex-sheehan-edx: 🤦

enterprise_access/apps/subsidy_access_policy/subsidy_api.py
alex-sheehan-edx: talked offline and decided that a larger migration of policy related views was needed and that's slightly beyond the scope of this ticket/work and TBD

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -851,3 +856,145 @@ def allocate(self, request, *args, **kwargs):
...
+    labels = {
+        'member_details.user_email': 'email',
+        'member_details.user_name': 'name',
+        'recent_action': 'Recent Action',
+        'enrollment_count': 'Enrollment Number',
+        'activated_at': 'Activation Date',
+    }
+
+
+class SubsidyAccessPolicyGroupViewset(UserDetailsFromJwtMixin, PermissionRequiredMixin, viewsets.GenericViewSet):
```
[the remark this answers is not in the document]
alex-sheehan-edx: talked offline- to summarize: If I end up moving the viewset to it’s own file, something like groups.py I think I’d probably want to change the route as a whole. My initial reasoning was that the whole thing was revolving around a requested access policy. And so `/subsidy-access-policies/<uuid>/group-members/` made sense for a policy aggregated group members list. But if it’s in it’s own file/viewset category of “groups” then the path would probably be something like /group/<group uuid>/policy-aggregate. Not to say that a file has to correspond to a unique path format/structure, but if we are considering reformatting to make things more readable/approachable then the singular file is slightly more clear.

enterprise_access/apps/subsidy_access_policy/subsidy_api.py
alex-sheehan-edx: I didn't realize this was only relevant to the lifetime of the request- nice catch

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -712,3 +712,103 @@ class SubsidyAccessPolicyAllocationResponseSerializer(serializers.Serializer):
             'learner email(s), and content for this action.'
         ),
     )
+
+
+class GroupMembersWithAggregatesMemberDetailsSerializer(serializers.Serializer):
```
iloveagent57, line 717: nit: This could just be named `GroupMembersDetailsSerializer` - there's nothing specific to aggregates in here (even though it is *consumed* by an aggregates serializer - but there's other parent serializers that might want this as a child, too).

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -851,3 +856,145 @@ def allocate(self, request, *args, **kwargs):
...
+        if request_serializer.validated_data.get('format_csv', False):
+            request.accepted_renderer = SubsidyHydratedGroupMembersCsvRenderer()
+            request.accepted_media_type = SubsidyHydratedGroupMembersCsvRenderer().media_type
+            return Response(list(member_results), status=status.HTTP_200_OK, content_type='text/csv')
+
+        # Since we are essentially forwarding all request params to platform, we only need to replace the `next` and
+        # `previous` url values from the response returned by platform to construct a valid response object for the
+        # requester.
+        current_url = request.build_absolute_uri()
+        if page:
```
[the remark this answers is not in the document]
iloveagent57: I think you *have* to do something about `page` even when the caller doesn't include it in their query.

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -115,6 +129,66 @@ def unlink_users_from_enterprise(self, enterprise_customer_uuid, user_emails, is
...
+            - ``traverse_pagination`` (bool, optional): Indicates that the lms client should traverse and fetch all
+            pages.
+            of data. Cannot be supplied if ``page`` is supplied.
+            - ``page`` (int, optional): Which page of paginated data to return. Cannot be supplied if
+            ``traverse_pagination`` is supplied.
+        """
+        if traverse_pagination and page:
+            raise FetchGroupMembersConflictingParamsException(
+                'Params `traverse_pagination` and `page` are in conflict, only supply one or the other'
+            )
```
iloveagent57, line 162: Hmm, maybe you want to copy this type of validation into the request serializer?

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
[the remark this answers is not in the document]
iloveagent57: **or**, do the thing where the serializer enforces that `page` is required unless the caller wants to traverse all pages.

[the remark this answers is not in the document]
alex-sheehan-edx: I agree- we always need to do page work unless the requester supplied true for `traverse_pagination`

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -712,3 +712,114 @@ class SubsidyAccessPolicyAllocationResponseSerializer(serializers.Serializer):
...
+    def validate(self, attrs):
+        """
+        Raises a ValidationError both `traverse_pagination` and `page` are both supplied as they enable conflicting
+        behaviors.
+        """
+        if attrs.get('page') and attrs.get('traverse_pagination'):
+            raise serializers.ValidationError(
+                "Can only support one param of the following at a time: `page` or `traverse_pagination`"
+            )
+        return attrs
```
iloveagent57, line 825: I think you want the converse, too, right? At least one of them is required?

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -851,3 +857,152 @@ def allocate(self, request, *args, **kwargs):
...
+            request.accepted_media_type = GroupMembersWithAggregatesCsvRenderer().media_type
+            return Response(list(member_results), status=status.HTTP_200_OK, content_type='text/csv')
+
+        # Since we are essentially forwarding all request params to platform, we only need to replace the `next` and
+        # `previous` url values from the response returned by platform to construct a valid response object for the
+        # requester.
+        current_url = request.build_absolute_uri()
+        if not traverse_pagination:
+            if member_response.get('next'):
+                member_response['next'] = current_url.replace(f"page={page}", f"page={str(int(page) + 1)}")
```
iloveagent57, line 1002: nit: it's possible that the caller provided neither `page` nor `traverse_pagination` in their request. In which case this `replace()` won't do anything.

iloveagent57 (approved):
One more nit about `page`, otherwise looks g2g!

enterprise_access/apps/api/serializers/subsidy_access_policy.py
alex-sheehan-edx: yeah I guess that's true haha! Ok, 
```
if bool(attrs.get('page')) == bool(attrs.get('traverse_pagination')):
```
it is!

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
alex-sheehan-edx: good point, with the https://github.com/openedx/enterprise-access/pull/446#discussion_r1583793562 change we should be covered though, correct?

PR #448, opened 2024-04-25
docs: proposed ADR for a RedemptionAttempt model

iloveagent57:
Proposes the introduction of a new Django model, called ``RedemptionAttempt`` (or similar), to persist
business logic outcomes and workflow results in the course of the redemption workflow.

### DIFF
--- a/docs/decisions/0020-redemption-attempt-record.rst
+++ b/docs/decisions/0020-redemption-attempt-record.rst
@@ -1,0 +1,63 @@
+0020 Redemption Attempt Record
+******************************
+
+Status
+======
+
+Proposed - April 2024
+
+Context
+=======
+The primary responsibilities of the ``subsidy_access_policy`` module are as follows:
+
+1. Act as a "read-orchestrator", determining which learner-credit-based redemptions
+   are allowed under various conditions.
+2. Act as the entry-point for the redemption workflow, via the ``redeem`` view
+   and related ``SubsidyAccessPolicy.redeem()`` method. This responsibility depends on
+   the read-orchestrator responsibility above - the redemption flow must determine if redemption
+   is allowed before proceeding to request that the enterprise-subsidy service write
+   any ``transaction`` records.
+
+There is a general observability problem with the redemption workflow: outcomes that
+result in errors or non-redemption can only be inspected in a meaningful way from
+logs, or sometimes from the error details of the redemption HTTP response. This makes
+it difficult to debug the redemption workflow and gather metrics for engineering and
+non-engineering stakeholders alike.
+
+Decision
+========
+We'll introduce a new Django model, called ``RedemptionAttempt`` (or similar), to persist
+business logic outcomes and workflow results in the course of the redemption workflow. We should
+think of it as an audit-log type of persisted data - the presence or absence of a ``RedemptionAttempt``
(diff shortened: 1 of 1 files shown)


docs/decisions/0020-redemption-attempt-record.rst
```diff
@@ -0,0 +1,60 @@
...
+Status
+======
+
+Proposed - April 2024
+
+Context
+=======
+The primary responsibilities of the ``subsidy_access_policy`` module are as follows:
+
+1. Act as a "read-orchtestrator", determining which learner-credit-based redemptions
```
adamstankiewicz, line 13: nit: typo

```suggestion
1. Act as a "read-orchestrator", determining which learner-credit-based redemptions
```

docs/decisions/0020-redemption-attempt-record.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Decision
+========
+We'll introduce a new Django model, called ``RedemptionAttempt`` (or similar), to persist
+business logic outcomes and workflow results in the course of the redemption workflow. We should
+think of it as an audit-log type of persisted data - the presence or absence of a ``RedemptionAttempt``
+record should have no bearing on subsequent attempts of the redemption workflow.
+It should record the following types of data:
+
+1. The core inputs of the redeem workflow, including identifiers for the policy, subsidy,
```
adamstankiewicz, line 35: [curious] Would the `RedemptionAttempt` also optionally include any related request records (for when we support BnR with Learner Credit)? Probably not necessary to mention in this ADR since we haven't started BnR with Learner Credit in earnest yet, though.

docs/decisions/0020-redemption-attempt-record.rst
```diff
@@ -0,0 +1,60 @@
...
+   catalog, learner, content, and optionally any related ``LearnerContentAssignment`` record.
+2. The state of the policy at the time of the redemption attempt - pointing at a historical
```
adamstankiewicz, line 37: Agreed, pointing at a historical `SubsidyAccessPolicy` makes sense versus needing to add specific model fields representing the state of the policy at the time.

docs/decisions/0020-redemption-attempt-record.rst
```diff
@@ -0,0 +1,60 @@
...
+   ``SubsidyAccessPolicy`` record would be useful here.
+3. The value ``can_redeem()`` returned at the start of the flow, and if that value was ``False``,
+   the associated reason for disallowed redemption.
+4. The success state of the response from the enterprise-subsidy ``redeem`` request, including
+   identifiers for the transaction, fulfillment(s), and the price.
+5. The optional error state of the enterprise-subsidy response, including any error messages and codes.
+
+Consequences
```
adamstankiewicz, line 45: [question] Will these `RedemptionAttempt` records be exposed via any API(s) for consumption outside of the enterprise-access system as part of the initial proposal? Probably OK if not, until we have a use case that necessitates it.

adamstankiewicz (commented):
This proposal seems reasonable, and definitely should help mitigate some of the observability issues you've outlined that we have today.

Just had a couple clarifying questions 😄

adamstankiewicz (approved):

PR #450, opened 2024-04-25
docs: WIP ADR about transaction aggregates

iloveagent57:
Still very open to feedback, but here's a proposed approach to help us de-couple runtime dependencies in the ``can_redeem`` and ``redeem`` flows, and in the domain of groups.


docs/decisions/0021-transaction-aggregates.rst
```diff
@@ -0,0 +1,135 @@
...
+Inter-service Communication
+---------------------------
+There are several options for how transaction/reversal data can make its way to the aggregates
+layer in the enterprise-access service:
+
+1. Via management commands that reads transaction and reversal records via the enterprise-subsidy
+   REST API.  This approach may work for backpopulation, but is generally disfavorable because
+   it won't help us keep the aggregate records up-to-date in a timely manner (we'd have to run
+   it on some cron/schedule).
+2. Updating the aggregates based on the ``redeem`` response body (on successful redemption).
```
macdiesel, line 66: nit: If we will not use option 2 or 3 should they be moved to the "Alternatives Considered" block?

docs/decisions/0021-transaction-aggregates.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Additionally, we may eventually want aggregates at the **(policy, learner, group)** level, to support
+business record configurations where multiple groups relate to a single policy.
+
+Once these aggregates are in place, replace requests to the enterpirse-subsidy service
+in the ``can_redeem`` and ``redeem`` flows with database queries against these aggregates.
+
+Maintaining these aggregates depends on sourcing transaction and reversal data as follows:
+
+1. Starting from a zero-state in which Aggregate models exist but are not yet populated (backpopulation).
```
macdiesel, line 53: I think we should also call out recalculation here too.

docs/decisions/0021-transaction-aggregates.rst
```diff
@@ -0,0 +1,135 @@
...
+3. Emiting transaction/reversal commit events from the enterprise-subsidy service. If we can
+   emit these to an event broker, consumable by enterprise-access, then we can update aggregates
+   as transactions and reversals are committed. Furthermore, we could utilize a Django management
+   command to emit events that already exist (perhaps with ``modified`` times
+   earlier than some adjustable value) as a way to replay history. This would allow us to backpopulate
+   aggregate records.
+
+We'll utilize option (3) to both backpopulate and update aggregate records in real time.
+
+Ensuring Data Consistency
```
macdiesel, line 79: I'm operating under the assumption that aggregations are calculated initially from a zero state and are thereafter cumulative. 

If this is the case I think we will need some type of job to monitor the aggregation and ensure they are correct. Loops back to my earlier comment that we will need a way to recalculate as well.

docs/decisions/0021-transaction-aggregates.rst
```diff
@@ -0,0 +1,135 @@
...
+
+If we replace the synchronous querying of the enterprise-subsidy service for aggregates, with
+database reads of aggregate tables populated by asynchronous events, we've introduced asynchronicity
+into our overall flow.  This introduces the risk of running ``can_redeem`` flows with transaction
+aggregates that are stale, i.e. individual transaction records exist that have not been
+taken into accounts in the aggregate records. Without mitigation, this risks that our overall
+``redeem`` flow becomes less consistent, and timely consistency is one of the key cross-functional
+requirements of learner credit redemption.
+
+One possible mitigation is to record the timestamp of the most recent transaction or reversal
```
macdiesel, line 104: +1 solid idea.

Thinking out loud I wonder what other ways we could go about check summing this. This approach would ensure we are pointing at the same index but it doesn't ensure we've read all the records... I'm not sure, maybe the offset is enough.

macdiesel (approved):
Seems reasonable. I left some thoughts. I do think that there could be a performance win here, might be worth calling that out more prominently.

docs/decisions/0021-transaction-aggregates.rst
iloveagent57: I view "Alternatives considered" as a place to document different approaches to the general problem.  This is more like the rejected alternatives to a sub-problem: having accepted that we'll maintain aggregates in enterprise-access, how will the data get there?

docs/decisions/0021-transaction-aggregates.rst
iloveagent57: Good call, added a note about this, specifically for use in error correction.

docs/decisions/0021-transaction-aggregates.rst
iloveagent57: It's a good idea, added it as another possible mitigation.

docs/decisions/0021-transaction-aggregates.rst
iloveagent57: Yeah, it's a good point. I think there's one more thing we have to take into account, which is the absence of a needed aggregate for some `redeem` flow - if you query for aggregates on a given (policy, learner), and no record is in the DB, do we assume the metric is 0, or that we haven't yet computed the aggregate? I think we can hook into the `can-redeem` flow, plus maybe formalize a `zero aggregate` concept to mitigate this. I'll add a paragraph to the doc.

docs/decisions/0021-transaction-aggregates.rst
macdiesel: No strong feelings.

PR #452, opened 2024-04-29
feat: Updates groups learner record check to cached learner record

brobro10000:
**Description:**
Updates usage of `get_enterprise_user` in `includes_learner` for groups, to include an additional argument `enterprise_user_record` which defaults to `None` if nothing is passed. This allows us to pass the learner record in if it exists. If no `enterprise_user_record` exists, it will attempt to retrieve it from the cache, or fetch and initialize the cache via the `get_and_cache_enterprise_learner_record` function.
Modifies `includes_user` to `enterprise_user_record` to return to user record as oppose to resolve its boolean value whether the enterprise_user_record exists. This was modified to reduce the amount of calls to `get_enterprise_user` and the cache by allowing the check of its existence to happen in line, and using the `enterprise_user_record` object if it does exist within `includes_learner`

**Jira:**
[ENT-8603](https://2u-internal.atlassian.net/browse/ENT-8603)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/tests/mixins.py
+++ b/enterprise_access/apps/subsidy_access_policy/tests/mixins.py
@@ -5,6 +5,8 @@ from unittest.mock import patch
 
 from django.core.cache import cache as django_cache
 
+from test_utils import TEST_USER_RECORD
+
 from ..models import SubsidyAccessPolicy
 
 
@@ -49,6 +51,7 @@ class MockPolicyDependenciesMixin:
         )
 
         self.mock_enterprise_user_record = enterprise_user_record_patcher.start()
+        self.mock_enterprise_user_record.return_value = TEST_USER_RECORD
 
         self.addCleanup(subsidy_client_patcher.stop)
         self.addCleanup(transactions_cache_for_learner_patcher.stop)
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/subsidy_access_policy/tests/mixins.py
```diff
@@ -44,17 +44,16 @@ def setUp(self):
...
 
-        includes_user_patcher = patch.object(
-            SubsidyAccessPolicy, 'includes_user'
+        enterprise_user_record_patcher = patch.object(
+            SubsidyAccessPolicy, 'enterprise_user_record'
         )
 
-        self.mock_includes_user = includes_user_patcher.start()
-        self.mock_includes_user.return_value = True
+        self.mock_enterprise_user_record = enterprise_user_record_patcher.start()
```
iloveagent57, line 51: nit: it'd be good to set a reasonable default return value, like `self.mock_enterprise_user_record.return_value = {}`

iloveagent57 (approved):

adamstankiewicz (approved):

PR #454, opened 2024-04-29
fix: missing content metadata no longer crashes nudge email script

pwnage101:
This is a fix for a regression I recently introduced. It's actually a supported code path for this particular dict key to point to a Null value if no content metadata was found.

ENT-8789

### DIFF
--- a/enterprise_access/apps/content_assignments/management/commands/tests/test_automatically_nudge_assignments.py
+++ b/enterprise_access/apps/content_assignments/management/commands/tests/test_automatically_nudge_assignments.py
@@ -612,82 +612,54 @@ class TestAutomaticallyNudgeAssignmentCommand2(TestCase):
 
     @ddt.data(
         # Happy case: course starts in 14 days, and everything about the course and assignment are nudgeable.
+        {},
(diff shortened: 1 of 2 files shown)


enterprise_access/apps/content_assignments/management/commands/tests/test_automatically_nudge_assignments.py
```diff
@@ -612,82 +612,54 @@ def setUp(self):
...
+        {},
+        # Not nudgeable due to course metadata being missing from enterprise-catalog.
         {
-            'course_starts_in_x_days': 14,
-            'course_type': 'executive-education-2u',
-            'assignment_preferred_course_run_key': COURSE_RUN_KEY,
-            'assignment_state': LearnerContentAssignmentStateChoices.ACCEPTED,
-            'expected_task_call_days_before_course_start_date': 14,
+            'missing_content_metadata': True,
+            'expected_task_call_days_before_course_start_date': None,
```
pwnage101, line 619: Here's the added test case, sorry it's a needle in a haystack.

macdiesel (approved):

PR #462, opened 2024-05-06
feat: allowing for sort by enrollment count on group members endpoint

alex-sheehan-edx:
**Description:**
In order to sort by enrollment counts, the group data with aggregates endpoint needs to:
- fetch all pages of data from platform
- sort the data based on fetched subsidy data
- manually inject pagination + `next`/`previous` values
 
**Jira:**
https://2u-internal.atlassian.net/browse/ENT-8865

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


iloveagent57 (commented):
Is it possible to have the edx-enterprise endpoint support sorting at that layer? Then you can just pass a query parameter along from here into that endpoint and you won't end up force-traversing pagination without the client understanding that's what'll happen.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -985,6 +1001,26 @@ def get_group_member_data_with_aggregates(self, request, uuid):
...
+        # so we can properly construct the `next` and `previous` pages
+        if sort_by_enrollment_count:
+            member_results.sort(key=lambda result: result.get('enrollment_count'), reverse=(not is_reversed))
+            # Only if the request asked for a specific page of data do we need to restore page and traverse_pagination
+            if request_page:
+                # Needed to construct `next` and `previous` values for the response
+                num_member_results = len(member_results)
+
+                # Cut down the returned "all data" to the page and size requested
+                page_index_start = (request_page - 1) * 10
```
marlonkeating, line 1015: Will page size always be 10?  If so, should live in a constant.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -998,10 +1034,19 @@ def get_group_member_data_with_aggregates(self, request, uuid):
         # requester.
         current_url = request.build_absolute_uri()
         if not traverse_pagination:
-            if member_response.get('next'):
-                member_response['next'] = current_url.replace(f"page={page}", f"page={str(int(page) + 1)}")
-            if member_response.get('previous'):
-                member_response['previous'] = current_url.replace(f"page={page}", f"page={str(int(page) - 1)}")
+            # If sorting by enrollment count and a provided page, we have to be more cleaver about constructing next
```
marlonkeating, line 1037:
```suggestion
            # If sorting by enrollment count and a provided page, we have to be more clever about constructing next
```

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -998,10 +1034,19 @@ def get_group_member_data_with_aggregates(self, request, uuid):
...
+            # and previous values
+            if sort_by_enrollment_count:
+                if page_index_start > 0:
+                    member_response['previous'] = current_url.replace(f"page={page}", f"page={str(int(page) - 1)}")
```
marlonkeating, line 1041: nit: Shouldn't need to cast to a string inside an f-string expression.  Also, could make this a helper function.
```suggestion
                    member_response['previous'] = current_url.replace(f"page={page}", f"page={int(page) - 1}")
```

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -156,9 +179,9 @@ def fetch_group_members(
             - ``page`` (int, optional): Which page of paginated data to return. Cannot be supplied if
             ``traverse_pagination`` is supplied.
         """
-        if traverse_pagination and page:
+        if traverse_pagination == page:
```
marlonkeating, line 182: This check only works if page == 1.  Numbers above 1 aren't considered equal to True.

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -170,6 +193,22 @@ def fetch_group_members(
...
+                is_reversed,
+            )
+            cached_response = TieredCache.get_cached_response(cache_key)
+            if cached_response.is_found:
+                logger.info(
+                    f'all_enterprise_group_members cache hit for group_uuid {group_uuid}.'
+                )
+                return cached_response.value
+
+            params['page_size'] = 500
```
marlonkeating, line 211: Why are we setting page_size to 500?

marlonkeating (changes requested):

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
alex-sheehan-edx: oops hahaha

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
alex-sheehan-edx: for the sake of the datatable, yes

enterprise_access/apps/api_client/lms_client.py
alex-sheehan-edx: When traversing all pages of group member data, whether we are gathering CSV data or sorting by enrollment count, we want to make as few requests to platform as possible. When testing, I've found that some of our backends support page sizes up to 100, and a few others- 500. Django will just cap at whatever limit it has set if you request over that so I figured just make it super large to ensure we make as few requests as possible.

marlonkeating (approved):

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -73,6 +75,60 @@
...
+    """Helper method to create the next page url"""
+    return current_url.replace(f"page={page}", f"page={int(page) + 1}")
+
+
+def group_members_with_aggregates_previous_page(current_url, page):
+    """Helper method to create the previous page url"""
+    return current_url.replace(f"page={page}", f"page={int(page) - 1}")
+
+
+def _format_group_members_with_aggregate_members_response(
```
iloveagent57, line 88: nit: maybe `def _pagination_params_for_group_aggregates(...)`

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -966,25 +1024,58 @@ def get_group_member_data_with_aggregates(self, request, uuid):
...
+        request_page = None
+        request_traverse_pagination = None
+        if sort_by_enrollment_count := (sort_by == SORT_BY_ENROLLMENT_COUNT):
+            # Mock out data sent to platform to fetch the data we need to sort
+            # ie set `traverse_pagination` to True
+            sort_by = None
+            request_traverse_pagination = traverse_pagination
+            traverse_pagination = True
+            request_page = page
+            page = None
```
iloveagent57, line 1038: There should really be a warning, probably in the request serializer help text for the `sort_by` field, that setting `sort_by_enrollment_count` will cause the incoming request to wait for _all_ pages of data to be fetched from the edx-enterprise/LMS endpoint.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -966,25 +1024,58 @@ def get_group_member_data_with_aggregates(self, request, uuid):
...
+                page_index_start = (request_page - 1) * GROUP_MEMBERS_WITH_AGGREGATES_DEFAULT_PAGE_SIZE
+                member_results = member_results[
+                    page_index_start: page_index_start + GROUP_MEMBERS_WITH_AGGREGATES_DEFAULT_PAGE_SIZE
+                ]
+
+                # Restore the values
+                page = request_page
+                # No need to restore `traverse_pagination` if request_page is None because traverse must be true
+                # if no page is supplied
+                traverse_pagination = request_traverse_pagination
```
iloveagent57, line 1077: This is a code smell pointing at the fact that you're mutating query param values on the way in for the sake of making requests to the `fetch_group_members` endpoint. Instead of doing that, why don't you compute a different set of variables used only for that call? And even better would be to extract a helper method to both compute those variables and make the call?

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -966,25 +1024,58 @@ def get_group_member_data_with_aggregates(self, request, uuid):
             policy.uuid
         )
 
+        # If `sort_by_enrollment_count` is true, then we need to fetch all the group members records and do the sorting
+        # ourselves since platform is unaware of enrollment data numbers.
+        request_page = None
```
iloveagent57, line 1029: nit: a lot of the code below would read much cleaner if you renamed this as `page_requested_by_client` or something.

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -6,16 +6,39 @@
...
 
 from enterprise_access.apps.api_client.base_oauth import BaseOAuthClient
 from enterprise_access.apps.api_client.exceptions import FetchGroupMembersConflictingParamsException
+from enterprise_access.cache_utils import versioned_cache_key
 from enterprise_access.utils import should_send_email_to_pecu
 
 logger = logging.getLogger(__name__)
 
 
+def all_enterprise_group_members_cache_key(
```
iloveagent57, line 20: nit: not really "all" if user_query is not null.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
alex-sheehan-edx: as it's directly updating the values within member_response, I decided on `_update_pagination_params_for_group_aggregates`

enterprise_access/apps/api_client/lms_client.py
alex-sheehan-edx: "all_pages" I think fits a bit better then?

iloveagent57 (approved):
Much easier to understand and maintain now, great work.

PR #471, opened 2024-05-20
feat: allowing for group members to be specified when fetched

alex-sheehan-edx:
**Description:**
when fetching group member data with subsidy aggregate data, allowing for the specification of individual learners to be fetched

**Jira:**
https://2u-internal.atlassian.net/browse/ENT-8974

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


katrinan029 (approved):
🚢

PR #477, opened 2024-05-28
feat: validates spend limit against subsidy balance

brobro10000:
**Description:**
Updates the validation for a policies `spend_limit` where the `spend_limit` may not exceed the `total_deposits`.
The `total_deposits` field from the Subsidy model is the sum of all adjustments and the starting balance.
It will take into account all policies from the the parent `subsidy_uuid`.

The Django Admin screen will display a validation error at the top of the screen if this constraint is violated.
Screenshot 2024-06-04 at 11 53 33 AM

The validation that exists in the `clean()` function on the `SubsidyAccessPolicy` modal only runs if the expected updated `spend_limit` does not equal the current db value of the spend limit. This is done to allow modifications to other fields without running through the specific spend limit logic. This is also done as a precautionary measure for any policies that may already exist that violate this new constraint.

Updates the serializer validation helper function `policy_pre_write_validation` to run the `clean()` method to also check for this constraint. This is the [recommended method](https://twou.slack.com/archives/D04A0TDBBDL/p1717102085835939) for running clean on a validation serializer. 

Example: A (subsidy/plan) with 2 (policies/budgets) associated. When attempting to update the spend limit, it will take into account the values associated with all (policies/budgets) associated with the (subsidy/plan).

**Jira:**
[ENT-9004](https://2u-internal.atlassian.net/browse/ENT-9004)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/serializers/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/serializers/subsidy_access_policy.py
@@ -37,6 +37,13 @@ def policy_pre_write_validation(policy_instance_or_class, field_values_by_name):
     If a constraint occurs, raises a `ValidationError`.
     """
     violations = []
+
+    if isinstance(policy_instance_or_class, SubsidyAccessPolicy):
+        policy_instance_or_class.clean()
+    else:
+        instance = policy_instance_or_class(**field_values_by_name)
+        instance.clean()
+
     constraints = policy_instance_or_class.FIELD_CONSTRAINTS
 
     for field_name, new_value in field_values_by_name.items():
--- a/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
@@ -10,6 +10,7 @@ from uuid import UUID, uuid4
 
 import ddt
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from requests.exceptions import HTTPError
 from rest_framework import status
 from rest_framework.reverse import reverse
@@ -95,6 +96,8 @@ class CRUDViewTestMixin:
             'expiration_datetime': self.tomorrow,
             'current_balance': 4,
             'is_active': True,
+            'starting_balance': 4,
+            'total_deposits': 4,
         }
         subsidy_client_patcher = patch.object(
             SubsidyAccessPolicy, 'subsidy_client'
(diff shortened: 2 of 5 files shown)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -298,6 +300,37 @@ def get_policy_class_by_type(cls, policy_type):
...
+            policy_uuid = getattr(policy, 'uuid')
+            if policy_uuid != self.uuid:
+                spend_available_usd_cents = getattr(policy, 'spend_limit')
+                policy_balances.append(spend_available_usd_cents)
+        policy_balances.append(self.spend_limit)
+        return sum(policy_balances)
+    
+    @property
+    def is_spend_limit_updated(self):
+        if SubsidyAccessPolicy.objects.filter(uuid=self.uuid).first():
```
brobro10000, line 330: This conditional is specifically added to make the feature testable without it affecting every existing test.

Without employing this conditional, test related to the `SubsidyAccessPolicy` would need to be updated.
(Previous CI run: https://github.com/openedx/enterprise-access/actions/runs/9368112719/job/25789314544)
The conditional has no effect on the initial intention of this ticket which is to validate on changes to `spend_limit`

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -298,6 +299,40 @@ def get_policy_class_by_type(cls, policy_type):
...
+    def get_all_policies_associated_to_subsidy(self):
+        """
+        Retrives all policies associated to instance's subsidy
+        """
+        enterprise_customer_uuid = str(self.enterprise_customer_uuid)
+        subsidy_uuid = str(self.subsidy_uuid)
+        return SubsidyAccessPolicy.objects.filter(
+            enterprise_customer_uuid=enterprise_customer_uuid,
+            subsidy_uuid=subsidy_uuid,
+        )
```
iloveagent57, line 312: I might actually just move the body of this method into the method below.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -298,6 +299,40 @@ def get_policy_class_by_type(cls, policy_type):
...
+        for policy in policies:
+            policy_uuid = getattr(policy, 'uuid', self.uuid)
+            if policy_uuid != self.uuid:
+                spend_limit_usd_cents = getattr(policy, 'spend_limit', 0)
+                policy_balances.append(spend_limit_usd_cents)
+        if self.spend_limit is None:
+            policy_balances.append(0)
+        else:
+            policy_balances.append(self.spend_limit)
+        return sum(policy_balances)
```
iloveagent57, line 328: 1. You can use Django ORM to [exclude](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#exclude) _this_ policy from the rest of the policy records you want to consider.
2. It might clarify this to get the policy balance of `self` first, and then get all the other ones.
So in total, you can probably get this all down to:
```python
policy_balances = []
policy_balances.append(self.spend_limit or 0)
sibling_policies = SubsidyAccessPolicy.objects.filter(
    subsidy_uuid=self.subsidy_uuid,
).exclude(uuid=self.uuid)
for sibling in sibling_policies:
    policy_balances.append(sibling.spend_limit or 0)
return sum(policy_balances)
```

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -403,6 +444,13 @@ def subsidy_balance(self):
         current_balance = self.subsidy_record().get('current_balance') or 0
         return int(current_balance)
 
+    def subsidy_total_deposits(self):
+        """
+        Returns total remaining balance for the associated subsidy ledger.
+        """
+        total_deposits = self.subsidy_record().get('total_deposits')
+        return int(total_deposits)
```
iloveagent57, line 452: nit: maybe `return int(total_deposits or 0)` in case `total_deposits` is null.

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -37,6 +37,13 @@ def policy_pre_write_validation(policy_instance_or_class, field_values_by_name):
     If a constraint occurs, raises a `ValidationError`.
     """
     violations = []
+
+    if isinstance(policy_instance_or_class, SubsidyAccessPolicy):
+        policy_instance_or_class.clean()
```
iloveagent57, line 42: why would this if condition ever evaluate to false?

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -648,6 +651,50 @@ def test_update_views(self, is_patch, request_payload):
...
+                'retired': True,
+                'catalog_uuid': str(uuid4()),
+                'subsidy_uuid': str(uuid4()),
+                'access_method': AccessMethods.ASSIGNED,
+                'spend_limit': 6,
+                'per_learner_spend_limit': 10000,
+            },
+        },
+    )
+    @ddt.unpack
```
iloveagent57, line 671: nit: you don't really need ddt if you only have one test case.

enterprise_access/apps/api/serializers/subsidy_access_policy.py
pwnage101: I think the intent of this check is to only perform validation if `policy_instance_or_class` is an instance rather than a class.

```
>>> isinstance(AssignedLearnerCreditAccessPolicy, SubsidyAccessPolicy)
False
```

iloveagent57 (approved):

PR #479, opened 2024-06-03
feat: add enterprise_groups to project apps

katrinan029:
**Description:**
We created a cronjob to trigger the management command `groups_reminder_emails` but the command is currently not being recognized. To resolve this, this PR adds `enterprise_access.apps.enterprise_groups` to the `PROJECT_APPS` in order to run the management command `groups_reminder_emails`.

Additionally, this PR adds the `get_catalog_metadata_count` method to retrieve just the catalog count and adds logging. Previously, we were using an existing method `catalog_content_metadata`; however that method requires the `content_keys` param. 

Screenshot 2024-06-03 at 10 57 18 AM

**Jira:**
ENT-9057
https://2u-internal.atlassian.net/browse/ENT-9057

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/enterprise_groups/management/commands/groups_reminder_emails.py
+++ b/enterprise_access/apps/enterprise_groups/management/commands/groups_reminder_emails.py
@@ -51,7 +51,7 @@ class Command(BaseCommand):
             catalog_uuid = policy_group_association.subsidy_access_policy.catalog_uuid
             catalog_count = enterprise_catalog_client.get_content_metadata_count(
                 catalog_uuid
-            )['count']
+            )
 
             for pending_enterprise_customer_user in pending_enterprise_customer_users:
                 pending_enterprise_customer_user["subsidy_expiration_datetime"] = (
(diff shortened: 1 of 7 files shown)


enterprise_access/apps/enterprise_groups/management/commands/groups_reminder_emails.py
```diff
@@ -49,7 +49,7 @@ def handle(self, *args, **options):
                 policy_group_association.subsidy_access_policy.subsidy_expiration_datetime
             )
             catalog_uuid = policy_group_association.subsidy_access_policy.catalog_uuid
-            catalog_count = enterprise_catalog_client.catalog_content_metadata(
+            catalog_count = enterprise_catalog_client.get_content_metadata_count(
                 catalog_uuid
             )['count']
```
marlonkeating, line 54: Does this work?  `get_content_metadata_count` looks to be returning a bare value, not an object.

marlonkeating (approved):

PR #482, opened 2024-06-10
feat: Send action required by time to braze campaign

brobro10000:
**Description:**
Sends the time in the format `%H:%M` to be parsed by the braze client to the users local time zone. 
It is only updating the task for the assignment reminders to run a stage test on the email campaign with the updated development email template.

**Jira:**
[ENT-8888](https://2u-internal.atlassian.net/browse/ENT-8888)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/tasks.py
+++ b/enterprise_access/apps/content_assignments/tasks.py
@@ -447,6 +447,7 @@ def send_reminder_email_for_pending_assignment(assignment_uuid):
         'course_partner',
         'course_card_image',
         'learner_portal_link',
+        'action_required_by',
         'action_required_by_timestamp'
     )
     campaign_uuid = settings.BRAZE_ASSIGNMENT_REMINDER_NOTIFICATION_CAMPAIGN
@@ -490,6 +491,7 @@ def send_email_for_new_assignment(new_assignment_uuid):
         'course_partner',
         'course_card_image',
         'learner_portal_link',
+        'action_required_by',
         'action_required_by_timestamp'
     )
     campaign_uuid = settings.BRAZE_ASSIGNMENT_NOTIFICATION_CAMPAIGN
--- a/enterprise_access/apps/content_assignments/tasks.py
+++ b/enterprise_access/apps/content_assignments/tasks.py
@@ -59,6 +59,7 @@ class BrazeCampaignSender:
         'course_card_image',
         'learner_portal_link',
         'action_required_by',
+        'action_required_by_time'
     }
 
     def __init__(self, assignment):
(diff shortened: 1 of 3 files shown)


enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -213,6 +214,16 @@ def get_action_required_by(self):
             return None
         return format_datetime_obj(action_required_by['date'])
 
+    def get_action_required_by_time(self):
```
iloveagent57, line 217: Could we instead call this `action_required_by_datetime` and send an entire datetime (e.g. `Jan 01, 2021 12:00`)? That way, we won't need to glue together two fields downstream in Braze (and could eventually get rid of the old `action_required_by` field).

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -372,12 +386,10 @@ def send_exec_ed_enrollment_warmer(assignment_uuid, days_before_course_start_dat
         'contact_admin_link',
         'organization',
         'course_title',
-        'enrollment_deadline',
         'start_date',
         'course_partner',
         'course_card_image',
         'learner_portal_link',
-        'action_required_by',
```
brobro10000, line 380: These 2 attributes are not being used by this email campaign

enterprise_access/apps/content_assignments/tasks.py
brobro10000: I went ahead and made the change. I kept the `get_action_required_by` function because there are a couple of tests that uses the `get_action_required_by` that would require a refactor, but no other task currently utilize the `action_required_by` field.

I opted to go ahead and make the full change in this PR for both new and pending assignments with the intention of testing both in the same PR. Based on the testing i've been doing locally and in Braze, I am fairly confident its a quick switch from the `action_required_by` to `action_required_by_timestamp`.

From what I've noticed, if we are localizing the time stamp, Braze still requires a timestamp format directly in the campaign template. Testing it with the timestamp `%-I:%M %p %b %d, %Y` (ex. 12:00 PM Jan 01, 2021), Braze able to parse and localize the timestamp 👍🏽

enterprise_access/apps/content_assignments/tests/test_tasks.py
```diff
@@ -334,6 +334,7 @@ def test_send_reminder_email_for_pending_assignment(
             'uuid': self.policy.subsidy_uuid,
             'expiration_datetime': (now() + timedelta(days=1)).strftime('%Y-%m-%d %H:%M:%SZ'),
         }
+        print(mock_subsidy)
```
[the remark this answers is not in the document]
brobro10000: 😅 woops removed 👍🏽

enterprise_access/apps/content_assignments/tasks.py
brobro10000: UPDATE: Re-added `action_required_by` field to be remove once campaigns in prod are updated to `action_required_by_timestamp` field instead of `action_required_by`

iloveagent57 (approved):

PR #485, opened 2024-06-11
fix: datetime error in groups email reminder task

katrinan029:
**Description:**
Fix datetime object and conver to string for email task.

**Jira:**
ENT-9090
https://2u-internal.atlassian.net/browse/ENT-9090

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


alex-sheehan-edx (approved):
so long as the CI passes, lgtm! ✅

PR #494, opened 2024-06-18
fix: filters for only active policies for spend_limit check

brobro10000:
Updates the `spend_limit` constraint check to only check on policies that are active and to determine the sum of the `spend_limit`'s on only active policies.
Also updates how we sum sibling policies in the `total_spend_limit_for_all_policies_associated_to_subsidy` function by using the ORM to determine the sum of siblings and adding together the `spend_limit` in memory. 

**Description:**

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
@@ -610,7 +610,7 @@ class TestAuthenticatedPolicyCRUDViews(CRUDViewTestMixin, APITestWithMocks):
             enterprise_customer_uuid=self.enterprise_uuid,
             display_name='old display_name',
             spend_limit=5,
-            active=False,
+            active=True,
         )
 
         action = self.client.patch if is_patch else self.client.put
@@ -680,14 +680,13 @@ class TestAuthenticatedPolicyCRUDViews(CRUDViewTestMixin, APITestWithMocks):
             enterprise_customer_uuid=self.enterprise_uuid,
             display_name='old display_name',
             spend_limit=5,
-            active=False,
+            active=True,
         )
 
         request_payload = {
             'description': 'the new description',
             'display_name': 'new display_name',
             'active': True,
-            'retired': True,
             'catalog_uuid': str(uuid4()),
             'subsidy_uuid': str(uuid4()),
             'access_method': AccessMethods.ASSIGNED,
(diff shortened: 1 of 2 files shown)


enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -702,6 +701,43 @@ def test_update_views_with_exceeding_spend_limit(self):
...
+            display_name='old display_name',
+            spend_limit=5,
+            active=False,
+        )
+
+        request_payload = {
+            'description': 'the new description',
+            'display_name': 'new display_name',
+            'catalog_uuid': str(uuid4()),
+            'active': True,
```
pwnage101, line 726:
```suggestion
            'active': False,
```

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -318,10 +318,12 @@ def total_spend_limit_for_all_policies_associated_to_subsidy(self):
         Sums the policies spend_limit excluding the db's instance of this policy
         """
         policy_balances = []
-        policy_balances.append(self.spend_limit or 0)
+        if self.active:
+            policy_balances.append(self.spend_limit or 0)
```
brobro10000, line 322: Nested this logic in the conditional to check to see if the policy we are modifying is not active, we do not want to include it in the policy balances. This is in parity with the changes to the filtering that only looks for active sibling policies.

pwnage101 (approved):
Non-blocking idea:

FWIW, there's probably a way in pure ORM to this entire function all in one line and one SQL query without having to exclude self.uuid or loop over the results, e.g.:

```python
return SubsidyAccessPolicy.objects.filter(
    subsidy_uuid=self.subsidy_uuid,
    active=True,
).aggregate(Sum("spend_limit", default=0))["spend_limit__sum"]
```

Theoretically when this encounters a NULL spend_limit, The `default=0` takes over. Just an Idea that could make this function less complex and easier to understand/debug.

pwnage101:
Based on the standup discussion, it sounds like you want your clean function to say this instead:

```python
if self.is_spend_limit_updated and not self.active:
```

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -350,7 +358,7 @@ def clean(self):
         """
         Used to help validate field values before saving this model instance.
         """
-        if self.is_spend_limit_updated:
+        if self.active and (self.is_active_updated or self.is_spend_limit_updated):
```
brobro10000, line 361: Modified the conditional to avoid the path of a user disabling the policy. Setting an out of bounds spend_limit, then re-enabling the policy.
Before conditional added:
https://github.com/openedx/enterprise-access/assets/82611798/16a80f8d-de14-43ac-8ba3-efd71f9e5751


https://github.com/openedx/enterprise-access/assets/82611798/16a80f8d-de14-43ac-8ba3-efd71f9e5751

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -317,16 +317,13 @@ def total_spend_limit_for_all_policies_associated_to_subsidy(self):
         """
         Sums the policies spend_limit excluding the db's instance of this policy
         """
-        policy_balances = []
-        policy_balances.append(self.spend_limit or 0)
-        sibling_policies = SubsidyAccessPolicy.objects.filter(
+        spend_limit = self.spend_limit or 0
```
pwnage101, line 320: I know it's redundant to check self.active in this function because theoretically it's already covered by clean(), but it might be good to do it here too because there might be other future callers of this function besides clean().
```suggestion
        spend_limit = self.spend_limit if self.active and self.spend_limit or 0
```

pwnage101 (approved):

PR #495, opened 2024-06-20
fix: removing redundant group membership check

kiram15:
**Description:**
We're already doing the check of if the user is in the enterprise in a method that we call right after this check. 

**Jira:**
[ENT-9011](https://2u-internal.atlassian.net/browse/ENT-9011)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -532,17 +532,14 @@ class SubsidyAccessPolicy(TimeStampedModel):
             self.get_content_metadata(content_key),
         )
 
-    def includes_learner(self, lms_user_id, enterprise_user_record=None):
+    def includes_learner(self, lms_user_id):
         """
         Determine whether the lms user is associated properly with both the enterprise
         and the policy's group(s).
         """
-        if enterprise_user_record is not None:
-            learner_record = enterprise_user_record
-        else:
-            learner_record = self.enterprise_user_record(lms_user_id)
-            if not learner_record:
-                return False, REASON_LEARNER_NOT_IN_ENTERPRISE
+        learner_record = self.enterprise_user_record(lms_user_id)
+        if not learner_record:
+            return False, REASON_LEARNER_NOT_IN_ENTERPRISE
 
         associated_group_uuids = set(learner_record.get('enterprise_group', []))
         # if there are no policy groups, return early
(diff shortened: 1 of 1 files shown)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -689,9 +689,6 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
         # learner not associated to enterprise
         if not skip_customer_user_check:
             enterprise_user_record = self.enterprise_user_record(lms_user_id)
-            if not enterprise_user_record:
```
marlonkeating, line 692: Were there any unit tests around this condition?  If not, obviously we don't need to add them now, but it may point to testing gaps around `can_redeem` that should be addressed at some point.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -689,9 +689,6 @@ def can_redeem(self, lms_user_id, content_key, skip_customer_user_check=False):
...
-                self._log_redeemability(False, REASON_LEARNER_NOT_IN_ENTERPRISE, lms_user_id, content_key)
-                return (False, REASON_LEARNER_NOT_IN_ENTERPRISE, [])
```
adamstankiewicz, line 694: [suggestion] Given the only thing in this code path that relies on `enterprise_user_record` is getting removed, I might recommend removing `enterprise_user_record = self.enterprise_user_record(lms_user_id)` altogether, and only rely on the `self.enterprise_user_record` call within `includes_learner`.

I believe in this current state, `self.enterprise_user_record(lms_user_id)` is potentially queried twice unnecessarily (e.g., if `self.enterprise_user_record(lms_user_id)` return `None`, `self.enterprise_user_record(lms_user_id)` is executed again, with the same result, within `self.includes_learner`).

```python
included_in_policy, reason = self.includes_learner(lms_user_id)
```

```python
def includes_learner(self, lms_user_id, enterprise_user_record=None):
  """
  Determine whether the lms user is associated properly with both the enterprise
  and the policy's group(s).
  """
  learner_record = self.enterprise_user_record(lms_user_id)
  if not learner_record:
    return False, REASON_LEARNER_NOT_IN_ENTERPRISE

  # ...
```

marlonkeating (approved):

PR #502, opened 2024-06-26
feat: setup signal to consume transaction reveral events

iloveagent57:
**Description:**
* Adds a `reversed` assignment state and `reversed_at` field.
* Sets up a signal handler that will consume the LEDGER_TRANSACTION_REVERSED openedx event and modify a related assignment record's state to `reversed`.

**Jira:**
ENT-9009

**Testing**
1. Run migrations
2. Have https://github.com/openedx/enterprise-subsidy/pull/244 running (which implies that you have kafka etc running locally).
3. Ensure you have a reversible transaction related to some assignment.
4. In enterprise-access `make app-shell` and then run the consumer process with ```./manage.py consume_events -t enterprise-subsidy-ledger-transaction-lifecycle --group_id enterprise_access_dev```
5. Unenroll and refund the transaction from the enterprise-subsidy Transaction admin page.
6. Observe logs in the consumer process that the related *assignment* was reversed.
7. Observe in enterprise-access Django admin page for the assignment that the state has now changed to `reversed`.

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


pwnage101 (approved):
a new era!

PR #506, opened 2024-07-09
feat: backfill the `LearnerContentAssignment.allocated_at` field

iloveagent57:
In the past, we added and started programmatically populating an `allocated_at` field, but it has not yet been backfilled, and we need it to be backfilled in order to use it in features. This commit backfills that field. ENT-9038

An example historical record from local testing:
image

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/migrations/0021_backfill_allocated_at.py
+++ b/enterprise_access/apps/content_assignments/migrations/0021_backfill_allocated_at.py
@@ -1,0 +1,88 @@
+"""
+Data migration to backfill the ``LearnerContentAssignment.allocated_at`` field.
+"""
+
+from django.db import migrations
+from django.utils import timezone
+
+from enterprise_access.apps.content_assignments.constants import AssignmentActions
+
+
+BULK_OPERATION_BATCH_SIZE = 50
+
+
+def forwards_func(apps, schema_editor):
+    """
+    Populates the ``allocated_at`` field for all assignments whose value of that field is null.
+    """
+    LearnerContentAssignment = apps.get_model('content_assignments', 'LearnerContentAssignment')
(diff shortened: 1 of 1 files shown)


enterprise_access/apps/content_assignments/migrations/0021_backfill_allocated_at.py
```diff
@@ -0,0 +1,85 @@
...
+        if not last_notify_action:
+            allocation_time = assignment_record.created
+        else:
+            allocation_time = last_notify_action.completed_at
+
+        assignment_record.allocated_at = allocation_time
+        assignment_record.modified = timezone.now()
+        records_to_save.append(assignment_record)
+
+        # Note: The reason we need to manually create historical objects is that Django's bulk_create() built-in does not
```
adamstankiewicz, line 41: nit: should this comment reference `bulk_update` as opposed to `bulk_create` since `bulk_update` is used to update existing `LearnerContentAssignment` objects, not creating new records? Same principle applies, though.

enterprise_access/apps/content_assignments/migrations/0021_backfill_allocated_at.py
```diff
@@ -0,0 +1,85 @@
...
+        historical_records_to_save.append(historical_record)
+
+    LearnerContentAssignment.objects.bulk_update(
+        records_to_save,
+        ['allocated_at', 'modified'],
+        batch_size=50,
+    )
+    HistoricalLearnerContentAssignment.objects.bulk_create(
+        historical_records_to_save,
+        batch_size=50,
```
adamstankiewicz, line 64: nit: is it worth creating/using a local constant for `batch_size=50`?

adamstankiewicz (approved):
LGTM, with a couple tiny nits.

PR #508, opened 2024-07-10
fix: prefer allocated_at when determining assignment expiration

iloveagent57:
Helps address
* https://2u-internal.atlassian.net/browse/ENT-9190 (assignment expiration job looks at created instead of allocated_at for many assignments)

Note that we recently backfilled `allocated_at` via a data migration - there should be no assignment records with a null value for this field anymore.

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/models.py
+++ b/enterprise_access/apps/content_assignments/models.py
@@ -460,15 +460,7 @@ class LearnerContentAssignment(TimeStampedModel):
         sending a reminder notification for an assignment does not
         reset the auto-expiration date.
         """
-        last_notification = self.get_last_successful_notified_action()
-        # If we're currently sending the first notification message,
-        # we won't yet have a successful action, so use the created time of
-        # the assignment as the starting_point
-        if not last_notification:
-            starting_point = self.created
-        else:
-            starting_point = last_notification.completed_at
-        allocation_timeout_expiration = starting_point + timezone.timedelta(days=NUM_DAYS_BEFORE_AUTO_EXPIRATION)
+        allocation_timeout_expiration = self.allocated_at + timezone.timedelta(days=NUM_DAYS_BEFORE_AUTO_EXPIRATION)
         allocation_timeout_expiration = allocation_timeout_expiration.replace(tzinfo=UTC)
         return allocation_timeout_expiration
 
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/content_assignments/models.py
```diff
@@ -461,10 +461,10 @@ def get_allocation_timeout_expiration(self):
         """
         last_notification = self.get_last_successful_notified_action()
         # If we're currently sending the first notification message,
-        # we won't yet have a successful action, so use the created time of
+        # we won't yet have a successful action, so use the allocated_at or created time of
         # the assignment as the starting_point
         if not last_notification:
-            starting_point = self.created
+            starting_point = self.allocated_at or self.created
```
adamstankiewicz, line 467: [curious/question] In the current state, we were largely using the `last_notification` timestamp as a proxy for the most recent "allocation". Now that `allocated_at` is backfilled and the field is updated upon re-allocation, is there still benefit to relying primarily on the `last_notification` here or should this method's logic be migrated to primarily refer to `allocated_at` instead?

Upon assignment re-allocation, I believe this changed logic would still reference the `last_notification` timestamp despite `last_notification` only previously acting as a proxy for `allocated_at`,  where the notification of the assignment re-allocation could fail, and have the expiration date returned by this method continue to reference the original assignment allocation instead of the actual re-allocation timestamp (as determined by the updated `allocated_at`).

iloveagent57: Hmm, that's a great point.  Ok, so do we want
```python
# fully believe in ourselves
starting_point = self.allocated_at

# or be slightly more defensive
starting_point = self.allocated_at or self.created

# or be _too_ defensive
starting_point = self.allocated_at or self.get_the_latest_notification_time or self.created
```
Only reason I ask is that we don't have any constraint (yet) that _forces_ `allocated_at` to be non-null.

adamstankiewicz: My two cents would be for the "slightly more defensive" option:

```
# or be slightly more defensive
starting_point = self.allocated_at or self.created
```

Rationale: With the `allocated_at` backfill, the `allocated_at` field programmatically updated represents the same most recent successful notification so I don't think we necessarily need to defend against that particular timestamp here anymore.

Without the non-null constraint yet, falling back to `self.created` should account for any records where `allocated_at` are manually updated to be null in the interim.

adamstankiewicz: Ah, with https://github.com/openedx/enterprise-access/pull/509, then my two cents changes to the "fully believe in ourselves" option above haha.

adamstankiewicz (approved):
LGTM!

PR #510, opened 2024-07-10
fix: use allocated_at to determine recent_action time

iloveagent57:
This is an easy change since we previously backfilled `allocated_at` and set it to be NOT NULL via
* https://github.com/openedx/enterprise-access/pull/506
* https://github.com/openedx/enterprise-access/pull/509
ENT-9038

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


adamstankiewicz (approved):
:shipit:

PR #518, opened 2024-07-15
fix: we still need coalesce as a default when there is no last reminder

iloveagent57:
Manual testing revealed a bug in https://github.com/openedx/enterprise-access/pull/516
This PR fixes the bug.

**Jira:**
ENT-9245

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/models.py
+++ b/enterprise_access/apps/content_assignments/models.py
@@ -9,7 +9,8 @@ from uuid import UUID, uuid4
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db import models
 from django.db.models import Case, Exists, F, Max, OuterRef, Q, Value, When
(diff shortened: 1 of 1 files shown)


enterprise_access/apps/content_assignments/models.py
```diff
@@ -751,9 +755,9 @@ def annotate_dynamic_fields_onto_queryset(cls, queryset):
                 ),
                 When(
                     GreaterThan(F('most_recent_reminder'), F('allocated_at')),
-                    then=Value(AssignmentRecentActionTypes.ASSIGNED),
+                    then=Value(AssignmentRecentActionTypes.REMINDED),
```
adamstankiewicz, line 758: lol d'oh.

adamstankiewicz (approved):

PR #526, opened 2024-08-05
feat: publish content assignment list endpoint

sameeramin:
**Description:**
Published Content assignment list endpoint

**Jira:**
ENT-8983

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/api-compact.yaml
+++ b/api-compact.yaml
@@ -8,6 +8,7 @@
 #
 # Available service endpoints -- note that alternate endpoints may be presented at the API Gateway tier
 # POST  /api/v1/policy-allocation/{policy_uuid}/allocate/
+# GET   /api/v1/assignment-configurations/{assignment_configuration_uuid}/admin/assignments/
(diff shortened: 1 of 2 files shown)


api-compact.yaml
```diff
@@ -141,6 +142,136 @@ endpoints:
...
+            required: false
+          - in: query
+            name: content_key__in
+            type: array
+            description: "The globally unique content identifier to assign to the learner.
+              Joinable with ContentMetadata.content_key in enterprise-catalog."
+            required: false
+            items:
+              type: string
+              description: "content key"
```
sameenfatima78, line 171: what's the difference b/w these two parameters? Also, description is same.

sameenfatima78 (commented):
Left a comments, also wondering why do we have two params for the same attribute? e:g "learner_email" and "learner_email__in"

api-compact.yaml
sameeramin: There isn’t much difference between these two, first one takes a single value while the second one takes an array of `content_key` values

sameenfatima78 (approved):

PR #531, opened 2024-08-06
feat: add Exec Ed support for ForcedPolicyRedemption

iloveagent57:
Add some non-persisted admin model fields for `ForcedPolicyRedemption` so that we can pass along the required GEAG key/value pairs in our forced redeem call. ENT-9137

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/admin/__init__.py
+++ b/enterprise_access/apps/subsidy_access_policy/admin/__init__.py
@@ -6,6 +6,7 @@ from django.conf import settings
 from django.contrib import admin, messages
 from django.http import HttpResponseRedirect
 from django.urls import re_path, reverse
+from django.utils import timezone
 from django.utils.safestring import mark_safe
 from django.utils.text import Truncator  # for shortening a text
 from django.utils.translation import gettext_lazy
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/subsidy_access_policy/admin/__init__.py
```diff
@@ -436,7 +443,15 @@ def save_model(self, request, obj, form, change):
             return
 
         try:
-            obj.force_redeem()
+            extra_metadata = {
+                key: form.cleaned_data.get(key)
+                for key in FORCED_REDEMPTION_GEAG_KEYS
+            }
+            if extra_metadata:
```
pwnage101, line 450: Hm, won't this always create a truthy `extra_metadata` variable? If the form didn't have the geag values set, extra_metadata would still be a dict with 3 keys, which is truthy. I *think* what you want is this:
```suggestion
            extra_metadata = {
                key: form.cleaned_data.get(key)
                for key in FORCED_REDEMPTION_GEAG_KEYS
                if form.cleaned_data.get(key)
            }
            if extra_metadata:
```

pwnage101 (approved):

PR #538, opened 2024-08-14
feat: can_redeem checks for enrollment deadline

iloveagent57:
**Description:**
We want the `can_redeem` query to return false for a given (user, content key)
if the enrollment deadline for the content key is in the past. This is necessary because
some courses have runs with an advertised run for which the deadline is in the future,
but also have available recent runs where the upgrade deadline is in the past.

Depends on https://github.com/openedx/enterprise-catalog/pull/909 and https://github.com/openedx/enterprise-subsidy/pull/286 

A `can_redeem` false payload for a passed enroll by date will have a `reasons` value like this: 
```
"reasons": [
            {
                "reason": "beyond_enrollment_deadline",
                "user_message": "You can't enroll right now because the enrollment deadline for this course has passed.",
                "policy_uuids": [
                    "3cc0ad2c-ede1-476c-8271-97683855e41d"
                ],
                "metadata": {
                        {
                            "email": "enterprise_admin_test-enterprise@example.com",
                            "lms_user_id": 145
                        }
                    ]
                }
            }
        ]
```

**Jira:**
ENT-7472

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
@@ -37,6 +37,7 @@ from enterprise_access.apps.events.signals import SUBSIDY_REDEEMED
 from enterprise_access.apps.events.utils import send_subsidy_redemption_event_to_event_bus
 from enterprise_access.apps.subsidy_access_policy.constants import (
     GROUP_MEMBERS_WITH_AGGREGATES_DEFAULT_PAGE_SIZE,
+    REASON_BEYOND_ENROLLMENT_DEADLINE,
     REASON_CONTENT_NOT_IN_CATALOG,
     REASON_LEARNER_ASSIGNMENT_CANCELLED,
     REASON_LEARNER_ASSIGNMENT_FAILED,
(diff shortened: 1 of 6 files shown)


enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -768,7 +770,7 @@ def can_redeem(self, request, enterprise_customer_uuid):
                     enterprise_customer_uuid, lms_user_id, content_key
                 )
 
-            if not redemptions and not redeemable_policies:
+            if not successful_redemptions and not redeemable_policies:
```
iloveagent57, line 773: In local testing, I noticed I wasn't getting any `reasons` in my test responses even when I can an expected `can_redeem=False` result. I tend to use reversals a lot to set up test state, so I have many unsuccessful redemptions.

enterprise_access/apps/subsidy_access_policy/tests/test_models.py
```diff
@@ -483,6 +508,7 @@ def test_learner_enrollment_cap_policy_can_redeem(
             # Expected can_redeem result: False
             'policy_is_active': True,
             'catalog_contains_content': True,
+            'enroll_by_date': '2099-01-01T00:00:00Z',
```
adamstankiewicz, line 527: [nit/question/sanity check] I might just not be seeing it, but is there a test case permutation where `enroll_by_date` is missing?

iloveagent57: ah, nope, can add one

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
adamstankiewicz: This seems reasonable!

adamstankiewicz (approved):

PR #543, opened 2024-08-19
feat: allow staff users to override `lms_user_id` in `can-redeem` API

adamstankiewicz:
**Description:**

Supports an optional `lms_user_id` query parameter in the `can-redeem` API to override the default user id used for the request. **Only works if the authenticated request user `is_staff=True`**.

```
http://localhost:18270/api/v1/policy-redemption/enterprise-customer/{enterprise_customer_uuid}/can-redeem/?lms_user_id=1234
```

image

**Jira:**
[ENT-9291](https://2u-internal.atlassian.net/browse/ENT-9291)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
@@ -33,6 +33,7 @@ from enterprise_access.apps.subsidy_access_policy.constants import (
     REASON_LEARNER_ASSIGNMENT_CANCELLED,
     REASON_LEARNER_ASSIGNMENT_FAILED,
     REASON_LEARNER_NOT_ASSIGNED_CONTENT,
+    REASON_LEARNER_NOT_IN_ENTERPRISE_GROUP,
     REASON_NOT_ENOUGH_VALUE_IN_SUBSIDY,
     AccessMethods,
     MissingSubsidyAccessReasonUserMessages,
@@ -1108,7 +1109,7 @@ class TestPolicyRedemptionAuthNAndPermissionChecks(APITestWithMocks):
         response = self.client.get(reverse('api:v1:policy-redemption-credits-available'), query_params)
         self.assertEqual(response.status_code, expected_response_code)
 
-        # The can_redeem endpoint
+        # The can-redeem endpoint
         url = reverse(
             "api:v1:policy-redemption-can-redeem",
             kwargs={"enterprise_customer_uuid": self.enterprise_uuid},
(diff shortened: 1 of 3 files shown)


enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -478,7 +478,7 @@ def get_queryset(self):
             enterprise_customer_uuid=self.enterprise_customer_uuid,
         ).order_by('-created')
 
-    def evaluate_policies(self, enterprise_customer_uuid, lms_user_id, content_key):
+    def evaluate_policies(self, enterprise_customer_uuid, lms_user_id, content_key, skip_customer_user_check=False):
```
adamstankiewicz, line 481: [inform] Called with `skip_customer_user_check=True` when `request.user.is_staff=True` and `lms_user_id` query parameter is provided.

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -2231,6 +2238,92 @@ def test_can_redeem_subsidy_client_http_error(self, mock_get_client):
             'subsidy_status_code': str(status.HTTP_503_SERVICE_UNAVAILABLE),
         }
 
+    @ddt.data(
+        {'is_staff': True, 'lms_user_id_override': 1234},
+        {'is_staff': True, 'lms_user_id_override': None},
+        {'is_staff': False, 'lms_user_id_override': 5678},
+        {'is_staff': False, 'lms_user_id_override': None}
```
iloveagent57, line 2245: [totally optional] Sometimes for clarity, it's nice to put `expected_can_redeem` as a ddt parameter in here just to make it super obvious what the expected output is for a given set of input. Like
```python
{'is_staff': True, 'lms_user_id_override': 1234, 'expected_can_redeem': True},
{'is_staff': True, 'lms_user_id_override': None, 'expected_can_redeem': False},
```

iloveagent57 (approved):
Looks good, few minor nits.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
adamstankiewicz: Oh, yep... d'oh. I said it backwards above! Thanks for catching.

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -2231,6 +2238,92 @@ def test_can_redeem_subsidy_client_http_error(self, mock_get_client):
...
+    ):
+        """
+        Test that the can_redeem endpoint allows staff to override the LMS user ID.
+        """
+        self.user.lms_user_id = TEST_USER_RECORD['user']['id']
+        # Authenticate as a staff user
+        if is_staff:
+            self.user.is_staff = True
+        else:
+            self.user.is_staff = False
```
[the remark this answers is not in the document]
adamstankiewicz: Good catch; will update.

PR #547, opened 2024-08-27
feat: adds backoff to enterprise-catalog client calls

brobro10000:
Adds `backoff` decorator to the enterprise access api client functions as a means to retry calls that may face connection or timeout errors.

I opted for a conservative amount of retries (3) given the exponential nature of the `wait_gen` or backoff strategy and `contains_content_items` usage in the `can-redeem` function.

**Description:**

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/api_client/constants.py
```diff
@@ -0,0 +1,16 @@
...
+import backoff
+from requests.exceptions import ConnectionError as RequestsConnectionError
+from requests.exceptions import HTTPError
+from requests.exceptions import Timeout as RequestsTimeoutError
+
+# wait_gen references the field in the backoff decorator representing
+# the backoff strategy for retries
+wait_gen = backoff.expo
+
+exception = (
```
macdiesel, line 13: Should be plural. Perhaps backoff_exceptions or retry_exceptions?

enterprise_access/apps/api_client/constants.py
```diff
@@ -0,0 +1,16 @@
...
+Constants for API client
+"""
+import backoff
+from requests.exceptions import ConnectionError as RequestsConnectionError
+from requests.exceptions import HTTPError
+from requests.exceptions import Timeout as RequestsTimeoutError
+
+# wait_gen references the field in the backoff decorator representing
+# the backoff strategy for retries
+wait_gen = backoff.expo
```
macdiesel, line 11: What does wait_gen mean? Could we make this something more descriptive? Perhaps this already matches other implementations and I just just let it go?

brobro10000: I sourced the name from the [arguments](https://github.com/litl/backoff/blob/d82b23c42d7a7e2402903e71e7a7f03014a00076/backoff/_decorator.py#L123-L137) of `on_exception` but it makes sense that the naming should be more description. Updated 👍🏽

enterprise_access/apps/api_client/constants.py
brobro10000: Updated 👍🏽

adamstankiewicz (approved):
Seems reasonable to me, but I might also defer to a review from someone (e.g.,. Troy?) more familiar with best practices around implementation of this backoff strategy.

macdiesel (approved):

pwnage101 (approved):

PR #560, opened 2024-09-10
feat: update start_date logic to consider self paced and old start_dates

brobro10000:
**Description:**
Checks if the `start_date` is before the `START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS` value (currently 14) and sends that value to braze to be parsed for tasks related to learner credit assignment. The main intentions behind this addition is to handle OCM course start dates which start very far in the past. 

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/content_assignments/utils.py
```diff
@@ -0,0 +1,55 @@
...
+from dateutil import parser
+from pytz import UTC
+
+from enterprise_access.apps.content_assignments.constants import (
+    BRAZE_ACTION_REQUIRED_BY_TIMESTAMP_FORMAT,
+    START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS
+)
+
+
+def is_within_minimum_start_date_threshold(today, start_date):
```
macdiesel, line 15: nit: could we rename today to curr_date or now or something else? 

Why? Feels ambiguous? it could possibly not really be today anymore at execution time?

macdiesel (approved):
Nice PR Hamzah.

PR #561, opened 2024-09-10
feat: support course run based assignments in credits_available, expiration, emails, etc.

adamstankiewicz:
**Description:**

Relying on the update to enterprise-catalog's `get_content_metadata` REST API to return the top-level course object when passing course run key(s) to the `?content_keys` query parameter and the addition of `normalized_metadata_by_run`, this PR updates the `get_content_metadata_for_assignments` method and where it's consumed to support `credits_available`, expiration management commands, and email notifications related to assignments.

**Jira:**
[ENT-8876](https://2u-internal.atlassian.net/browse/ENT-8876)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api_client/enterprise_catalog_client.py
+++ b/enterprise_access/apps/api_client/enterprise_catalog_client.py
@@ -65,7 +65,7 @@ class EnterpriseCatalogApiClient(BaseOAuthClient):
             'traverse_pagination': traverse_pagination,
             **kwargs,
         }
-        endpoint = f'{self.enterprise_catalog_endpoint}{str(catalog_uuid)}/get_content_metadata/'
+        endpoint = f'{self.enterprise_catalog_endpoint}{catalog_uuid}/get_content_metadata/'
 
         response = self.client.get(endpoint, params=query_params)
         response.raise_for_status()
--- a/enterprise_access/apps/api_client/enterprise_catalog_client.py
+++ b/enterprise_access/apps/api_client/enterprise_catalog_client.py
@@ -65,7 +65,7 @@ class EnterpriseCatalogApiClient(BaseOAuthClient):
             'traverse_pagination': traverse_pagination,
             **kwargs,
         }
-        endpoint = self.enterprise_catalog_endpoint + str(catalog_uuid) + '/get_content_metadata/'
+        endpoint = f'{self.enterprise_catalog_endpoint}{str(catalog_uuid)}/get_content_metadata/'
 
         response = self.client.get(endpoint, params=query_params)
         response.raise_for_status()
--- a/enterprise_access/apps/content_assignments/api.py
+++ b/enterprise_access/apps/content_assignments/api.py
@@ -23,7 +23,12 @@ from enterprise_access.apps.content_assignments.tasks import (
 )
 from enterprise_access.apps.core.models import User
 from enterprise_access.apps.subsidy_access_policy.content_metadata_api import get_and_cache_content_metadata
-from enterprise_access.utils import chunks, get_automatic_expiration_date_and_reason, localized_utcnow
+from enterprise_access.utils import (
+    chunks,
+    get_automatic_expiration_date_and_reason,
+    get_normalized_metadata_for_assignment,
+    localized_utcnow
+)
 
 from .constants import AssignmentAutomaticExpiredReason, LearnerContentAssignmentStateChoices
 from .models import AssignmentConfiguration, LearnerContentAssignment
@@ -761,8 +766,13 @@ def nudge_assignments(assignments, assignment_configuration_uuid, days_before_co
             enterprise_catalog_uuid,
             [assignment],
         )
+        print('content_metadata_for_assignments?!?!', content_metadata_for_assignments)
         content_metadata = content_metadata_for_assignments.get(assignment.content_key, {})
-        start_date = content_metadata.get('normalized_metadata', {}).get('start_date')
+        print('content_metadata?!?!', content_metadata)
+        normalized_metadata = get_normalized_metadata_for_assignment(assignment, content_metadata)
+        print('normalized_metadata?!?!', normalized_metadata)
+
+        start_date = normalized_metadata.get('start_date')
         course_type = content_metadata.get('course_type')
 
         # check if the course_type is an executive-education course
--- a/enterprise_access/apps/content_assignments/api.py
+++ b/enterprise_access/apps/content_assignments/api.py
@@ -766,11 +766,8 @@ def nudge_assignments(assignments, assignment_configuration_uuid, days_before_co
             enterprise_catalog_uuid,
             [assignment],
         )
-        print('content_metadata_for_assignments?!?!', content_metadata_for_assignments)
         content_metadata = content_metadata_for_assignments.get(assignment.content_key, {})
-        print('content_metadata?!?!', content_metadata)
         normalized_metadata = get_normalized_metadata_for_assignment(assignment, content_metadata)
-        print('normalized_metadata?!?!', normalized_metadata)
 
         start_date = normalized_metadata.get('start_date')
         course_type = content_metadata.get('course_type')
--- a/enterprise_access/apps/content_assignments/content_metadata_api.py
+++ b/enterprise_access/apps/content_assignments/content_metadata_api.py
@@ -17,23 +17,6 @@ DATE_INPUT_PATTERNS = [
 DEFAULT_STRFTIME_PATTERN = '%b %d, %Y'
 
 
-def _process_course_metadata(course_metadata, is_run_based_by_content_key, metadata_by_key):
-    """
-    Helper function to update the metadata_by_key dictionary with the course metadata based on
-    either a course run key (for run-based assignments) or a course key.
-    """
-    course_key = course_metadata.get('key')
-
-    # If a course run is assigned, update the metadata for the course run key. Otherwise,
-    # update the metadata for the course key.
-    for assignment_content_key, is_assigned_course_run in is_run_based_by_content_key.items():
-        if is_assigned_course_run:
-            metadata_by_key[assignment_content_key] = course_metadata
-        else:
-            # if not is_assigned_course_run, we can assume it's a (legacy) course-based assignment
-            metadata_by_key[course_key] = course_metadata
-
-
 def get_content_metadata_for_assignments(enterprise_catalog_uuid, assignments):
     """
     Fetches (from cache or enterprise-catalog API call) content metadata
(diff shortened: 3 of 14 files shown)


enterprise_access/apps/content_assignments/content_metadata_api.py
```diff
@@ -28,11 +28,28 @@ def get_content_metadata_for_assignments(enterprise_catalog_uuid, assignments):
...
     """
-    content_keys = sorted({assignment.content_key for assignment in assignments})
-    content_metadata_list = get_and_cache_catalog_content_metadata(enterprise_catalog_uuid, content_keys)
-    metadata_by_key = {
-        record['key']: record for record in content_metadata_list
+    content_keys = {
+        (assignment.content_key, assignment.is_assigned_course_run)
+        for assignment in assignments
     }
+    content_metadata_list = get_and_cache_catalog_content_metadata(enterprise_catalog_uuid, content_keys)
```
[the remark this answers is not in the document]
iloveagent57: I think you might actually want something like:
```python
is_run_based_by_content_key = {
    assignment.content_key: assignment.is_assigned_course_run
    for assignment in assignments
}
content_metadata_list = get_and_cache_catalog_content_metadata(enterprise_catalog_uuid, is_run_based_by_content_key.keys())
```

enterprise_access/apps/content_assignments/content_metadata_api.py
```diff
@@ -28,11 +28,28 @@ def get_content_metadata_for_assignments(enterprise_catalog_uuid, assignments):
...
+            if is_assigned_course_run:
+                metadata_by_key.update({
+                    content_key: record
+                })
+                break  # Stop searching after the match is found
+            if record_key == content_key:
+                metadata_by_key.update({
+                    record_key: record
+                })
+                break  # Stop searching after the match is found
```
pwnage101, line 51: There's something weird going on here cause both if-blocks do the same thing so they could just be collapsed into one (like below), but that doesn't seem right because if the first row of `content_keys` corresponds to an assigned course run, then metadata_by_key would always only have 1 key in it.
```suggestion
            if is_assigned_course_run or record_key == content_key:
                metadata_by_key.update({
                    content_key: record
                })
                break  # Stop searching after the match is found
```

I haven't thought through what the solution would be, just flagging it for now.

enterprise_access/apps/content_assignments/content_metadata_api.py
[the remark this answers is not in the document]
adamstankiewicz: Good catch! It was technically still working due to sending up two query params for each tuple, `?content_keys={course run key}&content_keys=True`, which still returned the expected course metadata from the API 😄 

I've updated it to your suggestion in, which resolved the issue: https://github.com/openedx/enterprise-access/pull/561/commits/263dd2fd3e20c67977d5cb9c05361d94508e6467

enterprise_access/apps/content_assignments/content_metadata_api.py
adamstankiewicz: > There's something weird going on here cause both if-blocks do the same thing so they could just be collapsed into one (like below)

I don't believe the if-blocks were doing quite the same thing? The first one dealing with `if is_assigned_course_run` uses the assignment's `content_key` whereas the second one regarding `if record_key == content_key:` uses the top-level course key.

The above suggestion would both rely on the assignment's `content_key`, or the course run key in this case and then fail to lookup the content metadata based on the `assignment.content_key` to top-level courses.

I just pushed up a new commit https://github.com/openedx/enterprise-access/pull/561/commits/69b0ff2ddcc9945b023d09d8d6c426682dbba7ba that might improve the readability/understandability of the logic some?

enterprise_access/apps/content_assignments/content_metadata_api.py
```diff
@@ -17,22 +17,49 @@
...
+    course_key = course_metadata.get('key')
+
+    # If a course run is assigned, update the metadata for the course run key. Otherwise,
+    # update the metadata for the course key.
+    for assignment_content_key, is_assigned_course_run in is_run_based_by_content_key.items():
+        if is_assigned_course_run:
+            metadata_by_key[assignment_content_key] = course_metadata
+        else:
+            # if not is_assigned_course_run, we can assume it's a (legacy) course-based assignment
+            metadata_by_key[course_key] = course_metadata
```
pwnage101, line 34: Aren't these still the same logic in each block? If the assignment is a legacy assignment, then `course_key` is equivalent to `assignment_content_key`, is it not?
```suggestion
        metadata_by_key[assignment_content_key] = course_metadata
```

enterprise_access/apps/content_assignments/api.py
```diff
@@ -761,8 +766,13 @@ def nudge_assignments(assignments, assignment_configuration_uuid, days_before_co
             enterprise_catalog_uuid,
             [assignment],
         )
+        print('content_metadata_for_assignments?!?!', content_metadata_for_assignments)
         content_metadata = content_metadata_for_assignments.get(assignment.content_key, {})
-        start_date = content_metadata.get('normalized_metadata', {}).get('start_date')
+        print('content_metadata?!?!', content_metadata)
+        normalized_metadata = get_normalized_metadata_for_assignment(assignment, content_metadata)
+        print('normalized_metadata?!?!', normalized_metadata)
```
pwnage101, line 773: leftover print statements from testing?

enterprise_access/apps/content_assignments/content_metadata_api.py
```diff
@@ -17,22 +17,49 @@
...
+
 def get_content_metadata_for_assignments(enterprise_catalog_uuid, assignments):
     """
     Fetches (from cache or enterprise-catalog API call) content metadata
     in bulk for the `content_keys` of the given assignments, provided
     such metadata is related to the given `enterprise_catalog_uuid`.
 
+    Note that the `content_keys` of the provided assignments may be
+    either course run keys or course keys.
+
```
pwnage101, line 45: Nit: maybe also clarify that the metadata returned are all at the course-level, whether the key of the returned mapping is a run key or course key.

enterprise_access/apps/content_assignments/management/commands/automatically_nudge_assignments.py
```diff
@@ -155,23 +156,20 @@ def handle(self, *args, **options):
...
-                    can_send_nudge_notification_in_advance = is_date_n_days_from_now(
-                        target_datetime=datetime_start_date,
-                        num_days=days_before_course_start_date
+                    can_send_nudge_notification_in_advance = (
+                        is_date_n_days_from_now(
+                            target_datetime=datetime_start_date,
+                            num_days=days_before_course_start_date
+                        )
+                        if start_date is not None
+                        else False
```
pwnage101, line 172: 👍  Nit: I think datetime_start_date would be a better/more consistent value to use:
```suggestion
                    can_send_nudge_notification_in_advance = (
                        is_date_n_days_from_now(
                            target_datetime=datetime_start_date,
                            num_days=days_before_course_start_date
                        )
                        if datetime_start_date is not None
                        else False
```

pwnage101 (commented):
Just nits for the rest!

enterprise_access/apps/content_assignments/content_metadata_api.py
adamstankiewicz: @pwnage101 Ahhh, I understand now; you're absolutely correct. I pushed another commit https://github.com/openedx/enterprise-access/pull/561/commits/91ac3ab7e4032481622e1808ec7f77b64018ce7d with a refactor. Let me know what you think!

enterprise_access/apps/content_assignments/content_metadata_api.py
```diff
@@ -48,18 +31,9 @@ def get_content_metadata_for_assignments(enterprise_catalog_uuid, assignments):
...
-        enterprise_catalog_uuid,
-        is_run_based_by_content_key.keys()
-    )
-    metadata_by_key = {}
-    for course_metadata in content_metadata_list:
-        _process_course_metadata(course_metadata, is_run_based_by_content_key, metadata_by_key)
-
+    content_keys = {assignment.content_key for assignment in assignments}
+    content_metadata_list = get_and_cache_catalog_content_metadata(enterprise_catalog_uuid, content_keys)
+    metadata_by_key = dict(zip(content_keys, content_metadata_list))
```
adamstankiewicz, line 37: [inform] Hm, actually... still noodling as this approach with `zip` assumes the same ordering between `content_keys` and `content_metadata_list`. I was trying to avoid what would have otherwise been nested `for` loops.

enterprise_access/apps/content_assignments/content_metadata_api.py
adamstankiewicz: [inform] @pwnage101 I updated once more in https://github.com/openedx/enterprise-access/pull/561/commits/cb9d27bf0a90abbdabb0a3f21bfcb6b87475f3bd.

enterprise_access/apps/api_client/enterprise_catalog_client.py
```diff
@@ -65,7 +65,7 @@ def catalog_content_metadata(self, catalog_uuid, content_keys, traverse_paginati
             'traverse_pagination': traverse_pagination,
             **kwargs,
         }
-        endpoint = self.enterprise_catalog_endpoint + str(catalog_uuid) + '/get_content_metadata/'
+        endpoint = f'{self.enterprise_catalog_endpoint}{str(catalog_uuid)}/get_content_metadata/'
```
iloveagent57, line 68: nit: I think the call to `str()` is superflous now.

iloveagent57 (approved):
Looks great overall! Couple small things we could simplify.

enterprise_access/apps/content_assignments/content_metadata_api.py
```diff
@@ -17,21 +17,42 @@
...
         could be found for a given key.
     """
-    content_keys = sorted({assignment.content_key for assignment in assignments})
-    content_metadata_list = get_and_cache_catalog_content_metadata(enterprise_catalog_uuid, content_keys)
+    content_keys = {assignment.content_key for assignment in assignments}
+    course_metadata_list = get_and_cache_catalog_content_metadata(enterprise_catalog_uuid, content_keys)
     metadata_by_key = {
-        record['key']: record for record in content_metadata_list
+        assignment.content_key: _content_metadata_for_assignment(assignment, course_metadata_list)
+        for assignment in assignments
```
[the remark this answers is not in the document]
adamstankiewicz: ah, great catch. you are correct! will update before merge.

PR #562, opened 2024-09-13
feat: support run-based assignments in `can-redeem`; ensure assigned policies have highest priority

adamstankiewicz:
# Description

## Make run-based assignments discoverable in `can-redeem`

In `get_assignment_for_learner`, we perform a lookup of assignments based on a top-level course key, even for run-based assignments. Given this, a run-based assignment would never be found. To mitigate, the queryset `get` now considered the assignment's `parent_content_key` as well.

## Ensuring assignment-based policies have highest priority

When a learner can redeem with both a `PerLearnerEnrollmentCreditAccessPolicy` _and_ a `AssignedLearnerCreditAccessPolicy`, the `can-redeem` API response does not prioritize the assignment-based policy for redemption over other policy types.

The lack of priority for assignment-based policies was noticed when performing stage QA and observed that the learner portal Course page was suggesting the course was assigned (via a badge and the assignment-specific price messaging), but the redemption was attempting to go through the `PerLearnerEnrollmentCreditAccessPolicy` instead of the expected `AssignedLearnerCreditAccessPolicy`, not quite matching the UX.

**Jira:**
N/A

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/api.py
+++ b/enterprise_access/apps/content_assignments/api.py
@@ -216,9 +216,16 @@ def get_assignment_for_learner(
             f'No assignment found with content_key or parent_content_key {content_key} '
             f'for {assignment_configuration} and lms_user_id {lms_user_id}',
         )
+    except LearnerContentAssignment.MultipleObjectsReturned as exc:
+        logger.error(
+            f'Multiple assignments found with content_key or parent_content_key {content_key} '
+            f'for {assignment_configuration} and lms_user_id {lms_user_id}',
+        )
+        raise exc
 
     # If no exact match was found, try to normalize the content key and find a match. This happens when
-    # the content_key is a course run key and the assignment's content_key is a course key
+    # the content_key is a course run key and the assignment's content_key is a course key, as depicted
+    # by row 2 in the above docstring matrix.
     content_key_to_match = _normalize_course_key_from_metadata(assignment_configuration, content_key)
     if not content_key_to_match:
         logger.error(f'Unable to normalize content_key {content_key} for {assignment_configuration} and {lms_user_id}')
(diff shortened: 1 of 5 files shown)


brobro10000 (approved):

enterprise_access/apps/content_assignments/api.py
```diff
@@ -200,19 +201,40 @@ def get_assignment_for_learner(
...
+    queryset = LearnerContentAssignment.objects.select_related('assignment_configuration')
+
+    try:
+        # First, try to find a corresponding assignment based on the specific content key,
+        # considering both assignments' content key and parent content key.
+        return queryset.get(
+            Q(content_key=content_key) | Q(parent_content_key=content_key),
+            assignment_configuration=assignment_configuration,
+            lms_user_id=lms_user_id,
+        )
```
pwnage101, line 213: For completeness, should we handle-log-raise the "multiple records returned" exception too?  I'm imagining the only time we'd see this is if there's simultaneously a legacy and new assignment for the same course & customer, which shouldn't happen because our business logic should prevent it. That said, explicitly handling the exception might answer any questions from future code readers about why it's not handled.

enterprise_access/apps/content_assignments/api.py
```diff
@@ -200,19 +201,40 @@ def get_assignment_for_learner(
...
+    except LearnerContentAssignment.DoesNotExist:
+        logger.info(
+            f'No assignment found with content_key or parent_content_key {content_key} '
+            f'for {assignment_configuration} and lms_user_id {lms_user_id}',
+        )
+
+    # If no exact match was found, try to normalize the content key and find a match. This happens when
+    # the content_key is a course run key and the assignment's content_key is a course key
```
pwnage101, line 221: Nit: clarify everything after this line is in support of row 2 of the matrix.

pwnage101 (approved):
LGTM, just nits!

enterprise_access/apps/content_assignments/api.py
adamstankiewicz: @pwnage101 Updated to include an `except` block as follows:

```python
except LearnerContentAssignment.MultipleObjectsReturned as exc:
  logger.error(
    f'Multiple assignments found with content_key or parent_content_key {content_key} '
    f'for {assignment_configuration} and lms_user_id {lms_user_id}',
  )
  raise exc
```

PR #572, opened 2024-10-09
feat: introduce Gateway Aggregation / Backend-for-Frontend abstraction via LearnerPortalBFFAPIView

adamstankiewicz:
**Description:**

TBD

**Jira:**
[ENT-9427](https://2u-internal.atlassian.net/browse/ENT-9427)

**Merge checklist:**
- [ ] `./manage.py makemigrations` has been run
    - *Note*: This **must** be run if you modified any models.
      - It may or may not make a migration depending on exactly what you modified, but it should still be run.

**Post merge:**
- [ ] Ensure that your changes went out to the stage instance
- [ ] Deploy to prod instance

PR #577, opened 2024-10-28
feat: add bff handler context

brobro10000:
**Description:**
This application, bffs, has been added to enterprise-access. bff stands for "backend-for-frontend" and is designed to provide a consolidated API layer for supporting enterprise micro frontends (MFEs). The bffs application serves as an intermediary between frontend components and underlying services, centralizing business logic and data transformation within a single BFF layer. This approach aims to streamline data returned to the frontend, minimizing the need for additional transformation on the client side, and enabling frontend components to consume data in a more ready-to-use format.

Additionally, this PR introduces a HandlerContext component. HandlerContext will store the instantiated requests originating from API client calls, preparing them to be utilized by an upcoming response builder, designed to construct responses for each corresponding API call.

Each MFE will support route-based API calls, allowing for flexible orchestration of the frontend. This structure enables individual MFEs to manage route-specific API calls, ensuring a seamless flow between frontend routes and backend data, optimized for MFE interactions and reducing redundant API calls across the application.

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/bffs/context.py
+++ b/enterprise_access/apps/bffs/context.py
@@ -1,0 +1,59 @@
+"""
+HandlerContext for bffs app.
+"""
+from enterprise_access.apps.bffs import serializers
+
+
+class HandlerContext:
+    """
+    A context object for managing the state throughout the lifecycle of a Backend-for-Frontend (BFF) request.
+    The `HandlerContext` class stores request information, loaded data, and any errors and warnings
+    that may occur during the request.
+    Attributes:
+        request: The original request object containing information about the incoming HTTP request.
+        user: The original request user information about hte incoming HTTP request.
+        data: A dictionary to store data loaded and processed by the handlers.
+        errors: A list to store errors that occur during request processing.
+        warnings: A list to store warnings that occur during the request processing.
+        enterprise_customer_uuid: The enterprise customer the user is associated with.
+        lms_user_id: The id associated with the authenticated user.
+    """
+
+    def __init__(self, request):
+        """
+        Initializes the HandlerContext with request information, route, and optional initial data.
+        Args:
+            request: The incoming HTTP request.
+        """
+        self._request = request
+        self.data = {}  # Stores processed data for the response
+        self.errors = []  # Stores any errors that occur during processing
+        self.warnings = []  # Stores any warnings that occur during processing
+        self.enterprise_customer_uuid = None
+        self.lms_user_id = None
+
+    @property
+    def request(self):
+        return self._request
+
+    @property
+    def user(self):
+        return self._request.user
+
+    def add_error(self, **kwargs):
+        """
+        Adds an error to the context.
+        Output fields determined by the ErrorSerializer
+        """
+        serializer = serializers.ErrorSerializer(data=kwargs)
+        serializer.is_valid(raise_exception=True)
+        self.errors.append(serializer.data)
+
+    def add_warning(self, **kwargs):
+        """
+        Adds a warning to the context.
+        Output fields determined by the WarningSerializer
+        """
+        serializer = serializers.WarningSerializer(data=kwargs)
+        serializer.is_valid(raise_exception=True)
+        self.warnings.append(serializer.data)
--- a/enterprise_access/apps/bffs/serializers.py
+++ b/enterprise_access/apps/bffs/serializers.py
@@ -1,0 +1,30 @@
+"""
+Serializers for bffs.
+"""
+from rest_framework import serializers
+
+
+class BaseBFFMessageSerializer(serializers.Serializer):
+    """
+    Base Serializer for BFF messages.
+
+    Fields:
+        user_message (str): A user-friendly message.
+        developer_message (str): A more detailed message for debugging purposes.
+    """
+    developer_message = serializers.CharField()
+    user_message = serializers.CharField()
+
+    def create(self, validated_data):
+        return validated_data
+
+    def update(self, instance, validated_data):
+        return validated_data
+
+
+class ErrorSerializer(BaseBFFMessageSerializer):
+    pass
+
+
+class WarningSerializer(BaseBFFMessageSerializer):
+    pass
--- a/enterprise_access/apps/bffs/tests/test_context.py
+++ b/enterprise_access/apps/bffs/tests/test_context.py
@@ -1,0 +1,86 @@
+"""
+Text for the BFF context
+"""
+from django.test import RequestFactory, TestCase
+from rest_framework.exceptions import ValidationError
+
+from enterprise_access.apps.api_client.tests.test_constants import DATE_FORMAT_ISO_8601
+from enterprise_access.apps.bffs.context import HandlerContext
+from enterprise_access.apps.core.tests.factories import UserFactory
+from enterprise_access.utils import _curr_date
+
+
+class TestHandlerContext(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.factory = RequestFactory()
+        self.mock_user = UserFactory()
+
+    def test_handler_context_init(self):
+        request = self.factory.get('sample/api/call')
+        request.user = self.mock_user
+        context = HandlerContext(request)
+
+        self.assertEqual(context.request, request)
+        self.assertEqual(context.user, self.mock_user)
+        self.assertEqual(context.data, {})
+        self.assertEqual(context.errors, [])
+        self.assertEqual(context.warnings, [])
+        self.assertEqual(context.enterprise_customer_uuid, None)
+        self.assertEqual(context.lms_user_id, None)
+
+    def test_handler_context_add_error_serializer(self):
+        request = self.factory.get('sample/api/call')
+        request.user = self.mock_user
+        context = HandlerContext(request)
+        expected_output = {
+            "developer_message": "No enterprise uuid associated to the user mock-uuid",
+            "user_message": "You may not be associated with the enterprise.",
+        }
+        # Define kwargs for add_error
+        arguments = {
+            **expected_output,
+            "status": 403  # Add an attribute that is not explicitly defined in the serializer to verify
+        }
+        context.add_error(
+            **arguments
+        )
+        self.assertEqual(expected_output, context.errors[0])
+
+    def test_handler_context_add_error_serializer_is_valid(self):
+        request = self.factory.get('sample/api/call')
+        request.user = self.mock_user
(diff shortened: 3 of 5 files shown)


enterprise_access/apps/bffs/context.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    A context object for managing the state throughout the lifecycle of a Backend-for-Frontend (BFF) request.
+    The `HandlerContext` class stores request information, the current route, loaded data, and any errors
+    that may occur during the request.
+    Attributes:
+        request: The original request object containing information about the incoming HTTP request.
+        route: The route for which the response is being generated.
+        data: A dictionary to store data loaded and processed by the handlers.
+        errors: A list to store errors that occur during request processing.
+    """
```
adamstankiewicz, line 17: nit: docstring missing `warnings`, `enterprise_customer_uuid`, and `lms_user_id`.

enterprise_access/apps/bffs/context.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        self.request = request
+        self.user = request.user
+        self.data = {}  # Stores processed data for the response
+        self.errors = []  # Stores any errors that occur during processing
+        self.warnings = []  # Stores any warnings that occur during processing
+        self.enterprise_customer_uuid = None
+        self.lms_user_id = None
+
+    def add_error(self, user_message, developer_message):
```
adamstankiewicz, line 33: Should there be an `add_warning` method, too (might just still be WIP 😄)? See [reference prototype](https://github.com/openedx/enterprise-access/pull/572/files#diff-a2b7da65e3a660770612989f1169e4e5786851019b9e2c06b03822f7fa139bc9R53).

enterprise_access/apps/bffs/context.py
```diff
@@ -0,0 +1,49 @@
...
+        serializer = serializers.ErrorSerializer(data={
+            user_message,
+            developer_message,
+        })
+        serializer.is_valid(raise_exception=True)
+
+        self.errors.append({
+            "user_message": user_message,
+            "developer_message": developer_message,
+        })
```
adamstankiewicz, line 49: [suggestion] Would `self.errors.append(serializer.data)` also work here? The benefit of using the serializer is not just for the validation via `.is_valid` but also to return the serialized data.

enterprise_access/apps/bffs/context.py
```diff
@@ -0,0 +1,49 @@
+"""
+HandlerContext for bffs app.
+"""
+from enterprise_access.apps.bffs import serializers
+
+
+class HandlerContext:
+    """
+    A context object for managing the state throughout the lifecycle of a Backend-for-Frontend (BFF) request.
+    The `HandlerContext` class stores request information, the current route, loaded data, and any errors
```
adamstankiewicz, line 10: nit: it no longer stores "the current route".

adamstankiewicz, line 10: nit: it also stores warnings (e.g., `errors/warnings`)

enterprise_access/apps/bffs/context.py
```diff
@@ -0,0 +1,49 @@
...
+    that may occur during the request.
+    Attributes:
+        request: The original request object containing information about the incoming HTTP request.
+        route: The route for which the response is being generated.
```
adamstankiewicz, line 14: nit: no longer includes `route`.

enterprise_access/apps/bffs/context.py
```diff
@@ -0,0 +1,57 @@
...
+        Args:
+            request: The incoming HTTP request.
+        """
+        self.request = request
+        self.user = request.user
+        self.data = {}  # Stores processed data for the response
+        self.errors = []  # Stores any errors that occur during processing
+        self.warnings = []  # Stores any warnings that occur during processing
+        self.enterprise_customer_uuid = None
+        self.lms_user_id = None
```
brobro10000, line 34: _Note_: Something I considered but deferred on is the inclusion of url and params as a means of populating the `enteprise_customer_uuid` or `lms_user_id` within the `HandlerContext`. This would also help improve traceback of warnings/errors to specific calls within the BFF layer.

enterprise_access/apps/bffs/serializers.py
```diff
@@ -0,0 +1,40 @@
...
+        developer_message (str): A more detailed error message for debugging purposes.
+    """
+    developer_message = serializers.CharField()
+    user_message = serializers.CharField()
+
+    def create(self, validated_data):
+        return validated_data
+
+    def update(self, instance, validated_data):
+        return validated_data
```
brobro10000, line 22: Included to fix linting error of _abstract-method_
Screenshot 2024-10-31 at 10 34 38 AM

enterprise_access/apps/bffs/serializers.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    Serializer for error.
+
+    Fields:
+        user_message (str): A user-friendly error message.
+        developer_message (str): A more detailed error message for debugging purposes.
+    """
+    developer_message = serializers.CharField()
+    user_message = serializers.CharField()
+
```
brobro10000, line 17: _Note_:  Something I considered but deferred on was to serialize the request url and params as part of exposing. pertinent metadata within the errors for better observability.

enterprise_access/apps/bffs/context.py
```diff
@@ -0,0 +1,57 @@
...
+        lms_user_id: The id associated with the authenticated user.
+    """
+
+    def __init__(self, request):
+        """
+        Initializes the HandlerContext with request information, route, and optional initial data.
+        Args:
+            request: The incoming HTTP request.
+        """
+        self.request = request
```
iloveagent57, line 28: nit: you shouldn't be able to change the request for a context once the context is created, right?  If so, consider making the `request` protected and allowing "read-only access" via a property, e.g.
```python
def __init__(self, request):
    self._request = request
    # more instance variables

@property
def request(self):
    return self._request

@property
def user(self):
    return self._request.user
```

iloveagent57 (commented):
Looks good, small suggestion.

enterprise_access/apps/bffs/serializers.py
```diff
@@ -0,0 +1,40 @@
+"""
+Serializers for bffs.
+"""
+from rest_framework import serializers
+
+
+class ErrorSerializer(serializers.Serializer):
```
adamstankiewicz, line 7: [suggestion] Consider having a base serializer for `developer_message`, `user_message`, and the `create`/`update` methods.

enterprise_access/apps/bffs/context.py
adamstankiewicz: [inform] I agree. From our synchronous discussion, we could likely move the `initialize_common_context_data` from the reference prototype's `BaseHandler` into the `HandlerContext` so the route/base handler's themselves don't need to worry about it.

enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -0,0 +1,68 @@
...
+        request = self.factory.get('sample/api/call')
+        request.user = self.mock_user
+        context = HandlerContext(request)
+        expected_output = {
+            "developer_message": "No enterprise uuid associated to the user mock-uuid",
+            "user_message": "You may not be associated with the enterprise.",
+        }
+        arguments = {
+            **expected_output,
+            "status": 403
```
adamstankiewicz, line 46: [clarification] Does the `add_error` / `add_warning` methods accept/use `status` kwarg?

enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def test_handler_context_add_error(self):
+        request = self.factory.get('sample/api/call')
+        request.user = self.mock_user
+        context = HandlerContext(request)
+        expected_output = {
+            "developer_message": "No enterprise uuid associated to the user mock-uuid",
+            "user_message": "You may not be associated with the enterprise.",
+        }
+        arguments = {
```
adamstankiewicz, line 44: nit: is the `arguments` variable necessary vs. passing its value directly to `context.add_error({ ... })`.

enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -0,0 +1,68 @@
...
+            "user_message": "The data received might be out-dated",
+        }
+        arguments = {
+            **expected_output,
+            "status": 113
+        }
+        context.add_warning(
+            **arguments,
+        )
+        self.assertEqual(expected_output, context.errors[0])
```
adamstankiewicz, line 68: [clarification] This is asserting calling `add_warning` but is asserting on `context.errors[0]`. Should this be `context.warnings[0]`?

enterprise_access/apps/bffs/context.py
```diff
@@ -0,0 +1,64 @@
...
+        Adds a warning to the context.
+        Output fields determined by the WarningSerializer
+        Args:
+            user_message (str): A user-friendly warning message.
+            developer_message (str): A more detailed warning message for debugging purposes.
+        """
+        serializer = serializers.WarningSerializer(data=kwargs)
+        serializer.is_valid(raise_exception=True)
+
+        self.errors.append(serializer.data)
```
adamstankiewicz, line 64: `self.warnings.append`

adamstankiewicz (approved):

enterprise_access/apps/bffs/tests/test_context.py
brobro10000: It does not but its purpose is to specifically test that the serializer is correctly serializing only the relevant attributes. This also allows the output to be purely serializer driven.

enterprise_access/apps/bffs/tests/test_context.py
brobro10000: I can move it directly into the call for add_error. 👍🏽

enterprise_access/apps/bffs/tests/test_context.py
adamstankiewicz: Cool, makes sense. Just for posterity, might recommend just leaving a comment next to `status` to explain its rationale and intent to mitigate any confusion around why its included but not explicitly used.

adamstankiewicz (approved):
LGTM!

PR #581, opened 2024-11-01
feat: add base and learner portal bff handler

brobro10000:
**Description:**
Adds a `BaseHandler` which utilizes the `HandlerContext` responsible for the underlying structure for users of the `BaseHandler`.

Adds the `BaseLearnerPortalHandler` defines respective usable clients and loads data within the `HandlerContext`
It is the orchaestration layer for the MFE responsible for making the API calls and performing transformations. Currently the functions in this class are stubbed out to be completed in future work.

Adds the `DashboardHandler` which retrieves the data from the context to be consumed by the to be added `ResponseBuilder`

Adds MVP tests. The functions and tests will be iterated on specifically on within their separate pieces of work.

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/bffs/context.py
+++ b/enterprise_access/apps/bffs/context.py
@@ -32,6 +32,9 @@ class HandlerContext:
         self.enterprise_customer_uuid = None
         self.lms_user_id = None
 
+        # Set common context attributes
+        self.initialize_common_context_data()
+
     @property
     def request(self):
         return self._request
@@ -39,6 +42,24 @@ class HandlerContext:
     @property
     def user(self):
         return self._request.user
+
+    def initialize_common_context_data(self):
+        """
+        Initialize commonly used context attributes, such as enterprise customer UUID and LMS user ID.
+        """
+        enterprise_uuid_query_param = self.request.query_params.get('enterprise_customer_uuid')
+        enterprise_uuid_post_param = None
+        if self.request.method == 'POST':
+            enterprise_uuid_post_param = self.request.data.get('enterprise_customer_uuid')
+
+        enterprise_customer_uuid = enterprise_uuid_query_param or enterprise_uuid_post_param
+        if enterprise_customer_uuid:
+            self.enterprise_customer_uuid = enterprise_customer_uuid
+        else:
+            raise ValueError("enterprise_customer_uuid is required for this request.")
+
+        # Set lms_user_id from the authenticated user object in the request
+        self.lms_user_id = getattr(self.user, 'lms_user_id', None)
 
     def add_error(self, **kwargs):
         """
--- a/enterprise_access/apps/bffs/handlers.py
+++ b/enterprise_access/apps/bffs/handlers.py
@@ -1,0 +1,396 @@
+""""
+Handlers for bffs app.
+"""
+
+import logging
+
+from enterprise_access.apps.api_client.license_manager_client import LicenseManagerUserApiClient
+from enterprise_access.apps.bffs.context import HandlerContext
+
+logger = logging.getLogger(__name__)
+
+
+class BaseHandler:
+    """
+    A base handler class that provides shared core functionality for different BFF handlers.
+    The `BaseHandler` includes core methods for loading data and adding errors to the context.
+    """
+
+    def __init__(self, context: HandlerContext):
+        """
+        Initializes the BaseHandler with a HandlerContext.
+        Args:
+            context (HandlerContext): The context object containing request information and data.
+        """
+        self.context = context
+
+    def load_and_process(self):
+        """
+        Loads and processes data. This method should be extended by subclasses to implement
+        specific data loading and transformation logic.
+        """
+        raise NotImplementedError("Subclasses must implement `load_and_process` method.")
+
+    def add_error(self, **kwargs):
+        """
+        Adds an error to the context.
+        Output fields determined by the ErrorSerializer
+        """
+        self.context.add_error(**kwargs)
+
+    def add_warning(self, **kwargs):
+        """
+        Adds an error to the context.
+        Output fields determined by the WarningSerializer
+        """
+        self.context.add_warning(**kwargs)
+
+
+class BaseLearnerPortalHandler(BaseHandler):
+    """
+    A base handler class for learner-focused routes.
+
+    The `BaseLearnerHandler` extends `BaseHandler` and provides shared core functionality
+    across all learner-focused page routes, such as the learner dashboard, search, and course routes.
+    """
+    def __init__(self, context):
+        """
+         Initializes the BaseLearnerPortalHandler with a HandlerContext and API clients.
+         Args:
+             context (HandlerContext): The context object containing request information and data.
+         """
+        super().__init__(context)
+        self.license_manager_client = LicenseManagerUserApiClient(self.context.request)
+
+    def load_and_process(self):
+        """
+        Loads and processes data. This is a basic implementation that can be overridden by subclasses.
+
+        The method in this class simply calls common learner logic to ensure the context is set up.
+        """
+        try:
+            # Retrieve and process subscription licenses. Handles activation and auto-apply logic.
+            # TODO: retrieve enterprise customer metadata,ENT-9629
+            # self.load_enterprise_customer()
+            self.load_and_process_subscription_licenses()
+
+            # Retrieve default enterprise courses and enroll in the redeemable ones
+            self.load_default_enterprise_courses()
+            self.enroll_in_redeemable_default_courses()
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.exception("Error loading learner portal handler")
+            self.add_error(
+                user_message="An error occurred while loading and processing common learner logic.",
+                developer_message=f"Error: {e}",
+            )
+
+    def load_subscription_licenses(self):
+        """
+        Load subscription licenses for the learner.
+        """
+        subscriptions_result = self.license_manager_client.get_subscription_licenses_for_learner(
+            enterprise_customer_uuid=self.context.enterprise_customer_uuid
+        )
+        self.transform_subscriptions_result(subscriptions_result)
+
+    def get_subscription_licenses(self):
+        """
+        Get subscription licenses.
+        """
+        return self.context.data['subscriptions'].get('subscription_licenses', [])
+
+    def get_subscription_licenses_by_status(self):
+        """
+        Get subscription licenses by status.
+        """
+        return self.context.data['subscriptions'].get('subscription_licenses_by_status', {})
+
+    def transform_subscription_licenses(self, subscription_licenses):
+        """
+        Transform subscription licenses data if needed.
+        """
+        return [
+            {
+                'uuid': subscription_license.get('uuid'),
+                'status': subscription_license.get('status'),
+                'user_email': subscription_license.get('user_email'),
+                'activation_date': subscription_license.get('activation_date'),
+                'last_remind_date': subscription_license.get('last_remind_date'),
+                'revoked_date': subscription_license.get('revoked_date'),
+                'activation_key': subscription_license.get('activation_key'),
+                'subscription_plan': subscription_license.get('subscription_plan', {}),
+            }
+            for subscription_license in subscription_licenses
+        ]
+
+    def transform_subscriptions_result(self, subscriptions_result):
+        """
+        Transform subscription licenses data if needed.
+        """
+        subscription_licenses = subscriptions_result.get('results', [])
+        subscription_licenses_by_status = {}
+
+        transformed_licenses = self.transform_subscription_licenses(subscription_licenses)
+
+        for subscription_license in transformed_licenses:
+            status = subscription_license.get('status')
+            if status not in subscription_licenses_by_status:
+                subscription_licenses_by_status[status] = []
+
+            subscription_licenses_by_status[status].append(subscription_license)
+
+        subscriptions_data = {
+            'customer_agreement': subscriptions_result.get('customer_agreement', {}),
+            'subscription_licenses': transformed_licenses,
+            'subscription_licenses_by_status': subscription_licenses_by_status,
+        }
+        self.context.data['subscriptions'] = subscriptions_data
+
+    def check_has_activated_license(self):
+        """
+        Check if the user has an activated license.
+
+        Returns:
+            bool: True if the user has an activated license, False otherwise.
+        """
+        subscription_licenses_by_status = self.get_subscription_licenses_by_status()
+        return bool(subscription_licenses_by_status.get('activated'))
+
+    def process_subscription_licenses(self):
+        """
+        Process loaded subscription licenses, including performing side effects such as:
+            * Checking if there is an activated license
+            * Checking and activating assigned licenses
+            * Checking and auto applying licenses
+
+        This method is called after `load_subscription_licenses` to handle further actions based
+        on the loaded data.
+
+        If the `subscriptions` field does not exist, raises a KeyError
+        """
+        if not self.context.data['subscriptions']:
+            raise KeyError("Unable to retrieve subscriptions.")
+
+        # Check if user already has 'activated' license(s). If so, no further action is needed.
+        if self.check_has_activated_license():
+            return
+
+        # Check if there are 'assigned' licenses that need to be activated
+        self.check_and_activate_assigned_license()
+
+        # Check if the user should be auto-applied a license
+        self.check_and_auto_apply_license()
+
+    def load_and_process_subscription_licenses(self):
+        """
+        Helper to load subscription licenses into the context then processes them
+        by determining by:
+            * Checking if there is an activated license
+            * Checking and activating assigned licenses
+            * Checking and auto applying licenses
+        """
+        self.load_subscription_licenses()
+        self.process_subscription_licenses()
+
+    def check_and_activate_assigned_license(self):
+        """
+        Check if there are assigned licenses that need to be activated.
+        """
+        subscription_licenses = self.get_subscription_licenses()
+        subscription_licenses_by_status = self.get_subscription_licenses_by_status()
+        assigned_licenses = subscription_licenses_by_status.get('assigned', [])
+        activated_licenses = []
+        for subscription_license in assigned_licenses:
+            activation_key = subscription_license.get('activation_key')
+            if activation_key:
+                try:
+                    # Perform side effect: Activate the assigned license
+                    activated_license = self.license_manager_client.activate_license(activation_key)
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    logger.exception(f"Error activating license {subscription_license.get('uuid')}")
+                    self.add_error(
+                        user_message="An error occurred while activating a subscription license.",
+                        developer_message=f"License UUID: {subscription_license.get('uuid')}, Error: {e}",
+                    )
+                    return
+
+                # Update the subscription_license data with the activation status and date; the activated license is not
+                # returned from the API, so we need to manually update the license object we have available.
+                transformed_activated_subscription_license = self.transform_subscription_licenses(activated_license)
+                activated_licenses.append(transformed_activated_subscription_license)
+            else:
+                logger.error(f"Activation key not found for license {subscription_license.get('uuid')}")
+                self.add_error(
+                    user_message="An error occurred while activating a subscription license.",
+                    developer_message=f"Activation key not found for license {subscription_license.get('uuid')}",
+                )
+
+        # Update the subscriptions.subscription_licenses_by_status context with the modified licenses data
+        updated_activated_licenses = subscription_licenses_by_status.get('activated', [])
+        updated_activated_licenses.extend(activated_licenses)
+        subscription_licenses_by_status['activated'] = updated_activated_licenses
+        remaining_assigned_licenses = [
+            subscription_license
+            for subscription_license in assigned_licenses
+            if subscription_license not in activated_licenses
+        ]
+        if remaining_assigned_licenses:
+            subscription_licenses_by_status['assigned'] = remaining_assigned_licenses
+        else:
+            subscription_licenses_by_status.pop('assigned', None)
+        self.context.data['subscriptions']['subscription_licenses_by_status'] = subscription_licenses_by_status
+
+        # Update the subscriptions.subscription_licenses context with the modified licenses data
+        updated_subscription_licenses = []
+        for subscription_license in subscription_licenses:
+            for activated_license in activated_licenses:
+                if subscription_license.get('uuid') == activated_license.get('uuid'):
+                    updated_subscription_licenses.append(activated_license)
+                    break
+                updated_subscription_licenses.append(subscription_license)
+        self.context.data['subscriptions']['subscription_licenses'] = updated_subscription_licenses
+
+    def check_and_auto_apply_license(self):
+        """
+        Check if auto-apply licenses are available and apply them to the user.
+        """
+        subscription_licenses_by_status = self.get_subscription_licenses_by_status()
+        has_assigned_licenses = subscription_licenses_by_status.get('assigned', [])
+        if has_assigned_licenses or self.check_has_activated_license():
+            # Skip auto-apply if user already has an activated license or assigned licenses
+            return
+
+        customer_agreement = self.context.data['subscriptions'].get('customer_agreement', {})
+        has_subscription_plan_for_auto_apply = (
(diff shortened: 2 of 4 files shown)


enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+                if subscription_plan.get('enterprise_catalog_uuid') in course.get('applicable_catalog_uuids'):
+                    redeemable_default_courses.append((course, subscription_license))
+                    break
+
+        for redeemable_course, subscription_license in redeemable_default_courses:
+            # Enroll in redeemable courses (stubbed)
+            if not self.context.data.get('enrolled_default_courses'):
+                self.context.data['enrolled_default_courses'] = []
+
+            self.context.data['enrolled_default_courses'].append({
```
adamstankiewicz, line 332: nit: perhaps we rename `enrolled_default_courses` to `default_enterprise_enrollment_realizations` to more closely reflect the model objects created during successful auto-enrollment?

This area of code is likely to be updated a bit through https://2u-internal.atlassian.net/browse/ENT-9632 regardless, so we could defer for that ticket, too.

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+        This method overrides the `load_and_process` method in `BaseLearnerPortalHandler`.
+        """
+        # Call the common learner logic from the base class
+        super().load_and_process()
+
+        try:
+            # Load data specific to the dashboard route
+            self.context.data['enterprise_course_enrollments'] = self.get_enterprise_course_enrollments()
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            self.add_error(
```
adamstankiewicz, line 360: Should we include a `logger.exception` here?

enterprise_access/apps/bffs/tests/test_handlers.py
```diff
@@ -0,0 +1,125 @@
...
+        self.request.query_params = {
+            'enterprise_customer_uuid': self.mock_enterprise_customer_uuid
+        }
+        self.request.user = self.mock_user
+        self.context = HandlerContext(self.request)
+        self.base_handler = BaseHandler(self.context)
+        self.base_learner_portal_handler = BaseLearnerPortalHandler(self.base_handler)
+        self.dasboard_handler = DashboardHandler(self.base_learner_portal_handler)
+
+    # TODO: Update tests once stubbed our function updated
```
adamstankiewicz, line 118: nit: stubbed ou**t**?

enterprise_access/apps/bffs/tests/test_handlers.py
```diff
@@ -0,0 +1,125 @@
...
+    def test_load_and_process(self):
+        dashboard_handler = self.dasboard_handler
```
adamstankiewicz, line 120: nit: typo in `self.dasboard_handler`

enterprise_access/apps/bffs/tests/test_handlers.py
```diff
@@ -0,0 +1,125 @@
...
+        self.mock_enterprise_customer_uuid = self.faker.uuid4()
+        self.request = self.factory.get('sample/api/call')
+        self.request.query_params = {
+            'enterprise_customer_uuid': self.mock_enterprise_customer_uuid
+        }
+        self.request.user = self.mock_user
+        self.context = HandlerContext(self.request)
+        self.base_handler = BaseHandler(self.context)
+        self.base_learner_portal_handler = BaseLearnerPortalHandler(self.base_handler)
+        self.dasboard_handler = DashboardHandler(self.base_learner_portal_handler)
```
adamstankiewicz, line 116: [curious] Is it necessary to instantiate each of these yourself? When you instantiate `DashboardHandler`, you're already instantiating `BaseLearnerPortalHandler` and `BaseHandler` since they subclass each other.

E.g. I believe you could just do:

```python
self.dashboard_handler = DashboardHandler(self.context)
```

...without needing `self.base_handler` and `self.base_learner_portal_handler`.

(similar comment for the `TestBaseLearnerPortalHandler` class).

enterprise_access/apps/bffs/tests/test_handlers.py
```diff
@@ -0,0 +1,125 @@
...
+        self.request.user = self.mock_user
+        self.context = HandlerContext(self.request)
+        self.base_handler = BaseHandler(self.context)
+
+    def test_base_handler_uninitialized_load_and_process(self):
+        base_handler = self.base_handler
+        with self.assertRaises(NotImplementedError):
+            base_handler.load_and_process()
+
+    def test_base_handler_add_error_serializer(self):
```
adamstankiewicz, line 31: nit: might remove the `_serializer` suffix as the handler's `add_error` is largely just a wrapper around the `HandlerContext`'s `add_error`. Given we already have sufficient coverage on the `HandlerContext`'s `add_error` and `add_warning` methods, we probably don't necessarily need to re-test the passing of extra kwargs like `status`.

Similar feedback to testing `is_valid`. Given the serializer `is_valid` stuff is already tested via `test_context.py`, I'm not sure we need to retest its logic here. The most important thing to test here, IMHO, is just that the `BaseHandler` is wired up to the `HandlerContext`'s `add_error` and `add_warning` from a happy path perspective (valid kwargs).

enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -1,38 +1,80 @@
...
+            self.assertEqual(context.enterprise_customer_uuid, self.mock_enterprise_customer_uuid)
+            self.assertEqual(context.lms_user_id, self.mock_user.lms_user_id)
 
     def test_handler_context_add_error_serializer(self):
-        request = self.factory.get('sample/api/call')
-        request.user = self.mock_user
-        context = HandlerContext(request)
+        context = self.context
         expected_output = {
             "developer_message": "No enterprise uuid associated to the user mock-uuid",
```
adamstankiewicz, line 79: [curious] Where is `mock-uuid` coming from? Typically the user id is an integer, not a uuid so having this (and a few other places below) state its a uuid could be misleading?

enterprise_access/apps/bffs/tests/test_handlers.py
```diff
@@ -0,0 +1,125 @@
...
+        self.mock_user = UserFactory()
+        self.faker = Faker()
+
+        self.mock_enterprise_customer_uuid = self.faker.uuid4()
+        self.request = self.factory.get('sample/api/call')
+        self.request.query_params = {
+            'enterprise_customer_uuid': self.mock_enterprise_customer_uuid
+        }
+        self.request.user = self.mock_user
+        self.context = HandlerContext(self.request)
```
adamstankiewicz, line 23: Given these lines seem to be repeated fairly verbatim throughout the different test classes, I wonder if it'd make sense to create something like a `TestBaseHandlerMixin` that test cases to extend to be a bit more DRY?

```python
class TestBaseHandlerMixin(TestCase):
    def setUp(self):
        super().setUp()
        self.factory = RequestFactory()
        self.mock_user = UserFactory()
        self.faker = Faker()

        self.mock_enterprise_customer_uuid = self.faker.uuid4()
        self.request = self.factory.get('sample/api/call')
        self.request.query_params = {
            'enterprise_customer_uuid': self.mock_enterprise_customer_uuid
        }
        self.request.user = self.mock_user
        self.context = HandlerContext(self.request)

class TestBaseHandler(TestBaseHandlerMixin):
    # ...
```

enterprise_access/apps/bffs/tests/test_handlers.py
```diff
@@ -0,0 +1,125 @@
...
+        self.base_learner_portal_handler = BaseLearnerPortalHandler(self.base_handler)
+        self.dasboard_handler = DashboardHandler(self.base_learner_portal_handler)
+
+    # TODO: Update tests once stubbed our function updated
+    def test_load_and_process(self):
+        dashboard_handler = self.dasboard_handler
+        dashboard_handler.load_and_process()
+        self.assertEqual(
+            dashboard_handler.context.data['enterprise_course_enrollments'],
+            dashboard_handler.get_enterprise_course_enrollments()
```
adamstankiewicz, line 124: [curious] I believe we may want to hardcode the expected data vs. relying on calling `dashboard_handler.get_enterprise_course_enrollments()` here? Presumably the data in the `context.data['enterprise_course_enrollments']` will always equal `dashboard_handler.get_enterprise_course_enrollments`?

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+            # TODO: retrieve enterprise customer metadata,ENT-9629
+            # self.load_enterprise_customer()
+            self.load_subscription_licenses()
+            self.process_subscription_licenses()
+
+            # Retrieve default enterprise courses and enroll in the redeemable ones
+            self.load_default_enterprise_courses()
+            self.enroll_in_redeemable_default_courses()
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            self.add_error(
```
adamstankiewicz, line 80: nit: worthwhile to add `logger.exception` here?

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+        )
+        self.transform_subscriptions_result(subscriptions_result)
+
+    def get_subscription_licenses(self):
+        """
+        Get subscription licenses.
+        """
+        if not (subscriptions := self.context.data['subscriptions']):
+            self.load_subscription_licenses()
+        return subscriptions.get('subscription_licenses', [])
```
adamstankiewicz, line 103: As is, in the case where `self.load_subscription_licenses` is called, I don't believe this line would properly have the updated context data. For example, if we wanted to have the side effect of `load_subscription_licenses`, would it need to be something like the following:

```python
if not (subscriptions := self.context.data['subscriptions']):
    self.load_subscription_licenses()

# Retrieve subscriptions data from the context again, now that it's been updated via `load_subscription_licenses`.
subscriptions = self.context.data['subscriptions']
return subscriptions.get('subscription_licenses', [])
```

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def get_subscription_licenses_by_status(self):
+        """
+        Get subscription licenses by status.
+        """
+        if not (subscriptions := self.context.data['subscriptions']):
+            self.load_subscription_licenses()
+        return subscriptions.get('subscription_licenses_by_status', {})
```
adamstankiewicz, line 111: Similar feedback as `get_subscription_licenses`.

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+        on the loaded data.
+        """
+        # Check if user already has 'activated' license(s). If so, no further action is needed.
+        if self.check_has_activated_license():
+            return
+
+        # Check if there are 'assigned' licenses that need to be activated
+        self.check_and_activate_assigned_license()
+
+        # Check if there user should be auto-applied a license
```
adamstankiewicz, line 181: nit: 

```suggestion
        # Check if the user should be auto-applied a license
```

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+            # Retrieve and process subscription licenses. Handles activation and auto-apply logic.
+            # TODO: retrieve enterprise customer metadata,ENT-9629
+            # self.load_enterprise_customer()
+            self.load_and_process_subscription_licenses()
+
+            # Retrieve default enterprise courses and enroll in the redeemable ones
+            self.load_default_enterprise_courses()
+            self.enroll_in_redeemable_default_courses()
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.exception(f"Error loading learner portal handler: {e}")
```
adamstankiewicz, line 82: nit: I believe `logger.exception` already includes the stack trace; I don't believe including `{e}` is necessary here.

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+        assigned_licenses = subscription_licenses_by_status.get('assigned', [])
+        activated_licenses = []
+        for subscription_license in assigned_licenses:
+            activation_key = subscription_license.get('activation_key')
+            if activation_key:
+                try:
+                    # Perform side effect: Activate the assigned license
+                    activated_license = self.license_manager_client.activate_license(activation_key)
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    logger.exception(f"Error activating license {subscription_license.get('uuid')}: {e}")
```
adamstankiewicz, line 210: nit: When using `logger.exception`, I believe the exception's stack trace is already included; I think the `{e}` is unnecessary here.

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+        try:
+            # Perform side effect: Auto-apply license
+            auto_applied_license = self.license_manager_client.auto_apply_license(customer_agreement.get('uuid'))
+            if auto_applied_license:
+                # Update the context with the auto-applied license data
+                subscription_licenses_by_status['activated'] =\
+                    self.transform_subscription_licenses([auto_applied_license])
+                self.context.data['subscriptions']['subscription_licenses_by_status'] = subscription_licenses_by_status
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.exception(f"Error auto-applying license: {e}")
```
adamstankiewicz, line 286: nit: I believe the `{e}` is unnecessary here, as `logger.exception` should already included the exception stack trace.

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+        Loads and processes data for the learner dashboard route.
+
+        This method overrides the `load_and_process` method in `BaseLearnerPortalHandler`.
+        """
+        # Call the common learner logic from the base class
+        try:
+            # Load data specific to the dashboard route
+            self.context.data['enterprise_course_enrollments'] = self.get_enterprise_course_enrollments()
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.exception(f"Error retrieving enterprise_course_enrollments: {e}")
```
adamstankiewicz, line 359: nit: I believe the `{e}` here is unnecessary, as `logger.exception` includes the exception's stack trace already.

enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -1,40 +1,82 @@
...
+            'query_params': True,
+            'data': False,
+        },
+        {
+            'query_params': False,
+            'data': False,
+        }
+    )
+    @ddt.unpack
+    def test_handler_context_enterprise_customer_uuid(self, query_params, data):
```
adamstankiewicz, line 62: nit: perhaps rename slightly to suggest testing parsing params.

```suggestion
    def test_handler_context_enterprise_customer_uuid_param(self, query_params, data):
```

enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -1,40 +1,82 @@
...
+        self.request.data = {
+            'enterprise_customer_uuid': self.mock_enterprise_customer_uuid
+        }
+        self.context = HandlerContext(self.request)
 
     def test_handler_context_init(self):
-        request = self.factory.get('sample/api/call')
-        request.user = self.mock_user
-        context = HandlerContext(request)
+        context = HandlerContext(self.request)
```
adamstankiewicz, line 33: [nit/curious] Would it make sense to rely on `self.context` here as it's already instantiated in the `setUp`?

```python
self.assertEqual(self.context.request, self.request)
# ...
```

enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -1,40 +1,82 @@
...
+
+        self.mock_enterprise_customer_uuid = self.faker.uuid4()
+        self.request = self.factory.get('sample/api/call')
+        self.request.user = self.mock_user
+        self.request.query_params = {
+            'enterprise_customer_uuid': self.mock_enterprise_customer_uuid
+        }
+        self.request.data = {
+            'enterprise_customer_uuid': self.mock_enterprise_customer_uuid
+        }
```
adamstankiewicz, line 29: nit: I'm not sure if it makes sense to initialize the mock request with **both** `query_params` AND `data` simultaneously, as typically it'd be either one or the other.

enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -1,40 +1,82 @@
...
+        else:
+            context = HandlerContext(request)
+            self.assertEqual(context.enterprise_customer_uuid, self.mock_enterprise_customer_uuid)
+            self.assertEqual(context.lms_user_id, self.mock_user.lms_user_id)
 
     def test_handler_context_add_error_serializer(self):
-        request = self.factory.get('sample/api/call')
-        request.user = self.mock_user
-        context = HandlerContext(request)
+        context = self.context
```
adamstankiewicz, line 77: nit: is defining this `context` variable needed versus using `self.context` directly below? e.g.

```python
self.context.add_error(...)
```

enterprise_access/apps/bffs/tests/test_handlers.py
```diff
@@ -0,0 +1,113 @@
...
+
+class TestBaseLearnerPortalHandler(TestBaseHandlerMixin):
+    def setUp(self):
+        super().setUp()
+        self.base_learner_portal_handler = BaseLearnerPortalHandler(self.context)
+
+    # TODO: Test pertaining to currently stubbed out functions deferred for future tickets
+
+
+class TestDashboardHandler(TestBaseHandlerMixin):
```
adamstankiewicz, line 77: I think these extending a `TestBaseHandlerMixin` came out nicely :)

adamstankiewicz (commented):
Looking great, left a couple suggestions to consider and a few nits.

enterprise_access/apps/bffs/tests/test_context.py
brobro10000: You are correct but there is a test that checks for whether the query params or data is included. Selectively removing them from the request from `setUp` makes for a cleaner test.
``` py
  @ddt.data(
        {
            'query_params': True,
            'data': True,
        },
        {
            'query_params': False,
            'data': True,
        },
        {
            'query_params': True,
            'data': False,
        },
        {
            'query_params': False,
            'data': False,
        }
    )
    @ddt.unpack
    def test_handler_context_enterprise_customer_uuid(self, query_params, data):
        if not query_params:
            del self.request.query_params['enterprise_customer_uuid']
        if not data:
            del self.request.data['enterprise_customer_uuid']
        request = self.request
        if not (query_params or data):
            with self.assertRaises(ValueError):
                HandlerContext(request)
        else:
            context = HandlerContext(request)
            self.assertEqual(context.enterprise_customer_uuid, self.mock_enterprise_customer_uuid)
            self.assertEqual(context.lms_user_id, self.mock_user.lms_user_id)
 ```

File: enterprise_access/apps/bffs/handlers.py [hunk 23]
```diff
@@ -0,0 +1,396 @@
...
+    across all learner-focused page routes, such as the learner dashboard, search, and course routes.
+    """
+    def __init__(self, context):
+        """
+         Initializes the BaseLearnerPortalHandler with a HandlerContext and API clients.
+         Args:
+             context (HandlerContext): The context object containing request information and data.
+         """
+        super().__init__(context)
+        self.context = context
```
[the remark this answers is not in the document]
brobro10000: Removing it presents the following linting error
Screenshot 2024-11-06 at 1 28 32 PM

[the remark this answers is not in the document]
adamstankiewicz: Which line did you remove? It looks like you removed `super().__init__(context)`, not `self.context = context`?

enterprise_access/apps/bffs/handlers.py
```diff
@@ -0,0 +1,396 @@
...
+            enterprise_customer_uuid=self.context.enterprise_customer_uuid
+        )
+        self.transform_subscriptions_result(subscriptions_result)
+
+    def get_subscription_licenses(self):
+        """
+        Get subscription licenses.
+        """
+        if not (subscriptions := self.context.data['subscriptions']):
+            self.load_subscription_licenses()
```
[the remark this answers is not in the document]
brobro10000: Updated to include a new `KeyError` in `process_subscription_licenses`

``` py
        if not self.context.data['subscriptions']:
            raise KeyError("Unable to retrieve subscriptions.")
```

enterprise_access/apps/bffs/handlers.py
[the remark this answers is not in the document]
brobro10000: yep you're right, i misread the comment 👍🏽  updating now

enterprise_access/apps/bffs/tests/test_context.py
brobro10000: So I made a slight refactor to determine whether a `request_object` exist to parse out the `enterprise_customer_uuid`. Reason being was the original tuple was failing on retrieving the `data` object if neither existed.

It determines which `request_object` exist, either a `query_parameter` or `data` object, and throws a `ValueError` if neither exist. If a `request_object` exist, then we parse out the `enterprise_customer_uuid` and determine it does exist. If not, throw a `KeyError` that a enterprise_customer_uuid is required.

To the point of a data and query_param should not exist in tandem, it is still covered in the tests for full coverage to validate that we handle it correctly.

enterprise_access/apps/bffs/context.py
```diff
@@ -40,6 +43,23 @@ def request(self):
...
+        Initialize commonly used context attributes, such as enterprise customer UUID and LMS user ID.
+        """
+        if request_object := (getattr(self.request, 'query_params', {}) or getattr(self.request, 'data', None)):
+            enterprise_customer_uuid = request_object.get('enterprise_customer_uuid')
+            if enterprise_customer_uuid:
+                self.enterprise_customer_uuid = enterprise_customer_uuid
+            else:
+                raise KeyError("enterprise_customer_uuid is required for this request.")
+        else:
+            raise ValueError("A query parameter or data object is required for this request.")
```
adamstankiewicz, line 57: Generally, `query_params` will always exist in practice (outside of mocked test requests); `data` will only exist for POST requests. This `else` considers the case where neither is present, which I don't think can happen in practice.

Might be best to ensure the test more accurately reflects actual usage by making a POST call when `data` is truthy, and perhaps updating this method to account for `self.request.method === 'POST`, e.g.:

```python
def initialize_common_context_data(self):
        """
        Initialize commonly used context attributes, such as enterprise customer UUID and LMS user ID.
        """
        # Set enterprise_customer_uuid from request parameters or previously set context
        enterprise_uuid_query_param = self.request.query_params.get('enterprise_customer_uuid')
        enterprise_uuid_post_param = None
        if self.request.method == 'POST':
            enterprise_uuid_post_param = self.request.data.get('enterprise_customer_uuid')

        enterprise_customer_uuid = enterprise_uuid_query_param or enterprise_uuid_post_param
        if enterprise_customer_uuid:
            self.enterprise_customer_uuid = enterprise_customer_uuid
        else:
            raise ValueError("enterprise_customer_uuid is required for this request.")

        # Set lms_user_id from the authenticated user object in the request
        self.lms_user_id = self.user.lms_user_id
```

```python
@ddt.data(
        {
            'query_params': True,
            'data': True,
        },
        {
            'query_params': False,
            'data': True,
        },
        {
            'query_params': True,
            'data': False,
        },
        {
            'query_params': False,
            'data': False,
        },
    )
    @ddt.unpack
    def test_handler_context_enterprise_customer_uuid_param(self, query_params, data):
        if not query_params:
            self.request.query_params = {}

        if data:
            self.request = self.factory.post('sample/api/call')
            self.request.data = {
                'enterprise_customer_uuid': self.mock_enterprise_customer_uuid
            }

        if not (query_params or data):
            with self.assertRaises(ValueError):
                HandlerContext(self.request)
        else:
            self.assertEqual(self.context.enterprise_customer_uuid, self.mock_enterprise_customer_uuid)
            self.assertEqual(self.context.lms_user_id, self.mock_user.lms_user_id)
```

I feel this might be preferable over adding complexity to the implementation to appease mock test data? 🤔

brobro10000: Updated 👍🏽

adamstankiewicz (approved):
Nice job!

PR #584, opened 2024-11-07
fix: Check preferred_course_run_key for normalized_metadata

brobro10000:
**Description:**
Takes into account the `preferred_course_run_key` when extracting the `normalized_metadata` from assignments.

**Jira:**
ENT-9688

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/tests/test_utils.py
+++ b/enterprise_access/apps/content_assignments/tests/test_utils.py
@@ -5,13 +5,39 @@ Tests for Enterprise Access content_assignments utils.
 import ddt
 from django.test import TestCase
 
+from enterprise_access.apps.api_client.tests.test_constants import DATE_FORMAT_ISO_8601
 from enterprise_access.apps.content_assignments.constants import BRAZE_ACTION_REQUIRED_BY_TIMESTAMP_FORMAT
+from enterprise_access.apps.content_assignments.tests.factories import LearnerContentAssignmentFactory
 from enterprise_access.apps.content_assignments.utils import (
     get_self_paced_normalized_start_date,
     has_time_to_complete,
     is_within_minimum_start_date_threshold
 )
-from enterprise_access.utils import _curr_date, _days_from_now
+from enterprise_access.utils import _curr_date, _days_from_now, get_normalized_metadata_for_assignment
+
+mock_course_run_1 = {
+    'start_date': _days_from_now(-370, DATE_FORMAT_ISO_8601),
+    'end_date': _days_from_now(350, DATE_FORMAT_ISO_8601),
+    'enroll_by_date': _days_from_now(-363, DATE_FORMAT_ISO_8601),
+    'enroll_start_date': _days_from_now(-380, DATE_FORMAT_ISO_8601),
+    'content_price': 100,
+}
+
+mock_course_run_2 = {
+    'start_date': _days_from_now(-70, DATE_FORMAT_ISO_8601),
+    'end_date': _days_from_now(50, DATE_FORMAT_ISO_8601),
+    'enroll_by_date': _days_from_now(-63, DATE_FORMAT_ISO_8601),
+    'enroll_start_date': _days_from_now(-80, DATE_FORMAT_ISO_8601),
+    'content_price': 100,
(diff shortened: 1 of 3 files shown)


enterprise_access/apps/content_assignments/tests/test_utils.py
```diff
@@ -192,3 +236,41 @@ def test_get_self_paced_normalized_start_date_self_paced(self, start_date, end_d
...
+                'is_assigned_course_run': False,
+                'preferred_course_run_key': None,
+                'content_key': 'edX+DemoX',
+            },
+            'expected_output': mock_advertised_course_run,
+        },
+    )
+    @ddt.unpack
+    def test_get_normalized_metadata_for_assignment(self, assignment, expected_output):
+        assignment_obj = SimpleNamespace(**assignment)
```
adamstankiewicz, line 271: Might recommend using `LearnerContentAssignmentFactory` versus creating a pseudo-assignment instance based on the `assignment` dict in the test case data using `SimpleNamespace`. This would more accurately reflect how the util function is used in practice.

```python
from enterprise_access.apps.content_assignments.tests.factories import LearnerContentAssignmentFactory
assignment_obj = LearnerContentAssignmentFactory(**assignment)
```

enterprise_access/apps/content_assignments/tests/test_utils.py
```diff
@@ -192,3 +236,41 @@ def test_get_self_paced_normalized_start_date_self_paced(self, start_date, end_d
         else:
             assert get_self_paced_normalized_start_date(start_date, end_date, course_metadata) == \
                    start_date
+
+    @ddt.data(
+        # Expects the content_key for an is_assigned_course_run
```
adamstankiewicz, line 241: [nit/clarification] What is meant by `"for an is_assigned_course_run"`? Perhaps a minor revision to:

```suggestion
        # Expects course run associated with configured content_key for run-based assignment
```

enterprise_access/utils.py
```diff
@@ -224,11 +224,16 @@ def get_normalized_metadata_for_assignment(assignment, content_metadata):
...
     normalized_metadata_by_run = content_metadata.get('normalized_metadata_by_run', {})
-    return normalized_metadata_by_run.get(assignment.content_key, {})
+    # Course run-based assignments with an existing course-run based content_key
+    if assignment.is_assigned_course_run:
+        return normalized_metadata_by_run.get(content_key, {})
+    # A top level course assignment with a preferred course run key
+    if content_key := assignment.preferred_course_run_key:
+        return normalized_metadata_by_run.get(content_key, {})
+    # A top level assignment with no preferred course run.
+    return content_metadata.get('normalized_metadata', {})
```
adamstankiewicz, line 236: I believe we could potentially always rely on `preferred_course_run_key` regardless of assignment content type (course vs. course run), which would simplify this method's logic. For a run-based assignment, `preferred_course_run_key` should match the configured `content_key`; for a course-based assignment, `preferred_course_run_key` should represent the advertised run at the time of assignment allocation.

Querying the prod read-replica, there's no run-based assignments with `preferred_course_run_key=NULL` and every run-based assignment's `content_key` matches its `preferred_course_run_key`. There are ~8k course-based assignments with `preferred_course_run_key=NULL` (latest one was created May 1, which is around when this field was first introduced), which I don't think we can do much about.

What do you think of something like the following?

```python
def get_normalized_metadata_for_assignment(assignment, content_metadata):
    normalized_metadata_by_run = content_metadata.get('normalized_metadata_by_run', {})
    if preferred_course_run_key := assignment.preferred_course_run_key:
        return normalized_metadata_by_run.get(preferred_course_run_key, {})
  
    # Return current advertised course run metadata if preferred_course_run_key is NULL (i.e.,
    # impacting legacy course-based assignments created pre-May 2024).
    return content_metadata.get('normalized_metadata', {})
```

brobro10000: Updated 👍🏽

iloveagent57 (approved):

adamstankiewicz (approved):

PR #586, opened 2024-11-12
feat: initialize enterprise metadata in HandlerContext; complete auto-apply business logic

adamstankiewicz:
**Jira:**
[ENT-9629](https://2u-internal.atlassian.net/browse/ENT-9629)

**Description:**

* Converted additional `HandlerContext` class attributes as private, with `@property` getter functions.
* Refactors `initialize_common_context_data` as `_initialize_common_context_data`:
  * Extended to optionally support both `enterprise_customer_uuid` and `enterprise_customer_slug`.
  * Extended to include an attribute for `enterprise_features` (returned by `enterprise-learner` API, described below).
  * Initializes context `data` with enterprise customer metadata (mostly parity with transforms on the existing Learner Portal request) via the `enterprise-learner` LMS API.
    * If no slug/uuid is provided, resolves to the request user's `active=True` enterprise customer user.
    * If only slug is provided, finds linked enterprise customer matching the requested slug and sets the uuid.
    * If only uuid is provided, finds linked enterprise customer matching the requested uuid and sets the slug.
    * If uuid and/or slug is provided, but no matching linked enterprise customer is found, falls back to request the enterprise customer metadata as a staff user (if applicable).
* Completes the auto-apply business logic (to be tested at the viewset layer in separate ticket) by extending the conditional check to include whether the enterprise customer has an `identity_provider` configured.
* Integrates with `self.lms_user_api_client.get_default_enterprise_enrollment_intentions_learner_status` in `BaseLearnerPortalHandler`.
* Integrates with `self.lms_user_api_client.get_enterprise_course_enrollments` in `DashboardHandler`.
* Tests.

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api_client/lms_client.py
+++ b/enterprise_access/apps/api_client/lms_client.py
@@ -64,21 +64,31 @@ class LmsApiClient(BaseOAuthClient):
             "learners/",
         )
 
-    def get_enterprise_customer_data(self, enterprise_customer_uuid):
+    def get_enterprise_customer_data(self, enterprise_customer_uuid=None, enterprise_customer_slug=None):
         """
-        Gets the data for an EnterpriseCustomer for the given uuid.
+        Gets the data for an EnterpriseCustomer for the given uuid or slug.
 
         Arguments:
             enterprise_customer_uuid (string): id of the enterprise customer
+            enterprise_customer_slug (string): slug of the enterprise customer
         Returns:
             dictionary containing enterprise customer metadata
         """
+        if enterprise_customer_uuid:
+            endpoint = f'{self.enterprise_customer_endpoint}{enterprise_customer_uuid}/'
+        elif enterprise_customer_slug:
+            endpoint = f'{self.enterprise_customer_endpoint}?slug={enterprise_customer_slug}'
+        else:
+            raise ValueError('Either enterprise_customer_uuid or enterprise_customer_slug is required.')
 
         try:
-            endpoint = f'{self.enterprise_customer_endpoint}{enterprise_customer_uuid}/'
             response = self.client.get(endpoint, timeout=settings.LMS_CLIENT_TIMEOUT)
             response.raise_for_status()
-            return response.json()
+            if enterprise_customer_uuid:
+                # If we're fetching by UUID, we expect a single result
+                return response.json()
+            # If we're fetching by slug, we expect a list of results
+            return response.json().get('results', [])[0]
         except requests.exceptions.HTTPError as exc:
             logger.exception(exc)
             raise
@@ -406,12 +416,66 @@ class LmsUserApiClient(BaseUserApiClient):
     enterprise_api_base_url = f"{settings.LMS_URL}/enterprise/api/v1/"
     enterprise_learner_portal_api_base_url = f"{settings.LMS_URL}/enterprise_learner_portal/api/v1/"
 
+    enterprise_learner_endpoint = f"{enterprise_api_base_url}enterprise-learner/"
     default_enterprise_enrollment_intentions_learner_status_endpoint = (
         f'{enterprise_api_base_url}default-enterprise-enrollment-intentions/learner-status/'
     )
     enterprise_course_enrollments_endpoint = (
         f'{enterprise_learner_portal_api_base_url}enterprise_course_enrollments/'
     )
+
+    def get_enterprise_customers_for_user(self, username, traverse_pagination=False):
+        """
+        Fetches enterprise learner data for a given username.
+
+        Arguments:
+            username (str): Username of the learner
+
+        Returns:
+            dict: Dictionary representation of the JSON response from the API
+        """
+        query_params = {
+            'username': username,
+        }
+        results = []
+        initial_response_data = None
+        current_response = None
+        next_url = self.enterprise_learner_endpoint
+        try:
+            while next_url:
+                current_response = self.get(
+                    next_url,
+                    params=query_params,
+                    timeout=settings.LMS_CLIENT_TIMEOUT
+                )
+                current_response.raise_for_status()
+                data = current_response.json()
+
+                if not initial_response_data:
+                    # Store the initial response data (first page) for later use
+                    initial_response_data = data
+
+                # Collect results from the current page
+                results.extend(data.get('results', []))
+
+                # If pagination is enabled, continue with the next page; otherwise, break
+                next_url = data.get('next') if traverse_pagination else None
+
+            consolidated_response = {
+                **initial_response_data,
+                'next': None,
+                'previous': None,
+                'count': len(results),
+                'num_pages': 1,
+                'results': results,
+            }
+            return consolidated_response
+        except requests.exceptions.HTTPError as exc:
+            logger.exception(
+                f"Failed to fetch enterprise learner for learner {username}: {exc} "
+                f"Response content: {current_response.content if current_response else None}"
+            )
+            raise
 
     def get_default_enterprise_enrollment_intentions_learner_status(self, enterprise_customer_uuid):
         """
--- a/enterprise_access/apps/api_client/lms_client.py
+++ b/enterprise_access/apps/api_client/lms_client.py
@@ -74,8 +74,10 @@ class LmsApiClient(BaseOAuthClient):
             dictionary containing enterprise customer metadata
         """
         if enterprise_customer_uuid:
+            # Returns a dict
             endpoint = f'{self.enterprise_customer_endpoint}{enterprise_customer_uuid}/'
         elif enterprise_customer_slug:
+            # Returns a list of dicts
             endpoint = f'{self.enterprise_customer_endpoint}?slug={enterprise_customer_slug}'
         else:
             raise ValueError('Either enterprise_customer_uuid or enterprise_customer_slug is required.')
@@ -83,11 +85,10 @@ class LmsApiClient(BaseOAuthClient):
         try:
             response = self.client.get(endpoint, timeout=settings.LMS_CLIENT_TIMEOUT)
             response.raise_for_status()
-            if enterprise_customer_uuid:
-                # If we're fetching by UUID, we expect a single result
-                return response.json()
-            # If we're fetching by slug, we expect a list of results
-            return response.json().get('results', [])[0]
+            payload = response.json()
+            if results := payload.get('results'):
+                return results[0]
+            return payload
         except requests.exceptions.HTTPError as exc:
             logger.exception(exc)
             raise
--- a/enterprise_access/apps/api_client/lms_client.py
+++ b/enterprise_access/apps/api_client/lms_client.py
@@ -42,7 +42,6 @@ def all_pages_enterprise_group_members_cache_key(
 
 
 class LmsApiClient(BaseOAuthClient):
-
     """
     API client for calls to the LMS service.
     """
--- a/enterprise_access/apps/bffs/context.py
+++ b/enterprise_access/apps/bffs/context.py
@@ -188,13 +188,12 @@ class HandlerContext:
         )
 
         # Update the context data with the enterprise customer user information
-        self.data = {
-            **self.data,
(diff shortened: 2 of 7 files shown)


enterprise_access/apps/bffs/context.py
```diff
@@ -43,23 +54,203 @@ def request(self):
...
+        )
+        enterprise_customer = self._determine_enterprise_customer_for_display(
+            active_enterprise_customer=active_enterprise_customer,
+            requested_enterprise_customer=requested_enterprise_customer,
+            staff_enterprise_customer=staff_enterprise_customer,
+        )
+
+        # Update the context data with the enterprise customer user information
+        self.data = {
+            **self.data,
```
adamstankiewicz, line 192: TODO: use `self.data.update({ ... })` instead

enterprise_access/apps/bffs/handlers.py
```diff
@@ -69,29 +73,94 @@ def load_and_process(self):
...
+
             # Retrieve and process subscription licenses. Handles activation and auto-apply logic.
-            # TODO: retrieve enterprise customer metadata,ENT-9629
-            # self.load_enterprise_customer()
             self.load_and_process_subscription_licenses()
 
-            # Retrieve default enterprise courses and enroll in the redeemable ones
-            self.load_default_enterprise_courses()
-            self.enroll_in_redeemable_default_courses()
+            # # Retrieve default enterprise courses and enroll in the redeemable ones
```
adamstankiewicz, line 82: TODO: double comment

enterprise_access/apps/bffs/handlers.py
```diff
@@ -69,29 +73,94 @@ def load_and_process(self):
...
+            self.load_default_enterprise_enrollment_intentions()
+            self.enroll_in_redeemable_default_enterprise_enrollment_intentions()
         except Exception as e:  # pylint: disable=broad-exception-caught
             logger.exception("Error loading learner portal handler")
             self.add_error(
                 user_message="An error occurred while loading and processing common learner logic.",
                 developer_message=f"Error: {e}",
             )
 
+    def transform_enterprise_customers(self):
```
adamstankiewicz, line 92: TODO: i believe this is missing transforming the `staff_enterprise_customer` property

enterprise_access/apps/bffs/context.py
```diff
@@ -43,23 +54,203 @@ def request(self):
...
+                enterprise_customer_user
+                for enterprise_customer_user in enterprise_customer_users
+                if enterprise_customer_user.get('active', False)
+            ),
+            None
+        )
+        active_enterprise_customer = (
+            active_enterprise_customer_user.get('enterprise_customer')
+            if active_enterprise_customer_user else None
+        )
```
brobro10000, line 148: [nit/non-blocking] This functions doing a lot. I wonder if its worth chunking up some of the logic into a separate function?
Ex. `get_enterprise_customer_users_data`, `get_active_enterprise_customer`

enterprise_access/apps/bffs/context.py
```diff
@@ -43,23 +54,203 @@ def request(self):
...
+            try:
+                staff_enterprise_customer = self.lms_api_client.get_enterprise_customer_data(
+                    enterprise_customer_uuid=self.enterprise_customer_uuid,
+                    enterprise_customer_slug=self.enterprise_customer_slug,
+                )
+            except Exception as e:  # pylint: disable=broad-except
+                self.add_error(
+                    user_message='Error retrieving enterprise customer data',
+                    developer_message=str(e)
+                )
```
brobro10000, line 177: [nit/non-blocking] Another opportunity to chunk logic into a separate function here.
Ex. `get_staff_enterprise_customer`

enterprise_access/apps/bffs/context.py
```diff
@@ -43,23 +54,203 @@ def request(self):
...
+
+        slug_matches_active_enterprise_customer = (
+            active_enterprise_customer and active_enterprise_customer.get('slug') == self.enterprise_customer_slug
+        )
+        uuid_matches_active_enterprise_customer = (
+            active_enterprise_customer and active_enterprise_customer.get('uuid') == self.enterprise_customer_uuid
+        )
+        request_matches_active_enterprise_customer = (
+            slug_matches_active_enterprise_customer or uuid_matches_active_enterprise_customer
+        )
```
brobro10000, line 223: [nit/non-blocking] could this be extracted into a helper function
Ex. `validate_active_enterprise_customer`

brobro10000 (approved):
Great job. A couple of nits that can be deferred or ignored. Approving to unblock.

brobro10000 (approved):

enterprise_access/apps/bffs/handlers.py
```diff
@@ -69,29 +73,97 @@ def load_and_process(self):
...
+        """
+        Transform enterprise customer metadata retrieved by self.context.
+        """
+        if enterprise_customer := self.context.data.get('enterprise_customer'):
```
iloveagent57, line 96: nit: `if self.context.enterprise_customer`

iloveagent57: actually, how about a loop like this?
```python
for customer_record_key in ('enterprise_customer', 'active_enterprise_customer', 'staff_enterprise_customer'):
    if not customer_record := self.context.data.get(customer_record_key):
        continue
    self.context.data[customer_record_key] = self.transform_enterprise_customer(customer_record)
```

enterprise_access/apps/bffs/handlers.py
```diff
@@ -69,29 +73,97 @@ def load_and_process(self):
...
+    def transform_enterprise_customer_user(self, enterprise_customer_user):
+        """
+        Transform the enterprise customer user data.
+
+        Args:
+            enterprise_customer_user: The enterprise customer user data.
+        Returns:
+            The transformed enterprise customer user data.
+        """
+        enterprise_customer = enterprise_customer_user.get('enterprise_customer')
```
iloveagent57, line 119: The `enterprise_customer` dict in the ECU response payload won't be "full", right? Does that matter here?

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -406,13 +415,67 @@ class LmsUserApiClient(BaseUserApiClient):
...
+                    initial_response_data = data
+
+                # Collect results from the current page
+                results.extend(data.get('results', []))
+
+                # If pagination is enabled, continue with the next page; otherwise, break
+                next_url = data.get('next') if traverse_pagination else None
+
+            consolidated_response = {
+                **initial_response_data,
```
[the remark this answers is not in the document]
adamstankiewicz: Correct, and then the subsequent lines below overrides relevant fields like `next`, `prev`, etc. to reflect the traversed pagination results being a single page.

The primary reason to spread `**initial_response_data` is bring along any other existing pagination fields unrelated to page length, i.e. I believe `**initial_response_data` brings along `start: 0` as well without needing to explicitly define it. The `start` value from the initial response is unchanged regardless of `traverse_pagination`.

enterprise_access/apps/bffs/context.py
```diff
@@ -43,23 +54,224 @@ def request(self):
...
             enterprise_uuid_post_param = self.request.data.get('enterprise_customer_uuid')
+            enterprise_slug_post_param = self.request.data.get('enterprise_customer_slug')
 
         enterprise_customer_uuid = enterprise_uuid_query_param or enterprise_uuid_post_param
-        if enterprise_customer_uuid:
-            self.enterprise_customer_uuid = enterprise_customer_uuid
-        else:
-            raise ValueError("enterprise_customer_uuid is required for this request.")
+        self._enterprise_customer_uuid = enterprise_customer_uuid
+        enterprise_customer_slug = enterprise_slug_query_param or enterprise_slug_post_param
```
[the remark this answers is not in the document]
adamstankiewicz: Can you expand on what the proposed `assert` would be enforcing? Would this suggest that the `enterprise_customer_slug` is required?

This PR refactors to make both the `enterprise_customer_uuid` and `enterprise_customer_slug` params optional. When neither is provided, it returns the request user's linked enterprise customer with `active=True` (similar to some existing frontend business logic).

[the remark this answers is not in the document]
iloveagent57: Ah, ok, that makese sense. Nix this one.

enterprise_access/apps/bffs/context.py
```diff
@@ -43,23 +54,224 @@ def request(self):
...
+                for enterprise_customer_user in enterprise_customer_users
+                if self._enterprise_customer_matches_slug_or_uuid(enterprise_customer_user.get('enterprise_customer'))
+            ),
+            None
+        )
+
+        # If no enterprise customer user is found for the requested customer (i.e., request user not explicitly
+        # linked), but the request user is staff, attempt to retrieve enterprise customer metadata from the
+        # `/enterprise-customer/` LMS API endpoint instead.
+        staff_enterprise_customer = self._get_staff_enterprise_customer(enterprise_customer_user_for_requested_customer)
```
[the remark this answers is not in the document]
adamstankiewicz: Agreed, good suggestion for clean up to make it easier to grok.

enterprise_access/apps/bffs/handlers.py
adamstankiewicz: I like the loop suggestion to make the implementation more DRY. I believe the conditional logic around `if enterprise_customer_users := self.context.data.get('all_linked_enterprise_customer_users'):` will still need to remain, though.

```python
def transform_enterprise_customers(self):
    """
    Transform enterprise customer metadata retrieved by self.context.
    """
    for customer_record_key in ('enterprise_customer', 'active_enterprise_customer', 'staff_enterprise_customer'):
        if not (customer_record := self.context.data.get(customer_record_key)):
            continue
        self.context.data[customer_record_key] = self.transform_enterprise_customer(customer_record)

    if enterprise_customer_users := self.context.data.get('all_linked_enterprise_customer_users'):
        self.context.data['all_linked_enterprise_customer_users'] = [
            self.transform_enterprise_customer_user(enterprise_customer_user)
            for enterprise_customer_user in enterprise_customer_users
        ]
```

enterprise_access/apps/bffs/handlers.py
adamstankiewicz: The `enterprise-learner` API uses the same `EnterpriseCustomerSerializer` that the list/detail enterprise customer API uses; it's the "full" metadata. This is the same enterprise customer metadata currently consumed by the frontends (both learner/admin).

That said, when cross-checking to verify, I noticed a discrepancy between `admin_users` list in the `enterprise-learner` API (this data source) vs. the `enterprise-customer` detail API. The `enterprise-learner` API _always_ denotes `admin_users` as an empty list, even if the customer has admin users...

I filed a [bug ticket](https://2u-internal.atlassian.net/browse/ENT-9741) and fixed the bug through this PR: https://github.com/openedx/edx-enterprise/pull/2287

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -64,21 +63,31 @@ def enterprise_group_members_endpoint(self, group_uuid):
...
         try:
-            endpoint = f'{self.enterprise_customer_endpoint}{enterprise_customer_uuid}/'
             response = self.client.get(endpoint, timeout=settings.LMS_CLIENT_TIMEOUT)
             response.raise_for_status()
-            return response.json()
+            if enterprise_customer_uuid:
+                # If we're fetching by UUID, we expect a single result
+                return response.json()
+            # If we're fetching by slug, we expect a list of results
+            return response.json().get('results', [])[0]
```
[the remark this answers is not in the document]
adamstankiewicz: Pushed up a commit with a refactor. Ended up with:

```python
payload = response.json()
if results := payload.get('results'):
  return results[0]
return payload
```

iloveagent57 (approved):

PR #588, opened 2024-11-13
feat: add response builder to the BFF

brobro10000:
**Description:**
Adds the response builder to the BFF app. The response builders main responsibility is to take the now loaded data from the context layer, and serialize it in preparation to be returned to the user. This includes adding warnings, errors and status codes.

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api_client/tests/test_utils.py
+++ b/enterprise_access/apps/api_client/tests/test_utils.py
@@ -1,10 +1,17 @@
 """
 Helper utilities for api_client tests.
 """
-import requests
+from django.test import TestCase
+from faker import Faker
+from requests import Response
+
+from enterprise_access.apps.api_client.constants import LicenseStatuses
+from enterprise_access.apps.api_client.tests.test_constants import DATE_FORMAT_ISO_8601, DATE_FORMAT_ISO_8601_MS
+from enterprise_access.apps.core.tests.factories import UserFactory
+from enterprise_access.utils import _days_from_now
 
 
-class MockResponse(requests.Response):
+class MockResponse(Response):
     """
     Useful for mocking HTTP responses, especially for code that relies on raise_for_status().
     """
@@ -15,3 +22,95 @@ class MockResponse(requests.Response):
 
     def json(self):  # pylint: disable=arguments-differ
         return self.json_data
+
+
+class MockEnterpriseMetadata(TestCase):
+    """
+    Mock enterprise metadata
+    """
+    def setUp(self):
+        super().setUp()
+        self.faker = Faker()
+
+        self.base_paginated_response = {
+            "next": None,
+            "previous": None,
+            "count": 0,
+            "results": [],
+        }
+
+        self.mock_enterprise_customer_uuid = self.faker.uuid4()
+        self.mock_enterprise_customer_slug = "test_slug"
+        self.mock_enterprise_catalog_uuid = self.faker.uuid4()
+        self.mock_user = UserFactory()
+        self.mock_user_email = self.mock_user.email
+
+
+class MockLicenseManagerMetadataMixin(MockEnterpriseMetadata):
+    """
+    Mixin for the TestLicenseManagerUserApiClient
+    """
+    def setUp(self):
+        super().setUp()
+
+        self.mock_learner_license_uuid = self.faker.uuid4()
+        self.mock_learner_license_activation_uuid = self.faker.uuid4()
+        self.mock_license_activation_key = self.faker.uuid4()
+        self.mock_auto_apply_uuid = self.faker.uuid4()
+        self.mock_subscription_plan_uuid = self.faker.uuid4()
+        self.mock_customer_agreement_uuid = self.faker.uuid4()
+
+        self.mock_subscription_plan = {
+            "title": "mock_title",
+            "uuid": self.mock_subscription_plan_uuid,
+            "start_date": _days_from_now(-50, DATE_FORMAT_ISO_8601),
+            "expiration_date": _days_from_now(50, DATE_FORMAT_ISO_8601),
+            "enterprise_customer_uuid": self.mock_enterprise_customer_uuid,
+            "enterprise_catalog_uuid": self.mock_enterprise_catalog_uuid,
+            "is_active": True,
+            "is_current": True,
+            "is_revocation_cap_enabled": False,
+            "days_until_expiration": 50,
+            "days_until_expiration_including_renewals": 50,
+            "is_locked_for_renewal_processing": False,
+            "should_auto_apply_licenses": False,
+            "created": _days_from_now(-60, DATE_FORMAT_ISO_8601)
+        }
+        self.mock_customer_agreement = {
+            "uuid": self.mock_customer_agreement_uuid,
+            "enterprise_customer_uuid": self.mock_enterprise_customer_uuid,
+            "enterprise_customer_slug": self.mock_enterprise_customer_slug,
+            "default_enterprise_catalog_uuid": self.mock_enterprise_catalog_uuid,
+            "disable_expiration_notifications": False,
+            "net_days_until_expiration": 50,
+            "subscription_for_auto_applied_licenses": self.mock_subscription_plan_uuid,
+            "available_subscription_catalogs": [
+                self.mock_enterprise_catalog_uuid,
+            ],
+            "enable_auto_applied_subscriptions_with_universal_link": False,
+            "has_custom_license_expiration_messaging": False,
+            "modal_header_text": None,
+            "expired_subscription_modal_messaging": None,
+            "button_label_in_modal": None,
+            "url_for_button_in_modal": None
+        }
+        self.mock_subscription_license = {
+            "uuid": self.mock_learner_license_activation_uuid,
+            "status": LicenseStatuses.ACTIVATED,
+            "user_email": self.mock_user_email,
+            "activation_date": _days_from_now(-10, DATE_FORMAT_ISO_8601_MS),
+            "last_remind_date": _days_from_now(-5, DATE_FORMAT_ISO_8601_MS),
+            "subscription_plan_uuid": self.mock_subscription_plan_uuid,
+            "revoked_date": None,
+            "activation_key": self.mock_license_activation_key,
+            "subscription_plan": self.mock_subscription_plan
+        }
+        self.mock_learner_license_activation_response = {
+            **self.mock_subscription_license,
+            "uuid": self.mock_learner_license_activation_uuid,
+        }
+        self.mock_learner_license_auto_apply_response = {
+            **self.mock_subscription_license,
+            'customer-agreement': self.mock_customer_agreement,
+            'subscription_plan': self.mock_subscription_plan,
+        }
--- a/enterprise_access/apps/bffs/response_builder.py
+++ b/enterprise_access/apps/bffs/response_builder.py
@@ -1,0 +1,130 @@
+"""
+Response Builder Module for bffs app
+"""
+from enterprise_access.apps.bffs.serializers import LearnerDashboardResponseSerializer
+
+
+class BaseResponseBuilder:
+    """
+    A base response builder class that provides shared core functionality for different response builders.
+
+    The `BaseResponseBuilder` includes methods for building response data and can be extended by specific
+    response builders like `LearnerDashboardResponseBuilder` or `CourseResponseBuilder`.
+    """
+
+    def __init__(self, context):
+        """
+        Initializes the BaseResponseBuilder with a HandlerContext.
+
+        Args:
+            context (HandlerContext): The context object containing data, errors, and request information.
+        """
+        self.context = context
+
+    def build(self):
+        """
+        Builds the response data. This method should be overridden by subclasses to implement
+        specific response formatting logic.
+
+        Returns:
+            dict: A dictionary containing the response data.
+        """
+        raise NotImplementedError("Subclasses must implement the `build` method.")
+
+    def add_errors_warnings_to_response(self, response_data):
+        """
+        Adds any errors to the response data.
+        """
+        response_data['errors'] = self.context.errors
+        response_data['warnings'] = self.context.warnings
+        return response_data
+
+    # TODO Revisit this function in ENT-9633 to determine if 200 is ok for a nested errored response
+    def get_status_code(self):
+        """
+        Gets the current status code from the context.
+
+        Returns:
+            int: The HTTP status code.
+        """
+        return self.context.status_code
+
+
+class BaseLearnerResponseBuilder(BaseResponseBuilder):
+    """
+    A base response builder class for learner-focused routes.
+
+    The `BaseLearnerResponseBuilder` extends `BaseResponseBuilder` and provides shared core functionality
+    for building responses across all learner-focused page routes.
+    """
+
+    def common_response_logic(self, response_data=None):
+        """
+        Applies common response logic for learner-related responses.
+
+        Args:
+            response_data (dict): The initial response data.
+
+        Returns:
+            dict: The modified response data with common logic applied.
+        """
+        if not response_data:
+            response_data = {}
+        response_data['enterprise_customer_user_subsidies'] =\
+            self.context.data.get('enterprise_customer_user_subsidies', {})
+        return response_data
+
+    def build(self):
+        """
+        Builds the base response data for learner routes.
+
+        This method can be overridden by subclasses to provide route-specific logic.
+
+        Returns:
+            dict: A tuple containing the base response data and status code.
+        """
+        # Initialize response data with common learner-related logic
+        response_data = self.common_response_logic()
+
+        # Add any errors, etc.
+        response_data = self.add_errors_warnings_to_response(response_data)
+
+        # Return the response data and status code
+        return response_data, self.get_status_code()
+
+
+class LearnerDashboardResponseBuilder(BaseLearnerResponseBuilder):
+    """
+    A response builder for the learner dashboard route.
+
+    The `LearnerDashboardResponseBuilder` extends `BaseLearnerResponseBuilder` to extract and format data
+    relevant to the learner dashboard page.
+    """
+
+    def build(self):
+        """
+        Builds the response data for the learner dashboard route.
+
+        This method overrides the `build` method in `BaseResponseBuilder`.
+
+        Returns:
+            dict: A tuple containing the learner dashboard serialized response data and status code.
+        """
+        # Initialize the response data with common learner-related fields
+        response_data = self.common_response_logic()
+
+        # Add specific fields related to the learner dashboard
+        response_data.update({
+            'enterprise_course_enrollments': self.context.data.get('enterprise_course_enrollments', []),
+        })
+
+        # Add any errors and warnings to the response
+        response_data = self.add_errors_warnings_to_response(response_data)
+
+        # Serialize and validate the response
+        serializer = LearnerDashboardResponseSerializer(data=response_data)
+        serializer.is_valid(raise_exception=True)
+        serialized_data = serializer.validated_data
+
+        # Return the response data and status code
+        return serialized_data, self.get_status_code()
--- a/enterprise_access/apps/bffs/serializers.py
+++ b/enterprise_access/apps/bffs/serializers.py
@@ -28,3 +28,183 @@ class ErrorSerializer(BaseBFFMessageSerializer):
 
 class WarningSerializer(BaseBFFMessageSerializer):
     pass
+
+
+class BaseResponseSerializer(serializers.Serializer):
+    """
+    Serializer for base response.
+    """
+
+    errors = ErrorSerializer(many=True, required=False, default=list)
+    warnings = WarningSerializer(many=True, required=False, default=list)
+
+    def create(self, validated_data):
+        return validated_data
+
+    def update(self, instance, validated_data):
+        return validated_data
+
+
+class CustomerAgreementSerializer(serializers.Serializer):
+    """
+    Serializer for customer agreement.
+    """
+
+    uuid = serializers.UUIDField()
+    available_subscription_catalogs = serializers.ListField(child=serializers.UUIDField())
+    default_enterprise_catalog_uuid = serializers.UUIDField()
+    net_days_until_expiration = serializers.IntegerField()
+    disable_expiration_notifications = serializers.BooleanField()
+    enable_auto_applied_subscriptions_with_universal_link = serializers.BooleanField()
+    subscription_for_auto_applied_licenses = serializers.UUIDField(allow_null=True)
+
+    def create(self, validated_data):
+        return validated_data
+
+    def update(self, instance, validated_data):
+        return validated_data
+
+
+class SubscriptionPlanSerializer(serializers.Serializer):
+    """
+    Serializer for subscription plan.
+    """
+
+    uuid = serializers.UUIDField()
+    title = serializers.CharField()
+    enterprise_catalog_uuid = serializers.UUIDField()
+    is_active = serializers.BooleanField()
+    is_current = serializers.BooleanField()
+    start_date = serializers.DateTimeField()
+    expiration_date = serializers.DateTimeField()
+    days_until_expiration = serializers.IntegerField()
+    days_until_expiration_including_renewals = serializers.IntegerField()
+    should_auto_apply_licenses = serializers.BooleanField()
+
+    def create(self, validated_data):
+        return validated_data
+
+    def update(self, instance, validated_data):
+        return validated_data
+
+
+class SubscriptionLicenseSerializer(serializers.Serializer):
+    """
+    Serializer for subscription license.
+    """
+
+    uuid = serializers.UUIDField()
+    status = serializers.CharField()
+    user_email = serializers.EmailField()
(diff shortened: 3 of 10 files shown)


enterprise_access/apps/bffs/response_builder.py
```diff
@@ -0,0 +1,137 @@
...
+        Builds the base response data for learner routes.
+
+        This method can be overridden by subclasses to provide route-specific logic.
+
+        Returns:
+            dict: A dictionary containing the base response data.
+        """
+        # Initialize response data with common learner-related logic
+        response_data = {}
+        response_data = self.common_response_logic(response_data)
```
adamstankiewicz, line 95: I wonder if we should update `common_response_logic` to make `response_data` optional, e.g.

```python
def common_response_logic(self, response_data=None):
    if not response_data:
        response_data = {}
    # ...
```

enterprise_access/apps/bffs/response_builder.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Builds the response data for the learner dashboard route.
+
+        This method overrides the `build` method in `BaseResponseBuilder`.
+
+        Returns:
+            dict: A dictionary containing the learner dashboard response data.
+        """
+        # Initialize the response data with common learner-related fields
+        response_data = self.common_response_logic({})
```
adamstankiewicz, line 122: Related to other comment, the `{}` probably could be removed if `response_data` arg in `common_response_logic` is optional.

enterprise_access/apps/bffs/response_builder.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Add specific fields related to the learner dashboard
+        response_data.update({
+            'enterprise_course_enrollments': self.context.data.get('enterprise_course_enrollments', {}),
```
adamstankiewicz, line 126: I believe, at least for now, the enrollments data is a flat list. Should the fallback value be `[]` vs. `{}`?

enterprise_access/apps/bffs/response_builder.py
```diff
@@ -0,0 +1,137 @@
...
+        })
+
+        # Add any errors and warnings to the response
+        response_data = self.add_errors_warnings_to_response(response_data)
+
+        # Serialize and validate the response
+        serializer = LearnerDashboardResponseSerializer(response_data)
+        serialized_data = serializer.data
```
adamstankiewicz, line 134: [curious] Do we need to do `serializer.is_valid(raises_exception=True/False)` here?

Also, FWIW, I think if we do use `serializer.is_valid()`, we may want to use `serializer.validated_data` vs. `serializer.data`.

enterprise_access/apps/bffs/serializers.py
```diff
@@ -28,3 +30,196 @@ class ErrorSerializer(BaseBFFMessageSerializer):
...
+    def to_representation(self, instance):
+        """
+        Override to_representation method to return ordered representation
+        with errors/warnings at the end of the response.
+        """
+        representation = super().to_representation(instance)
+
+        ordered_representation = OrderedDict(representation)
+
+        # Remove errors and warnings from the main response (they will be re-added at the end)
```
adamstankiewicz, line 58: nit: maybe just a slight revision to the comment to add "at the end **of the response**" and/or perhaps combine the lines 63-64 below all under a single comment block.

```python
# Temporarily remove errors and warnings from the main response, and re-add them show they appear at the end of the response.
errors = ordered_representation.pop('errors', [])
warnings = ordered_representation.pop('warnings', [])
ordered_representation['errors'] = errors
ordered_representation['warnings'] = warnings
```

enterprise_access/apps/bffs/tests/test_builders.py
```diff
@@ -0,0 +1,161 @@
...
+            self.context.add_warning(
+                **self.mock_warning
+            )
+            expected_output['warnings'] = [self.mock_warning]
+        response_data = self.base_response_builder.add_errors_warnings_to_response(sample_response)
+        self.assertEqual(response_data, expected_output)
+
+    @ddt.data(
+        {
+            'status_code': 201
```
adamstankiewicz: nit: might recommend having the status code represent an error case (e.g., `400`) vs a `201`. Rationale: I'm not sure the API should return a 201 (versus a 200), even if it does have a side effect (e.g., auto-applying a subs license).

Considering using 

```python
from rest_framework import status
status.HTTP_400_BAD_REQUEST
```

enterprise_access/apps/bffs/tests/test_builders.py
```diff
@@ -0,0 +1,161 @@
...
+        },
+        {
+            'status_code': None
+        }
+    )
+    @ddt.unpack
+    def test_get_status_code(self, status_code):
```
adamstankiewicz: nit: might be worth dropping a `TODO` comment referencing https://2u-internal.atlassian.net/browse/ENT-9633, as  if we do add the `status_code` class attribute in `HandlerContext`, it may come with a minor change or two to this test case.

enterprise_access/apps/bffs/tests/test_response_builders.py
```diff
@@ -0,0 +1,143 @@
...
+from enterprise_access.apps.bffs.context import HandlerContext
+from enterprise_access.apps.bffs.response_builder import BaseLearnerResponseBuilder, BaseResponseBuilder
+from enterprise_access.apps.bffs.tests.test_handlers import TestBaseHandlerMixin
+
+
+@ddt.ddt
+class TestBaseResponseBuilder(TestBaseHandlerMixin):
+    def setUp(self):
+        super().setUp()
+        self.base_response_builder = BaseResponseBuilder(self.context)
```
adamstankiewicz, line 14: [inform] Note: In https://github.com/openedx/enterprise-access/pull/586, now that `HandlerContext` is making API requests during initialization, I ended up having to instantiate context/handlers outside of `setUp` to apply the mock via `@mock.patch`. I'm assuming a similar issue might happen here.

In the linked PR, I ended up removing `TestBaseHandlerMixin` in favor of `TestHandlerContextMixin` re-used as a test utility across the test suites for `HandlerContext` and the handlers. Do you think it makes sense to get that linked PR merged in first or land this PR first, given the likelihood of merge conflicts around these the test mixin?

Also, food for thought: I wonder at what point in the handlers/response we should be relying on a mocked `HandlerContext` so tests for the response builder, etc. don't necessary need to mock stuff related to the context (focusing more on the response builder logic instead). Curious for your two cents here, as well 😄

enterprise_access/apps/bffs/tests/test_response_builders.py
```diff
@@ -0,0 +1,143 @@
...
+            self.context.add_warning(
+                **self.mock_warning
+            )
+            expected_output['warnings'] = [self.mock_warning]
+        response_data = self.base_response_builder.add_errors_warnings_to_response(sample_response)
+        self.assertEqual(response_data, expected_output)
+
+    @ddt.data(
+        {
+            'status_code': 201
```
adamstankiewicz, line 61: nit: I'm not sure the API will resolve to a `201`, even if it has side effects like auto-applying a subscription license; in this case, I think the status code would still be 200. Given this, perhaps use the test to reflect an error status code like `400` instead?

nit x2: consider using `status.HTTP_400_BAD_REQUEST`

enterprise_access/apps/bffs/tests/test_response_builders.py
```diff
@@ -0,0 +1,143 @@
...
+    def test_get_status_code(self, status_code):
+        expected_output = status_code if status_code else 200
+        if status_code:
+            self.context.status_code = status_code
+        response_status_code = self.base_response_builder.get_status_code()
+        self.assertEqual(response_status_code, expected_output)
+
+
+@ddt.ddt
+class TestBaseLearnerResponseBuilder(TestBaseResponseBuilder, TestLicenseManagerUserApiClient):
```
adamstankiewicz, line 77: [curious] Why are the tests for the `BaseLearnerResponseBuilder` extending the test class for the `LicenseManagerUserApiClient`? Is it perhaps just to get access to the `self.mock_subscription_plan` used below?

Might actually have similar question for the intent in extending `TestBaseResponseBuilder` here as well.

enterprise_access/apps/bffs/response_builder.py
```diff
@@ -0,0 +1,137 @@
...
+            dict: The modified response data with common logic applied.
+        """
+        subscriptions_context = self.context.data.get('subscriptions', {})
+        enterprise_customer_user_subsidies = response_data.get('enterprise_customer_user_subsidies', {})
+        subscriptions = enterprise_customer_user_subsidies.get('subscriptions', {})
+        subscriptions.update(subscriptions_context)
+        enterprise_customer_user_subsidies.update({
+            'subscriptions': subscriptions,
+        })
+        response_data['enterprise_customer_user_subsidies'] = enterprise_customer_user_subsidies
```
adamstankiewicz, line 81: [consideration/thought question] Curious for your two cents here: I'm wondering if we should co-locate all subsidies-related metadata under `enterprise_customer_user_subsidies` within the `self.context.data`? Might co-locating subsidy types under a single dict in the handlers be worthwhile vs. having distinct properties (like `subscriptions`) at the top-level of `self.context.data`?

It _could_ make the handlers stuff more maintainable/easier to grok the context data if all subsidy types were grouped under `enterprise_customer_user_subsidies` before it gets to the response builder.

enterprise_access/apps/bffs/tests/test_response_builders.py
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+        self.assertEqual(response_data, expected_response)
+
+    def test_common_response_logic_subscription_context(self):
+        self.context.data['subscriptions'] = {
+            **self.mock_subscription_plan,
+            'uuid': self.mock_subscription_plan_uuid_1,
+            'enterprise_catalog_uuid': self.mock_enterprise_catalog_uuid_1,
+        }
```
adamstankiewicz, line 105: You mentioned this in Slack already, but the `subscriptions` data currently includes:

```python
{
  'customer_agreement': subscriptions_result.get('customer_agreement', {}),
  'subscription_licenses': transformed_licenses,
}
```

Eventually, it'll conform to further transforms similar what exists in the frontend ([source](https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/1205/files#diff-df539c9ca1128b6cb78a470cd3c3f45470ec7e0561a02df6d8f3e1f90cb58eeaR202)).

enterprise_access/apps/bffs/response_builder.py
brobro10000: Updated 👍🏽

enterprise_access/apps/api_client/tests/test_utils.py
```diff
@@ -15,3 +21,106 @@ def __init__(self, json_data, status_code):
 
     def json(self):  # pylint: disable=arguments-differ
         return self.json_data
+
+
+class TestLicenseManagerMetadata(TestCase):
```
adamstankiewicz, line 26: nit: this class isn't testing anything per se, so perhaps the name should be something more like `MockLicenseManagerMetadataMixin`?

enterprise_access/apps/api_client/tests/test_utils.py
```diff
@@ -15,3 +21,106 @@ def __init__(self, json_data, status_code):
...
+                self.mock_enterprise_catalog_uuid,
+            ],
+            "enable_auto_applied_subscriptions_with_universal_link": False,
+            "has_custom_license_expiration_messaging": False,
+            "modal_header_text": None,
+            "expired_subscription_modal_messaging": None,
+            "button_label_in_modal": None,
+            "url_for_button_in_modal": None
+        }
+        self.mock_learner_license_result_response = {
```
adamstankiewicz, line 85: nit: 
```suggestion
        self.mock_subscription_license = {
```

enterprise_access/apps/api_client/tests/test_utils.py
```diff
@@ -15,3 +21,106 @@ def __init__(self, json_data, status_code):
...
+        }
+        self.mock_learner_license_activation_response = {
+            "uuid": self.mock_learner_license_activation_uuid,
+            "status": LicenseStatuses.ACTIVATED,
+            "user_email": self.mock_user_email,
+            "activation_date": _days_from_now(-10, DATE_FORMAT_ISO_8601_MS),
+            "last_remind_date": _days_from_now(-5, DATE_FORMAT_ISO_8601_MS),
+            "subscription_plan_title": "mock_subscription_plan_title",
+            "subscription_plan_expiration_date": _days_from_now(50, DATE_FORMAT_ISO_8601),
+            "activation_link": self.faker.url()
```
adamstankiewicz, line 104: Are these fields actually returned by the `license-activation` API? I believe it uses `LicenseSerializer`, which has the following fields ([source](https://github.com/openedx/license-manager/blob/2f0c9f40969aa1994ef001f56d21e8b6a4e0a6ae/license_manager/apps/api/serializers.py#L462-L480)):

```python
fields = [
            'uuid',
            'status',
            'user_email',
            'activation_date',
            'last_remind_date',
            'subscription_plan_uuid',
            'revoked_date',
            'activation_key',
        ]
```

However, this mock attribute is including `subscription_plan_title`, `subscription_plan_expiration_date`, and `activation_link` 🤔

enterprise_access/apps/api_client/tests/test_utils.py
```diff
@@ -15,3 +21,106 @@ def __init__(self, json_data, status_code):
...
+            "status": LicenseStatuses.ACTIVATED,
+            "user_email": self.mock_user_email,
+            "activation_date": _days_from_now(-10, DATE_FORMAT_ISO_8601_MS),
+            "last_remind_date": _days_from_now(-5, DATE_FORMAT_ISO_8601_MS),
+            "subscription_plan_uuid": self.mock_subscription_plan_uuid,
+            "revoked_date": None,
+            "activation_key": self.mock_license_activation_key,
+            "subscription_plan": self.mock_subscription_plan
+        }
+        self.mock_learner_license_activation_response = {
```
adamstankiewicz, line 96: Any reason why we couldn't rely on the existing `self.mock_subscription_license` as the base metadata for a license, and override its fields as needed? Not sure if we need to have multiple attributes representing a single subscription license?

enterprise_access/apps/api_client/tests/test_utils.py
```diff
@@ -15,3 +22,101 @@ def __init__(self, json_data, status_code):
...
+        self.mock_user_email = self.mock_user.email
+
+
+class MockLicenseManagerMetadataMixin(MockEnterpriseMetadata):
+    """
+    Mixin for the TestLicenseManagerUserApiClient
+    """
+    def setUp(self):
+        super().setUp()
+        self.faker = Faker()
```
adamstankiewicz, line 48: nit: `self.faker` is added via `MockEnterpriseMetadata`

enterprise_access/apps/bffs/response_builder.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # TODO Revisit this function in ENT-9633 to determine if 200 is ok for a nested errored response
+    def get_status_code(self):
+        """
+        Gets the current status code from the context.
+
+        Returns:
+            int: The HTTP status code.
+        """
+        return self.context.status_code if hasattr(self.context, 'status_code') else status.HTTP_200_OK
```
adamstankiewicz, line 52: nit: if adding `self._status_code` (and possibly the related helper `@property` as a read-only getter) in this PR, we could implement/simplify this method as `return self.context.status_code`. Especially given the `status_code` is defaulted to `status.HTTP_200_OK` in the `HandlerContext` now.

enterprise_access/apps/bffs/serializers.py
```diff
@@ -28,3 +30,194 @@ class ErrorSerializer(BaseBFFMessageSerializer):
...
+    def to_representation(self, instance):
+        """
+        Override to_representation method to return ordered representation
+        with errors/warnings at the end of the response.
+        """
+        representation = super().to_representation(instance)
+
+        ordered_representation = OrderedDict(representation)
+
+        return ordered_representation
```
adamstankiewicz, line 58: [sanity check] Is this method needed anymore? All it's doing now is converting `representation` as an `OrderedDict`, but isn't overriding anything within it. As discussed, we'll do what this method's docstring says within the viewset layer instead.

enterprise_access/apps/bffs/tests/test_response_builders.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    @mock.patch('enterprise_access.apps.bffs.context.HandlerContext')
+    @ddt.unpack
+    def test_get_status_code(self, mock_handler_context, status_code):
+        mock_handler_context.return_value = self.mock_handler_context
+        mock_handler_context_instance = mock_handler_context.return_value
+        base_response_builder = BaseResponseBuilder(mock_handler_context_instance)
+        expected_output = status_code if status_code else 200
+        if status_code:
+            mock_handler_context_instance.status_code = status_code
```
adamstankiewicz, line 84: nit/inform: just a heads up, there is not currently a `status_code` property in `HandlerContext` as only the private `_status_code` property is exposed.

enterprise_access/apps/bffs/tests/test_response_builders.py
```diff
@@ -0,0 +1,140 @@
...
+        self.mock_enterprise_catalog_uuid_2 = self.faker.uuid4()
+
+        self.mock_subscription_plan_uuid_1 = self.faker.uuid4()
+        self.mock_subscription_plan_uuid_2 = self.faker.uuid4()
+
+        self.mock_customer_agreement_uuid_1 = self.faker.uuid4()
+        self.mock_customer_agreement_uuid_2 = self.faker.uuid4()
+
+    @mock.patch('enterprise_access.apps.bffs.context.HandlerContext')
+    def test_common_response_logic_initialization(self, mock_handler_context):
```
adamstankiewicz, line 103: Is this test case needed, given `common_response_logic` is covered via `test_build` below?

brobro10000: I would still say yes. It test's the `common_response_logic` initialization of the data-structure for the response based on the serializer and an empty request from the handler. Its a very explicit unit-test for initialization.

adamstankiewicz: [curious] What if there was something wrong within the `build` method in how it handles the "empty state" subsidies data? By only testing `common_response_logic` independently in this case, we don't necessarily have coverage on the `add_errors_warnings_to_response` called within `build` for the init case you're describing.

Any reason not to treat it as a single test method `test_build`, but with differing test inputs for the init vs. actual data case with `@ddt.data`?

For example, something like:

```python
@ddt.data(
    {'has_subscriptions_data': True},
    {'has_subscriptions_data': False}
)
@mock.patch('enterprise_access.apps.bffs.context.HandlerContext')
@ddt.unpack
def test_build(self, mock_handler_context, has_subscriptions_data):
    mock_handler_context.return_value = self.mock_handler_context
    mock_handler_context_instance = mock_handler_context.return_value

    base_learner_response_builder = BaseLearnerResponseBuilder(mock_handler_context_instance)
    mock_subscriptions_data = {
        "customer_agreement": None,
        "subscription_licenses": [],
        "subscription_licenses_by_status": {
            'activated': [],
            'assigned': [],
            'expired': [],
            'revoked': [],
        },
    }

    if has_subscriptions_data:
        mock_subscriptions_data.update({
            "customer_agreement": self.mock_customer_agreement,
            "subscription_licenses": [self.mock_subscription_license],
            "subscription_licenses_by_status": {
                **mock_subscriptions_data['subscription_licenses_by_status'],
                'activated': [self.mock_subscription_license],
            },
        })

    mock_handler_context_instance.data['enterprise_customer_user_subsidies'] = {
        'subscriptions': mock_subscriptions_data,
    }

    expected_response = {
        'enterprise_customer_user_subsidies': {
            'subscriptions': mock_subscriptions_data,
        },
        'errors': [],
        'warnings': [],
    }

    response_data, status_code = base_learner_response_builder.build()

    self.assertEqual(response_data, expected_response)
    self.assertEqual(status_code, status.HTTP_200_OK)
```

brobro10000: Yep this would absolutely cover both instances. Removing the individualized test case 👍🏽

enterprise_access/apps/api_client/tests/test_utils.py
```diff
@@ -15,3 +22,99 @@ def __init__(self, json_data, status_code):
...
+            "subscription_plan_uuid": self.mock_subscription_plan_uuid,
+            "revoked_date": None,
+            "activation_key": self.mock_license_activation_key,
+            "subscription_plan": self.mock_subscription_plan
+        }
+        self.mock_learner_license_activation_response = {
+            **self.mock_subscription_license,
+            "uuid": self.mock_learner_license_activation_uuid,
+            "status": LicenseStatuses.ACTIVATED,
+            "user_email": self.mock_user_email,
```
adamstankiewicz, line 112: nit: `user_email` is the same as `self.mock_subscription_license`, and is already spread via line 109 above.

nit x2: technically the same for `activation_date`, `last_remind_date`, and `status` too. It seems the only thing really different is the `uuid`?

enterprise_access/apps/bffs/tests/utils.py
```diff
@@ -34,6 +37,20 @@ def setUp(self):
...
+        self.mock_handler_context = mock.MagicMock(
+            _request=self.request,
+            _status_code=status.HTTP_200_OK,
+            _errors=[],
+            _warnings=[],
+            _enterprise_customer_uuid=self.mock_enterprise_customer_uuid,
+            _enterprise_customer_slug=self.mock_enterprise_customer_slug,
+            _lms_user_id=self.request.user.lms_user_id,
+            _enterprise_features={},
+            _data={},
```
adamstankiewicz, line 50: `data` is not a private attribute, i.e. `HandlerContext` uses `data` vs. `_data`. Rationale: `data` is still exposed for handlers (e.g., `DashboardHandler` to mutate freely, whereas the other private attributes are read-only).

enterprise_access/apps/bffs/tests/utils.py
```diff
@@ -34,6 +37,20 @@ def setUp(self):
...
+        )
+        self.override_handler_context_getter_properties()
```
adamstankiewicz, line 52: [suggestion] It may make sense to have the helper method be responsible for creating the `mock.MagicMock`, too? E.g., maybe something like this:

```python
def get_mock_handler_context(self, **kwargs):
    """
    Returns a mock handler context with optional overrides.
    Any attribute in the context can be overridden by passing it as a keyword argument.
    """
    # Create default values for the mock HandlerContext
    default_values = {
        '_request': self.request,
        '_status_code': status.HTTP_200_OK,
        '_errors': [],
        '_warnings': [],
        '_enterprise_customer_uuid': self.mock_enterprise_customer_uuid,
        '_enterprise_customer_slug': self.mock_enterprise_customer_slug,
        '_lms_user_id': self.request.user.lms_user_id,
        '_enterprise_features': {'feature_a': True},
        'data': {},
    }

    # Update default values with any overrides provided via kwargs
    default_values.update(kwargs)

    # Create the mock HandlerContext
    mock_handler_context = mock.MagicMock(**default_values)

    # Define a dictionary of private attributes to property names
    property_mocks = {
        'request': getattr(mock_handler_context, '_request'),
        'status_code': getattr(mock_handler_context, '_status_code'),
        'errors': getattr(mock_handler_context, '_errors'),
        'warnings': getattr(mock_handler_context, '_warnings'),
        'enterprise_customer_uuid': getattr(mock_handler_context, '_enterprise_customer_uuid'),
        'enterprise_customer_slug': getattr(mock_handler_context, '_enterprise_customer_slug'),
        'lms_user_id': getattr(mock_handler_context, '_lms_user_id'),
        'enterprise_features': getattr(mock_handler_context, '_enterprise_features'),
    }

    # Override the property getters
    for property_name, private_attr in property_mocks.items():
        setattr(
            type(mock_handler_context),  # The class of the mock object
            property_name,
            mock.PropertyMock(return_value=private_attr)
        )

    return mock_handler_context
```

This change would help simplify/negate the need to call `override_handler_context_getter_properties` within `test_response_builders` after overriding `_status_code` on the mocked `HandlerContext` instance.

For example:

```python
@ddt.data(
    {'status_code': status.HTTP_400_BAD_REQUEST},
    {'status_code': None}
)
@mock.patch('enterprise_access.apps.bffs.context.HandlerContext')
@ddt.unpack
def test_get_status_code(self, mock_handler_context, status_code):
    if status_code:
        mock_handler_context.return_value = self.get_mock_handler_context(
            _status_code=status_code
        )
    else:
        mock_handler_context.return_value = self.mock_handler_context

    mock_handler_context_instance = mock_handler_context.return_value
    expected_output = status_code if status_code else status.HTTP_200_OK

    base_response_builder = BaseResponseBuilder(mock_handler_context_instance)
    response_status_code = base_response_builder.get_status_code()

    self.assertEqual(response_status_code, expected_output)
```

adamstankiewicz (approved):
LGTM, with a couple additional test nits/suggestions.

PR #596, opened 2024-11-22
fix: check for active, current license

iloveagent57:
1. Fix conditional logic.
2. Change check for an activated license to a check for an activate, current license.

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/bffs/handlers.py
+++ b/enterprise_access/apps/bffs/handlers.py
@@ -229,14 +229,19 @@ class BaseLearnerPortalHandler(BaseHandler, BaseLearnerDataMixin):
             'subscription_licenses_by_status': subscription_licenses_by_status,
         }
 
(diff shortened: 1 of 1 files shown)


pwnage101 (approved):
LGTM!

enterprise_access/apps/bffs/handlers.py
```diff
@@ -216,14 +216,19 @@ def transform_subscriptions_result(self, subscriptions_result):
             'subscription_licenses_by_status': subscription_licenses_by_status,
         }
 
-    def check_has_activated_license(self):
+    @property
+    def current_active_license(self):
```
adamstankiewicz, line 220: Good call to move this to a `@property` :) Been thinking it might make sense to do that.

adamstankiewicz (approved):

PR #601, opened 2024-11-25
Aed/default enrollments bff

iloveagent57:
Realizes default enrollment intentions.

**Jira:**
ENT-9632

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/bffs/mixins.py
```diff
@@ -23,6 +23,10 @@ class EnterpriseCustomerUserSubsidiesDataMixin(BFFContextDataMixin):
     Mixin to access enterprise customer user subsidies data from the context.
     """
 
+    @property
+    def enterprise_customer_uuid(self):
+        return self.context.enterprise_customer.get('uuid')
```
adamstankiewicz, line 28: [nit/inform] There is a helper property `self.context.enterprise_customer_uuid` to make accessing the enterprise UUID more straightforward.

adamstankiewicz: [nit/consideration] Given `enterprise_customer_uuid` is available for all handler classes, does it make sense to only add this property within `EnterpriseCustomerUserSubsidiesDataMixin` vs. putting it somewhere more commonly accessible across all handlers, if it's needed (e.g., would it also potentially make sense to always grab the enterprise uuid from `self.context` throughout the handlers so its more explicit where it's coming from)?

enterprise_access/apps/bffs/serializers.py
```diff
@@ -261,7 +263,7 @@ class EnterpriseCourseEnrollmentSerializer(BaseBffSerializer):
     org_name = serializers.CharField()
     course_run_status = serializers.CharField()
     display_name = serializers.CharField()
-    emails_enabled = serializers.BooleanField()
+    emails_enabled = serializers.BooleanField(required=False, allow_null=True)
```
adamstankiewicz, line 266: Good catch that this is valid to be nullable! ([source](https://github.com/openedx/edx-platform/blob/139b4167b37b49d2d69cccdbd19d8ccef40d3374/lms/djangoapps/bulk_email/api.py#L45))

enterprise_access/apps/bffs/handlers.py
```diff
@@ -229,7 +232,7 @@ def transform_subscriptions_result(self, subscriptions_result):
             'subscription_licenses_by_status': subscription_licenses_by_status,
         }
 
-    @property
+    @cached_property
```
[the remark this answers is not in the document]
iloveagent57: yes, `@cached_property` caches only on the property instance, so this property would only be cached for however long a particular `Handler()` object exists. And we'll have one new handler instance created in each request.

enterprise_access/apps/bffs/handlers.py
```diff
@@ -68,6 +70,7 @@ def __init__(self, context):
         # API Clients
         self.license_manager_client = LicenseManagerUserApiClient(self.context.request)
         self.lms_user_api_client = LmsUserApiClient(self.context.request)
+        self.lms_user_id = self.context.request.user.lms_user_id
```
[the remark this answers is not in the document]
iloveagent57: Ah, I missed that when I was looking around for something like this. I'll switch to `self.context.lms_user_id`.

enterprise_access/apps/bffs/mixins.py
iloveagent57: Ah same here, I'll switch to self.context.enterprise_customer_uuid.

enterprise_access/apps/bffs/handlers.py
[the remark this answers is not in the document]
iloveagent57: Oh wait, a unit test just exposed that I can't actually do this. There's a point where we read it before licenses are loaded, and then the value gets cached as null, but we should have a non-null value later. I'll change back to just a property.

PR #604, opened 2024-12-02
fix: unittests for default realizations, remove errant break

iloveagent57:
**Description:**
* Remove an errant `break` from a loop in `enroll_in_redeemable_default_enterprise_enrollment_intentions()`
* Light refactoring in same function.
* Adds unittests for same function.

**Jira:**
ENT-9632

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


adamstankiewicz (approved):
LGTM!

Though, just a quick sanity check - are we good with only testing the auto-enrollment side effects within the `test_handlers` unit tests versus in the viewset (where we currently have more integration-style tests for the subscription license side effects of license activation and auto-applied licenses)?

I personally don't have strong feelings one way or another; what you have here is totally fine! :)

PR #605, opened 2024-12-02
fix: account for revoked licenses and staff users in auto-apply logic

adamstankiewicz:
**Description:**

Extends auto-apply subscription license business logic and tests to include:
* Check for revoked subscription licenses (i.e., no auto-apply with revoked license).
* Check for being explicitly linked to the requested enterprise customer (i.e., no auto-apply as an unlinked staff user).

These additional checks ensure parity with existing business logic around auto-applied subscription licenses in the frontend ([source](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/73fc10e94aa64cb9103e8131d3fa9d4031d0c410/src/components/app/data/services/subsidies/subscriptions.js#L184)).

**Jira:**
[ENT-9813](https://2u-internal.atlassian.net/browse/ENT-9813)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_bff_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_bff_views.py
@@ -448,7 +448,7 @@ class TestLearnerPortalBFFViewSet(TestHandlerContextMixin, MockLicenseManagerMet
             'has_plan_for_auto_apply': True,
             'should_auto_apply': True,
         },
-        # Existing activated license, identity provider exists, no universal link auto-apply,
+        # Existing activated license, identity provider exists, universal link auto-apply is enabled,
         # a plan for auto-apply exists, and not a staff request user.
         # Expected: Should not auto-apply.
         {
@@ -460,7 +460,7 @@ class TestLearnerPortalBFFViewSet(TestHandlerContextMixin, MockLicenseManagerMet
             'has_plan_for_auto_apply': True,
             'should_auto_apply': False,
         },
-        # Existing revoked license, identity provider exists, no universal link auto-apply,
+        # Existing revoked license, identity provider exists, universal link auto-apply is enabled,
         # a plan for auto-apply exists, and not a staff request user.
         # Expected: Should not auto-apply.
         {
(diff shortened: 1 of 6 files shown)


brobro10000 (approved):

enterprise_access/apps/api/v1/tests/test_bff_views.py
```diff
@@ -352,73 +352,142 @@ def test_dashboard_with_subscriptions_license_activation(
...
+        {
+            'has_existing_activated_license': True,
+            'has_existing_revoked_license': False,
+            'is_staff_request_user': False,
+            'identity_provider': True,
+            'auto_apply_with_universal_link': True,
+            'has_plan_for_auto_apply': True,
+            'should_auto_apply': False,
+        },
+        # Existing revoked license, identity provider exists, no universal link auto-apply,
```
iloveagent57, line 463: these say "no universal link auto-apply", but the data for both has `'auto_apply_with_universal_link': True,` - is there a test permutation we're missing? or maybe just copypasta in the comment?

enterprise_access/apps/bffs/handlers.py
```diff
@@ -363,13 +390,20 @@ def check_and_auto_apply_license(self):
...
         """
-        subscription_licenses_by_status = self.subscription_licenses_by_status
-        assigned_licenses = subscription_licenses_by_status.get('assigned', [])
-
-        if assigned_licenses or self.current_active_license:
-            # Skip auto-apply if user already has assigned license(s) or an already-activated license
+        if (
+            self.current_assigned_licenses or
+            self.current_activated_licenses or
+            self.current_revoked_licenses or
```
iloveagent57, line 396: Maybe we can add a helper to produce a list of all current licenses, regardless of status, and then use that here?

iloveagent57 (approved):
Couple small suggestions, looks good otherwise!

enterprise_access/apps/api/v1/tests/test_bff_views.py
adamstankiewicz: D'oh, just copy/pasta error. Will fix!

enterprise_access/apps/bffs/handlers.py
adamstankiewicz: Ah, good suggestion given all 3 allocatable license statuses are considered now; we can use the existing `self.subscription_licenses` 😄

PR #612, opened 2024-12-12
fix: don't cache if default intentions need enrollment

iloveagent57:
**Description:**
We don't want to cache default intentions data if it includes any enrollable intentions, because we don't want to re-attempt enrollment realization on the second of consecutive requests.

**Jira:**
ENT-9862

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/bffs/api.py
+++ b/enterprise_access/apps/bffs/api.py
@@ -125,6 +125,8 @@ def get_and_cache_default_enterprise_enrollment_intentions_learner_status(
 ):
     """
     Retrieves and caches default enterprise enrollment intentions for a learner.
+    This function does not cache this data if it includes any enrollable intentions,
+    because we don't want to re-attempt enrollment realization on the second of consecutive requests.
     """
     cache_key = default_enterprise_enrollment_intentions_learner_status_cache_key(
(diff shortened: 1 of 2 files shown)


macdiesel (approved):

pwnage101 (approved):

enterprise_access/apps/bffs/api.py
```diff
@@ -142,6 +144,13 @@ def get_and_cache_default_enterprise_enrollment_intentions_learner_status(
     response_payload = client.get_default_enterprise_enrollment_intentions_learner_status(
         enterprise_customer_uuid=enterprise_customer_uuid,
     )
+
+    # Don't set in the cache if there are enrollable intentions
+    if statuses := response_payload.get('enrollment_statuses', {}):
+        needs_enrollment = statuses.get('needs_enrollment', {})
+        if needs_enrollment.get('enrollment', []):
```
pwnage101, line 151: this shouldn't say "enrollable", should it?
```suggestion
        if needs_enrollment.get('enrollable', []):
```

iloveagent57: ah, yes, good call

PR #624, opened 2025-01-23
feat: incorporate enterprise customer logic into BFF

brobro10000:
Serializes fields in the BFF context to always return `all_linked_enterprise_customer_users` and `should_update_active_enterprise_customer_user`. The intention is to eventually remove `should_update_active_enterprise_customer_user` and take advantage of a net new API call to select an active enterprise customer. This work will be completed as part of a seperate ticket. 

It modifies the enterprise customer user object to update the metadata to the set active, and invalidate the enterprise customer user query cache. We return `should_update_active_enterprise_customer_user` back to the user to set the active enterprise customer in the frontend. We invalidate the query to ensure the next call to the BFF would return the now active enterprise customer without transformations on the enterprise customer user object in the BFF layer.

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/bffs/api.py
+++ b/enterprise_access/apps/bffs/api.py
@@ -8,7 +8,7 @@ from edx_django_utils.cache import TieredCache
 
 from enterprise_access.apps.api_client.license_manager_client import LicenseManagerUserApiClient
 from enterprise_access.apps.api_client.lms_client import LmsApiClient, LmsUserApiClient
-from enterprise_access.cache_utils import versioned_cache_key
+from enterprise_access.cache_utils import request_cache, versioned_cache_key
 
 logger = logging.getLogger(__name__)
 
@@ -41,13 +41,13 @@ def enterprise_course_enrollments_cache_key(enterprise_customer_uuid, lms_user_i
     return versioned_cache_key('get_enterprise_course_enrollments', enterprise_customer_uuid, lms_user_id)
 
 
-def get_and_cache_enterprise_customer_users(request, timeout=settings.ENTERPRISE_USER_RECORD_CACHE_TIMEOUT, **kwargs):
+def get_and_cache_enterprise_customer_users(request, **kwargs):
     """
     Retrieves and caches enterprise learner data.
     """
     username = request.user.username
     cache_key = enterprise_customer_users_cache_key(username)
-    cached_response = TieredCache.get_cached_response(cache_key)
+    cached_response = request_cache(namespace=REQUEST_CACHE_NAMESPACE).get_cached_response(cache_key)
     if cached_response.is_found:
         logger.info(
             f'enterprise_customer_users cache hit for username {username}'
@@ -59,7 +59,11 @@ def get_and_cache_enterprise_customer_users(request, timeout=settings.ENTERPRISE
         username=username,
         **kwargs,
     )
-    TieredCache.set_all_tiers(cache_key, response_payload, timeout)
+    logger.info(
+        'Fetched enterprise customer user for username %s',
+        username,
+    )
+    request_cache(namespace=REQUEST_CACHE_NAMESPACE).set(cache_key, response_payload)
     return response_payload
 
 
@@ -258,30 +262,33 @@ def _determine_enterprise_customer_for_display(
     Determine the enterprise customer user for display.
 
     Returns:
-        The enterprise customer user for display.
+        tuple(Dict, boolean): The enterprise customer user for display, and a boolean to determine
+        whether to update the active enterprise customer to the return value.
     """
+
     if not enterprise_customer_slug and not enterprise_customer_uuid:
         # No enterprise customer specified in the request, so return the active enterprise customer
-        return active_enterprise_customer
+        return active_enterprise_customer, False
 
     # If the requested enterprise does not match the active enterprise customer user's slug/uuid
     # and there is a linked enterprise customer user for the requested enterprise, return the
-    # linked enterprise customer.
+    # linked enterprise customer. By returning true, we are updating the current active enterprise
+    # customer to the requested_enterprise_customer
     request_matches_active_enterprise_customer = _request_matches_active_enterprise_customer(
         active_enterprise_customer=active_enterprise_customer,
         enterprise_customer_slug=enterprise_customer_slug,
         enterprise_customer_uuid=enterprise_customer_uuid,
     )
     if not request_matches_active_enterprise_customer and requested_enterprise_customer:
-        return requested_enterprise_customer
+        return requested_enterprise_customer, True
 
     # If the request user is staff and the requested enterprise does not match the active enterprise
     # customer user's slug/uuid, return the staff-enterprise customer.
     if staff_enterprise_customer:
-        return staff_enterprise_customer
+        return staff_enterprise_customer, False
 
     # Otherwise, return the active enterprise customer.
-    return active_enterprise_customer
+    return active_enterprise_customer, False
 
 
 def _request_matches_active_enterprise_customer(
@@ -360,7 +367,7 @@ def transform_enterprise_customer_users_data(data, request, enterprise_customer_
         enterprise_customer_user_for_requested_customer.get('enterprise_customer')
         if enterprise_customer_user_for_requested_customer else None
     )
-    enterprise_customer = _determine_enterprise_customer_for_display(
+    enterprise_customer, should_update_active_enterprise_customer_user = _determine_enterprise_customer_for_display(
         enterprise_customer_slug=enterprise_customer_slug,
         enterprise_customer_uuid=enterprise_customer_uuid,
         active_enterprise_customer=active_enterprise_customer,
@@ -373,4 +380,5 @@ def transform_enterprise_customer_users_data(data, request, enterprise_customer_
         'active_enterprise_customer': active_enterprise_customer,
         'all_linked_enterprise_customer_users': enterprise_customer_users,
         'staff_enterprise_customer': staff_enterprise_customer,
+        'should_update_active_enterprise_customer_user': should_update_active_enterprise_customer_user,
     }
--- a/enterprise_access/apps/bffs/context.py
+++ b/enterprise_access/apps/bffs/context.py
@@ -103,8 +103,16 @@ class HandlerContext:
         return self.data.get('active_enterprise_customer')
 
     @property
+    def staff_enterprise_customer(self):
+        return self.data.get('staff_enterprise_customer')
+
+    @property
     def all_linked_enterprise_customer_users(self):
         return self.data.get('all_linked_enterprise_customer_users')
+
+    @property
+    def should_update_active_enterprise_customer_user(self):
+        return self.data.get('should_update_active_enterprise_customer_user')
 
     @property
     def is_request_user_linked_to_enterprise_customer(self):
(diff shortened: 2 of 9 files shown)


enterprise_access/apps/bffs/api.py
```diff
@@ -260,9 +274,10 @@ def _determine_enterprise_customer_for_display(
     Returns:
         The enterprise customer user for display.
```
adamstankiewicz, line 275: nit: this method is returning a tuple now. docstring can be updated.

enterprise_access/apps/bffs/api.py
```diff
@@ -260,9 +274,10 @@ def _determine_enterprise_customer_for_display(
...
     """
+
     if not enterprise_customer_slug and not enterprise_customer_uuid:
         # No enterprise customer specified in the request, so return the active enterprise customer
-        return active_enterprise_customer
+        return active_enterprise_customer, False
 
     # If the requested enterprise does not match the active enterprise customer user's slug/uuid
     # and there is a linked enterprise customer user for the requested enterprise, return the
```
adamstankiewicz, line 283: nit: this docstring could be updated to indicate updating the `active` ECU, per the `True` in the returned tuple

enterprise_access/apps/bffs/api.py
```diff
@@ -360,7 +375,7 @@ def transform_enterprise_customer_users_data(data, request, enterprise_customer_
         enterprise_customer_user_for_requested_customer.get('enterprise_customer')
         if enterprise_customer_user_for_requested_customer else None
     )
-    enterprise_customer = _determine_enterprise_customer_for_display(
+    enterprise_customer, should_update_active_enterprise_customer = _determine_enterprise_customer_for_display(
```
adamstankiewicz, line 378: nit: consider calling this `should_update_active_enterprise_customer_user`

enterprise_access/apps/bffs/context.py
```diff
@@ -222,13 +227,41 @@ def _initialize_enterprise_customer_users(self):
                 self.enterprise_customer_uuid,
                 self.enterprise_customer_slug,
             )
-
+        # sets the active enterprise customer in the LMS
```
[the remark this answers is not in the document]
brobro10000: I believe this is resolved by triggering the invalidation on the cache if we are updating the active enterprise customer. 
If the customers initial point of entry is on a non-bff enabled route and the requested enterprise customer (slug) is not the same as the users current active enterprise customer, and they traverse to a bff enabled route, the request value's slug should be the same one we retrieve for the enterprise customer user. We would have already updated the active customer from the non-bff enabled route, so the active enterprise customer would be the same as the enterprise customer user retrieved from the slug.

If the customer's initial point of entry is on a bff enabled route, route and the requested enterprise customer (slug) is not the same as the users current active enterprise customer, we would optimistically update the return value from the enterprise customer user (which would be set in the frontend query cache), set the active enterprise customer to the requested enterprise customer, and invalidate the cache. The returned `should_update_active_enterprise_customer_user` field would update the active enterprise customer to the requested enterprise customer in the frontend. The next call on a bff enabled route would result in a new call and updated query cache with the requested enterprise customer.

If something in my logic seems off lmk but I believe the invalidation of the cache handles it. 👍🏽

enterprise_access/apps/bffs/serializers.py
```diff
@@ -152,12 +152,43 @@ class EnterpriseCustomerSerializer(BaseBffSerializer):
...
+class AllLinkedEnterpriseCustomerUsersSerializer(BaseBffSerializer):
+    """
+    Serializer for all linked enterprise customer users
+    """
+
+    id = serializers.IntegerField()
+    enterprise_customer = EnterpriseCustomerSerializer()
+    active = serializers.BooleanField()
+    user_id = serializers.IntegerField()
+    user = UserSerializer()
```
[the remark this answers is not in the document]
brobro10000: Great callout, for the most part we don't use any of hte fields related to all linked enterprise customer so this could be paired down to just enterpriseCustomer 👍🏽

enterprise_access/apps/bffs/serializers.py
```diff
@@ -152,12 +152,27 @@ class EnterpriseCustomerSerializer(BaseBffSerializer):
     show_integration_warning = serializers.BooleanField()
 
 
+class EnterpriseCustomerUsersSerializer(BaseBffSerializer):
```
adamstankiewicz, line 155: This represent a _single_ `EnterpriseCustomerUser` record, yeah? If so, should this be named as `EnterpriseCustomerUserSerializer`?

enterprise_access/apps/bffs/serializers.py
```diff
@@ -152,12 +152,27 @@ class EnterpriseCustomerSerializer(BaseBffSerializer):
...
 class BaseResponseSerializer(BaseBffSerializer):
     """
     Serializer for base response.
     """
 
     enterprise_customer = EnterpriseCustomerSerializer(required=False, allow_null=True)
+    all_linked_enterprise_customer_users = serializers.ListField(
+        child=EnterpriseCustomerUsersSerializer(),
+        allow_empty=True,
+    )
```
adamstankiewicz, line 174: nit: as a sanity check, the following doesn't work?

```python
all_linked_enterprise_customer_users = EnterpriseCustomerUsersSerializer(many=True, allow_empty=True, default=list)
```

(FWIW, using the `many=True` approach seems to pass all tests locally still).

enterprise_access/apps/bffs/context.py
```diff
@@ -222,13 +227,43 @@ def _initialize_enterprise_customer_users(self):
...
+                    'all_linked_enterprise_customer_users': [
+                        {
+                            **enterprise_customer_user,
+                            "active": (
+                                enterprise_customer_user["enterprise_customer"]["uuid"] == enterprise_customer_uuid
+                            )
+                        }
+                        for enterprise_customer_user in transformed_data.get('all_linked_enterprise_customer_users', [])
+                    ]
+                })
```
adamstankiewicz, line 249: [clarification] If the BFF is not yet calling the POST `/select/active` API to mutate the active ECU, is it appropriate for the BFF to do in-memory data transforms related to the active enterprise customer, besides returning the added `should_update_active_enterprise_customer_user` to tell the frontend when to call `/select/active`?

For example, at the time of this request when `should_update_active_enterprise_customer_user` is truthy, the active ECU has not yet been changed; it takes the frontend to make the POST (with an optimistic query cache update on the frontend), and then when the BFF API is next refetched, it should reflect the changed active ECU. Despite the active ECU not changing at the time of the BFF request, the `transformed_data` is largely be mutated as if it were.

Rather than calling `invalidate_enterprise_customer_users` and subsequently doing data transforms a bit prematurely, given the frontend is still going to be calling `/select/active` for the time being, I wonder if we should effectively change the cache timeout for the `enterprise-learner` API call within the BFF layer to `0` (aka effectively removing persisted `TieredCache`).

However, because the `enterprise-learner` is potentially called twice (i.e., once during the `@permission_required` check and again during the handler data loading), we should still have _some_ caching on it to avoid duplicative requests.

There might be an opportunity to change the caching the `enterprise-learner` API request to use `RequestCache` instead of `TieredCache` so the cached response is only valid for the duration of the request (i.e., response not persisted in cache across multiple requests).

That way, BFF is responsible for only telling the Learner Portal when to call `/select/active` per `should_update_active_enterprise_customer_user` and the next request to the BFF API will re-fetch `enterprise-learner` to get the updated active ECU in its response.

Let me know if you want to 🦆 about this approach further.

brobro10000: Request cache implemented 👍🏽  and the learner portal behavior is what we expect to see for setting an active customer and correctly returning the enterprise customer we expect

enterprise_access/apps/bffs/serializers.py
```diff
@@ -152,12 +152,24 @@ class EnterpriseCustomerSerializer(BaseBffSerializer):
...
+
+
 class BaseResponseSerializer(BaseBffSerializer):
     """
     Serializer for base response.
     """
 
     enterprise_customer = EnterpriseCustomerSerializer(required=False, allow_null=True)
+    all_linked_enterprise_customer_users = EnterpriseCustomerUserSerializer(many=True, allow_empty=True, default=list)
+    should_update_active_enterprise_customer_user = serializers.BooleanField(default=False)
```
adamstankiewicz, line 172: FWIW, I do think `serializers.BooleanField` does default to `False` on its own, where passing `default=False` isn't needed ([source](https://www.geeksforgeeks.org/boolean-fields-in-serializers-django-rest-framework/)).

image

All tests pass without the `default=False`.

enterprise_access/apps/bffs/context.py
```diff
@@ -222,13 +226,42 @@ def _initialize_enterprise_customer_users(self):
                 self.enterprise_customer_uuid,
                 self.enterprise_customer_slug,
             )
-
+        # Sets the active enterprise customer metadata in the LMS
+        if should_update_active_enterprise_customer_user := transformed_data.get(
```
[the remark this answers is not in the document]
adamstankiewicz: [follow-up] Ah, I'm understanding the current state of how this interacts with the corresponding learner portal [PR](https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/1263).

```tsx
// The logic to ensure active enterprise customer user is done in the BFF, but updating the
// active enterprise customer is still required
```

While these 2 pieces (the BFF API + learner portal) are intended to play nicely together, they should be developed as independent applications with minimal coupling. That is, when relying on frontend mutation via `select/active` to update the active ECU, it's a bit odd to assume the backend API return the expected mutated data prior to actually mutating the data. Consider if the API is called in isolation via Postman, where the `/select/active` is not yet called (since it's not called with Learner Portal).

If we mutate the in-memory data here in the Postman-only use case, the data is technically misleading as the `/select/active` API was not yet called.

enterprise_access/apps/bffs/api.py
```diff
@@ -41,13 +41,13 @@ def enterprise_course_enrollments_cache_key(enterprise_customer_uuid, lms_user_i
     return versioned_cache_key('get_enterprise_course_enrollments', enterprise_customer_uuid, lms_user_id)
 
 
-def get_and_cache_enterprise_customer_users(request, timeout=settings.ENTERPRISE_USER_RECORD_CACHE_TIMEOUT, **kwargs):
```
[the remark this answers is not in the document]
brobro10000: There are still a couple of instances of its usages:
[get_and_cache_enterprise_customer](https://github.com/openedx/enterprise-access/blob/545c22e2ca52ed63ecb854e1cbe1af77b31f4825/enterprise_access/apps/bffs/api.py#L69)
[get_and_cache_enterprise_learner_record](https://github.com/openedx/enterprise-access/blob/545c22e2ca52ed63ecb854e1cbe1af77b31f4825/enterprise_access/apps/subsidy_access_policy/customer_api.py#L20)

enterprise_access/apps/bffs/serializers.py
```diff
@@ -152,12 +152,26 @@ class EnterpriseCustomerSerializer(BaseBffSerializer):
     show_integration_warning = serializers.BooleanField()
 
 
+class EnterpriseCustomerUserSerializer(BaseBffSerializer):
+    """
+    Serializer for all linked enterprise customer users
+    """
+    id = serializers.IntegerField()
+    user_id = serializers.IntegerField()
```
adamstankiewicz, line 160: [very optional nit] could probably nix `user_id` from the response, as I imagine its value will be the same across all `EnterpriseCustomerUsers` in the list and if the frontend needed a user id for the authenticated user, it'd likely get it from `useContext(AppContext).authenticatedUser.userId` vs. this API response.

adamstankiewicz (approved):
LGTM, with a non-blocking nit. Nice job!

PR #628, opened 2025-01-30
fix: transform and group enrollments by status

adamstankiewicz:
**Description:**

Corresponding change in `frontend-app-learner-portal-enterprise`: https://github.com/openedx/frontend-app-learner-portal-enterprise/pull/1293 (**this PR to merge first**)

This PR brings some data transforms from the frontend (frontend-app-learner-portal-enterprise) to the BFF / API Gateway layer, to enable eventually removing business logic from the frontend long-term.

The overall strategy here is to:

1. Expose the additional fields resulting from the transform in the BFF API response (additive-only).
2. Remove the frontend data transforms on the enrollments data served by the BFF API response, relying solely on the BFF transformations.
3. Remove the no-longer consumed (deprecated) fields related to enrollments from the BFF API response.

Other changes:

* Creates `LearnerDashboardDataMixin` to house methods related to retrieving/transforming enterprise course enrollments related metadata.
* Implements and serializes an `all_enrollments_by_status` for the Learner Dashboard BFF API response.
  * Currently contains realized enterprise course enrollments returned by the `enterprise_course_enrollments` API.
  * In the future, it likely will also contain additional lists of requested and assigned enrollments.

**Jira:**
[ENT-9631](https://2u-internal.atlassian.net/browse/ENT-9631)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/bffs/constants.py
+++ b/enterprise_access/apps/bffs/constants.py
@@ -1,0 +1,25 @@
+"""
+Constants module for the BFFs.
+"""
+
+
+class COURSE_ENROLLMENT_STATUSES:
+    """
+    Course enrollment statuses.
+    """
+    IN_PROGRESS = 'in_progress'
+    UPCOMING = 'upcoming'
+    COMPLETED = 'completed'
+    SAVED_FOR_LATER = 'saved_for_later'
+    # Not a realized enrollment status, but used for the purpose of requesting enrollment
+    REQUESTED = 'requested'
+    # Not a realized enrollment status, but used for the purpose of assigned enrollment
+    ASSIGNED = 'assigned'
+
+
+UNENROLLABLE_COURSE_STATUSES = {
+    COURSE_ENROLLMENT_STATUSES.IN_PROGRESS,
+    COURSE_ENROLLMENT_STATUSES.UPCOMING,
+    COURSE_ENROLLMENT_STATUSES.COMPLETED,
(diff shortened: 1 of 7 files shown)


enterprise_access/apps/bffs/constants.py
```diff
@@ -0,0 +1,25 @@
...
+    Course enrollment statuses.
+    """
+    IN_PROGRESS = 'in_progress'
+    UPCOMING = 'upcoming'
+    COMPLETED = 'completed'
+    SAVED_FOR_LATER = 'saved_for_later'
+    # Not a realized enrollment status, but used for the purpose of requesting enrollment
+    REQUESTED = 'requested'
+    # Not a realized enrollment status, but used for the purpose of assigned enrollment
+    ASSIGNED = 'assigned'
```
iloveagent57, line 17: Are we using these two anywhere?

adamstankiewicz: Currently, no. Carried over from [here](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/83e555a6909a0f0ce05023f6cf12994101ab0553/src/constants/course.js#L2-L11) in `frontend-app-learner-portal-enterprise`.

Eventually, the newly serialized `all_enrollments_by_status` would include `REQUESTED` and `ASSIGNED` to fully map to the corresponding data structure in the frontend. Opted to include the constants now, with the assumption they would be utilized down the road.

PR #632, opened 2025-02-06
temp: add logs for debugging Braze emails campaign

hamzawaleed01:
**Description:**

**Jira:**
ENT-9890

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/tasks.py
+++ b/enterprise_access/apps/content_assignments/tasks.py
@@ -219,7 +219,7 @@ class BrazeCampaignSender:
         end_date = self.normalized_metadata.get('end_date')
         course_metadata = self.course_metadata
         logger.info(
-            f"[get_start_date] - start_date: {start_date}, "
+            f"[get_start_date] Assignment UUID: {self.assignment.uuid} - start_date: {start_date}, "
             f"end_date: {end_date}, "
             f"course_metadata: {course_metadata}"
         )
@@ -232,7 +232,7 @@ class BrazeCampaignSender:
         Returns the minimum of this assignment's auto-expiration date,
         the content's enrollment deadline, and the related policy's expiration timestamp.
         """
-        action_required_by_timestamp = get_automatic_expiration_date_and_reason(self.assignment, self.course_metadata) 
+        action_required_by_timestamp = get_automatic_expiration_date_and_reason(self.assignment, self.course_metadata)
         if not action_required_by_timestamp:
             return None
         return format_datetime_obj(
@@ -499,7 +499,7 @@ def send_email_for_new_assignment(new_assignment_uuid):
     Args:
         new_assignment_uuid: (string) the new assignment uuid
     """
-    assignment = _get_assignment_or_raise(new_assignment_uuid) 
+    assignment = _get_assignment_or_raise(new_assignment_uuid)
 
     campaign_sender = BrazeCampaignSender(assignment)
     braze_trigger_properties = campaign_sender.get_properties(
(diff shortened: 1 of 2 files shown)


enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -218,7 +218,11 @@ def get_start_date(self):
         start_date = self.normalized_metadata.get('start_date')
         end_date = self.normalized_metadata.get('end_date')
         course_metadata = self.course_metadata
-
+        logger.info(
+            f"[get_start_date] - start_date: {start_date}, "
+            f"end_date: {end_date}, "
+            f"course_metadata: {course_metadata}"
+        )
```
adamstankiewicz, line 225: [consider] Would it be helpful to include the `self.assignment.uuid` in this log (e.g., to query DD logs by assignment UUID), e.g.

```suggestion
        logger.info(
            f"[get_start_date] Assignment UUID: {self.assignment.uuid} - start_date: {start_date}, "
            f"end_date: {end_date}, "
            f"course_metadata: {course_metadata}"
        )
```

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -228,7 +232,7 @@ def get_action_required_by_timestamp(self):
         Returns the minimum of this assignment's auto-expiration date,
         the content's enrollment deadline, and the related policy's expiration timestamp.
         """
-        action_required_by_timestamp = get_automatic_expiration_date_and_reason(self.assignment, self.course_metadata)
+        action_required_by_timestamp = get_automatic_expiration_date_and_reason(self.assignment, self.course_metadata)
```
adamstankiewicz, line 235: nit: is the trailing space at the end of the line expected?

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -495,7 +499,7 @@ def send_email_for_new_assignment(new_assignment_uuid):
     Args:
         new_assignment_uuid: (string) the new assignment uuid
     """
-    assignment = _get_assignment_or_raise(new_assignment_uuid)
+    assignment = _get_assignment_or_raise(new_assignment_uuid)
```
adamstankiewicz, line 502: nit: is the trailing space at the end of the line expected?

enterprise_access/apps/content_assignments/utils.py
```diff
@@ -47,10 +50,18 @@ def get_self_paced_normalized_start_date(start_date, end_date, course_metadata):
     curr_date = datetime.now()
     pacing_type = course_metadata.get('pacing_type', {}) or None
     weeks_to_complete = course_metadata.get('weeks_to_complete', {}) or None
+    logger.info(f"Start Date: {start_date}, End Date: {end_date}, Course Metadata: {course_metadata}")
```
adamstankiewicz, line 53: [consider] Similar to related comment, I wonder if passing the `self.assignment.uuid` to `get_self_paced_normalized_start_date` to include the assignment identifier in the log messages would be helpful?

adamstankiewicz: That said, perhaps the logs you added in `get_start_date`, which also logs the same inputs as here, will suffice on its own? For example, could we discern which of the below code paths we're going through based on the log message from `get_start_date`?

enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -509,6 +513,9 @@ def send_email_for_new_assignment(new_assignment_uuid):
         'learner_portal_link',
         'action_required_by_timestamp'
     )
+
+    logger.info(f'Braze properties for assignment {assignment.uuid}: {braze_trigger_properties}')
```
adamstankiewicz, line 517: [inform] Upon further investigation, it looks like we _do_ already log the `braze_trigger_properties` [here](https://github.com/openedx/enterprise-access/blob/b756728a4acf73e92df3cdb945a42251bf3dc43a/enterprise_access/apps/content_assignments/tasks.py#L119-L125). We might not actually need to include this log line as a result?

The `[get_start_date]` logs will give us the additional missing information though, around what source metadata is used to derive the incorrect start date.

adamstankiewicz (approved):

PR #633, opened 2025-02-07
fix: use course-run specific metadata

hamzawaleed01:
**Description:**

**Jira:**
ENT-9890

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/utils.py
+++ b/enterprise_access/utils.py
@@ -245,3 +245,27 @@ def _days_from_now(days_from_now=0, date_format=None):
     if not date_format:
         return date
     return date.strftime(date_format)
+
+
+def get_course_run_metadata_for_assignment(assignment, content_metadata):
+    """
+    Retrieves metadata for a specific course run associated with an assignment. If the assignment has
+    a preferred course run, returns the metadata for that run. If the preferred run metadata is not 
+    found, returns normalized_metadata.
+
+    Args:
+        assignment (dict): The assignment object.
+        content_metadata (dict): The content metadata object.
+
+    Returns:
+        dict: Course run metadata if available, otherwise normalized_metadata.
+    """
+    if preferred_course_run_key := assignment.preferred_course_run_key:
+        course_runs = content_metadata.get('course_runs', [])
+        for course_run in course_runs:
+            if course_run.get('key') == preferred_course_run_key:
+                # manually applying existing key value mappings for consistent return structure
+                course_run['start_date'] = course_run.get('start')
+                course_run['end_date'] = course_run.get('end')
+                return course_run
+    return content_metadata.get('normalized_metadata', {})
--- a/enterprise_access/utils.py
+++ b/enterprise_access/utils.py
@@ -250,7 +250,7 @@ def _days_from_now(days_from_now=0, date_format=None):
 def get_course_run_metadata_for_assignment(assignment, content_metadata):
     """
     Retrieves metadata for a specific course run associated with an assignment. If the assignment has
-    a preferred course run, returns the metadata for that run. If the preferred run metadata is not 
+    a preferred course run, returns the metadata for that run. If the preferred run metadata is not
     found, returns normalized_metadata.
 
     Args:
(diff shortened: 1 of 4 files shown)


enterprise_access/utils.py
```diff
@@ -245,3 +245,27 @@ def _days_from_now(days_from_now=0, date_format=None):
...
+    Returns:
+        dict: Course run metadata if available, otherwise normalized_metadata.
+    """
+    if preferred_course_run_key := assignment.preferred_course_run_key:
+        course_runs = content_metadata.get('course_runs', [])
+        for course_run in course_runs:
+            if course_run.get('key') == preferred_course_run_key:
+                # manually applying existing key value mappings for consistent return structure
+                course_run['start_date'] = course_run.get('start')
+                course_run['end_date'] = course_run.get('end')
```
adamstankiewicz, line 269: Per my above related comment, if we continue to use `self.normalized_metadata` for the `start_date` and `end_date`, I believe these lines could be removed as this method would return an unmodified `course_run`?

enterprise_access/utils.py
```diff
@@ -245,3 +245,27 @@ def _days_from_now(days_from_now=0, date_format=None):
...
+        assignment (dict): The assignment object.
+        content_metadata (dict): The content metadata object.
+
+    Returns:
+        dict: Course run metadata if available, otherwise normalized_metadata.
+    """
+    if preferred_course_run_key := assignment.preferred_course_run_key:
+        course_runs = content_metadata.get('course_runs', [])
+        for course_run in course_runs:
+            if course_run.get('key') == preferred_course_run_key:
```
adamstankiewicz, line 266: nit: could opt to use `next` to extract the desired course run, if you wanted. might be a bit more "pythonic", e.g.:

```python
course_run = next((run for run in course_runs if run.get('key') == preferred_course_run_key), None)
if not course_run:
    logger.warning(
        'Course run metadata not found for preferred course run key %s in content metadata %s', 
        preferred_course_run_key,
        content_metadata.get('key'),
    )
    return course_run
```

(Opted to add logging when the `preferred_course_run_key` is not found in the list of `course_runs`).

enterprise_access/utils.py
```diff
@@ -245,3 +245,27 @@ def _days_from_now(days_from_now=0, date_format=None):
...
+                # manually applying existing key value mappings for consistent return structure
+                course_run['start_date'] = course_run.get('start')
+                course_run['end_date'] = course_run.get('end')
+                return course_run
+    return content_metadata.get('normalized_metadata', {})
```
adamstankiewicz, line 271: There are 2 different data structures returned by this method:
1. Above, a full `course_run` is returned. 
2. Here, only the `normalized_metadata` for the advertised course run is returned, which doesn't include the _full_ metadata.

Should this line be returning the full metadata of the advertised course run vs. just its `normalized_metadata` for consistent returns from this method? For example, `content_metadata.get('normalized_metadata', {})` won't contain the fields `weeks_to_complete` or `pacing_type`.

brobro10000, line 271: Great job including a fallback, but I would return the advertised course run's course run metadata as opposed to just normalized metadata.

Could be something along the lines of:
```py
def get_course_run_metadata_for_assignment(assignment, content_metadata):
  if preferred_course_run_key := assignment.preferred_course_run_key:
      ....
  advertised_course_run_uuid = content_metadata.get('advertised_course_run_uuid`);
  return next(course_run for course_run in content_metadata["course_runs"] if course["uuid"] == advertised_course_run_uuid)
  ```

**Review by brobro10000 — COMMENTED** [review]
Great job with the abstraction with a few suggestions.

File: enterprise_access/utils.py (hunk 3 above)
Comment (hamzawaleed01) [reply enterprise_access/utils.py]: Perfect! Adam also shared the same thought above and I wasn't sure what should be default return. Thanks for sharing!... just pushed it.

**Comment by hamzawaleed01** [comment]
Hey @brobro10000  @adamstankiewicz , thanks for the awesome insights! ❤️

I've resolved the comments I was sure about and left the rest for you folks to review.

File: enterprise_access/utils.py [hunk 4]
```diff
@@ -245,3 +245,31 @@ def _days_from_now(days_from_now=0, date_format=None):
...
+        if not course_run:
+            logger.warning(
+                'Metadata not found for preferred course run key %s in content metadata %s. Assignment UUID: %s',
+                preferred_course_run_key,
+                content_metadata.get('key'),
+                assignment.uuid
+            )
+        return course_run
+    advertised_course_run_uuid = content_metadata.get('advertised_course_run_uuid')
+    return next((run for run in course_runs if run.get("uuid") == advertised_course_run_uuid), None)
```
brobro10000, line 275: [nit] This turned out awesome. There is one potential hole the user could fall into where if there is if there exists a preferred course run key but no course run is returned from the course run key, we run the potential of returning the course run (which is a None value) anyway.

We can create a minor abstraction of `advertised_course_run_metadata` and utilize it in this function as such. 

``` py
def get_advertised_course_run_metadata(content_metadata):
    course_runs = content_metadata.get('course_runs', [])
    advertised_course_run_uuid = content_metadata.get('advertised_course_run_uuid')
    return next((run for run in course_runs if run.get('uuid') == advertised_course_run_uuid), None)

def get_course_run_metadata_for_assignment(assignment, content_metadata):
    course_runs = content_metadata.get('course_runs', [])

    # For run-based assignments, return metadata for the preferred course run if
    # available. If not, fallback to advertised run.
    if preferred_course_run_key := assignment.preferred_course_run_key:
        course_run = next((run for run in course_runs if run.get('key') == preferred_course_run_key), None)
        if not course_run:
            logger.warning(
                'Metadata not found for preferred course run key %s in content metadata %s. Assignment UUID: 
                %s',
                preferred_course_run_key,
                content_metadata.get('key'),
                assignment.uuid
            )
            # Fallback to advertised course run if preferred course run metadata is missing
            return get_advertised_course_run_metadata(content_metadata)
        return course_run
    
    # For course-based assignments, return metadata for the advertised course run
    return get_advertised_course_run_metadata(content_metadata)
```

brobro10000 (commented):
There is one final optimization that can be included to avoid a potential None value return for content metadata.

PR #636, opened 2025-02-10
feat: add an LmsApiClient.create_enterprise_customer() method

iloveagent57:
**Description:**
In support of https://2u-internal.atlassian.net/browse/ENT-9970

Adds a new API client method to create enterprise customer records.

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api_client/lms_client.py
+++ b/enterprise_access/apps/api_client/lms_client.py
@@ -105,6 +105,47 @@ class LmsApiClient(BaseOAuthClient):
             logger.exception(exc)
             raise
 
+    def create_enterprise_customer(self, *, name, slug, country, **kwargs):
+        """
+        Creates a new enterprise customer record.
+
+        Arguments:
+            name (string): The name of the customer.
+            slug (string): Slug of the enterprise customer.
+            country (string): The country code of the customer.
+            kwargs (dict): Any other fields to specify for the newly-created customer.
+        Returns:
+            dictionary containing enterprise customer metadata
+        """
+        payload = {
+            'name': name,
+            'slug': slug,
+            'country': country,
+            'site': {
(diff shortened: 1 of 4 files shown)


pwnage101 (approved):

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -105,6 +105,44 @@ def get_enterprise_customer_data(self, enterprise_customer_uuid=None, enterprise
...
+            )
+            response.raise_for_status()
+            payload = response.json()
+            logger.info(
+                'Successfully created customer %s with data %s',
+                name, payload,
+            )
+            return payload
+        except requests.exceptions.HTTPError as exc:
+            logger.exception('Failed to create enterprise customer with name %s', name)
```
adamstankiewicz, line 143: [consideration] Would it be helpful to log the entire `payload` here? E.g., maybe it errored due to one of the other fields besides `name`.

enterprise_access/apps/api_client/lms_client.py
```diff
@@ -105,6 +105,44 @@ def get_enterprise_customer_data(self, enterprise_customer_uuid=None, enterprise
...
+            response = self.client.post(
+                self.enterprise_customer_endpoint,
+                json=payload,
+                timeout=settings.LMS_CLIENT_TIMEOUT,
+            )
+            response.raise_for_status()
+            payload = response.json()
+            logger.info(
+                'Successfully created customer %s with data %s',
+                name, payload,
```
adamstankiewicz, line 139: nit: similar to Troy's comment above, should this log include the customer UUID (given name is not unique)?

iloveagent57: The uuid should be in `payload`.

adamstankiewicz (approved):

PR #639, opened 2025-02-13
fix: pass complete timestamp to Braze email trigger

hamzawaleed01:
**Description:**

**Jira:**
ENT-9890

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -225,7 +225,12 @@ def get_start_date(self):
             f"course_run_metadata: {course_run_metadata}"
         )
         return get_human_readable_date(
-            get_self_paced_normalized_start_date(start_date, end_date, course_run_metadata)
+            get_self_paced_normalized_start_date(
+                start_date,
+                end_date,
+                course_run_metadata
+            ),
+            BRAZE_ACTION_REQUIRED_BY_TIMESTAMP_FORMAT
```
adamstankiewicz, line 233: [suggestion] Should we abstract/rename the `BRAZE_ACTION_REQUIRED_BY_TIMESTAMP_FORMAT` to be more generically available to use? As is, the constant's name implies it should be used with the `*_ACTION_REQUIRED_BY_*` date, but here it's used with the start date.

For example, something like `BRAZE_TIMESTAMP_FORMAT`.

PR #648, opened 2025-02-25
feat: list_price is always populated in can_redeem

pwnage101:
This basically adds a fallback branch when calculating the list_price to serialize into the can-redeem endpoint.  The fallback logic relies on a net new call to enterprise-catalog to fetch content metadata containing price information agnostic of any catalog or customer.

Screenshot 2025-02-26 at 4 50 18 PM

ENT-9660

Local Dev Testing
===

For this manual test, I created an learner credit plan with only an assignable budget, but no assignments.

```
GET /api/v1/policy-redemption/enterprise-customer/298629a4-0fbf-44db-bcfb-9c604eca0bc8/can-redeem/?content_key=course-v1%3AedX%2BDemoX%2BDemo_Course
```

<details>
<summary>can-redeem response BEFORE commit has null list_price</summary>

```json
[
    {
        "content_key": "course-v1:edX+DemoX+Demo_Course",
        "list_price": null,
        "redemptions": [],
        "redeemable_subsidy_access_policy": null,
        "can_redeem": false,
        "reasons": [
            {
                "reason": "reason_learner_not_assigned_content",
                "user_message": "You can't enroll right now because this course is not assigned to you.",
                "policy_uuids": [
                    "8b460a30-ff5d-4810-8b56-18689677d672"
                ],
                "metadata": {
                        {
                            "email": "[edx@example.com](mailto:edx@example.com)",
                            "lms_user_id": 3
                        }
                    ]
                }
            }
        ]
    }
]
```
</details>

<details>
<summary>can-redeem response AFTER commit has populated list_price</summary>

```json
[
    {
        "content_key": "course-v1:edX+DemoX+Demo_Course",
        "list_price": {
            "usd": 149.0,
            "usd_cents": 14900
        },
        "redemptions": [],
        "redeemable_subsidy_access_policy": null,
        "can_redeem": false,
        "reasons": [
            {
                "reason": "reason_learner_not_assigned_content",
                "user_message": "You can't enroll right now because this course is not assigned to you.",
                "policy_uuids": [
                    "8b460a30-ff5d-4810-8b56-18689677d672"
                ],
                "metadata": {
                        {
                            "email": "[edx@example.com](mailto:edx@example.com)",
                            "lms_user_id": 3
                        }
                    ]
                }
            }
        ]
    }
]
```
</details>

Stage testing plan
===

This URL currently returns null list_price, but it should become non-null after merging:

https://enterprise-access.stage.edx.org/api/v1/policy-redemption/enterprise-customer/b705ea3b-f2ab-484f-83b9-78cf488f1704/can-redeem/?content_key=course-v1%3Aedx%2B101010%2B1T2025

Prod testing plan
===

This URL currently returns null list_price, but it should become non-null after merging:

https://enterprise-access.edx.org/api/v1/policy-redemption/enterprise-customer/89cbcdd9-c3cd-4dd7-8a33-2034b50b240e/can-redeem/?content_key=course-v1%3AUniversityofCambridge%2BSUP%2B3T2025

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
@@ -1784,6 +1784,13 @@ class BaseCanRedeemTestMixin:
         lms_client_instance = lms_client.return_value
         lms_client_instance.get_enterprise_user.return_value = TEST_USER_RECORD
 
+        catalog_get_and_cache_content_metadata_patcher = mock.patch(
+            'enterprise_access.apps.api.v1.views.subsidy_access_policy.get_and_cache_content_metadata',
+        )
+        self.mock_catalog_get_and_cache_content_metadata = catalog_get_and_cache_content_metadata_patcher.start()
+        self.mock_catalog_get_and_cache_content_metadata.return_value = {}
+
+        self.addCleanup(catalog_get_and_cache_content_metadata_patcher.stop)
         self.addCleanup(lms_client_patcher.stop)
         self.addCleanup(subsidy_client_patcher.stop)
         self.addCleanup(contains_key_patcher.stop)
@@ -1956,6 +1963,12 @@ class TestSubsidyAccessPolicyCanRedeemView(BaseCanRedeemTestMixin, APITestWithMo
             'unit': 'usd_cents',
             'all_transactions': [],
         }
+        self.mock_catalog_get_and_cache_content_metadata.return_value = {
+            'normalized_metadata': {
+                'content_price': 50,  # normalized_metadata always serializes USD.
+            },
+            'normalized_metadata_by_run': {},
+        }
         test_content_key_1 = "course-v1:edX+Privacy101+3T2020"
         test_content_key_2 = "course-v1:edX+Privacy101+3T2020_2"
         test_content_key_1_metadata_price = 29900
@@ -1996,9 +2009,9 @@ class TestSubsidyAccessPolicyCanRedeemView(BaseCanRedeemTestMixin, APITestWithMo
 
         # Check the response for the first content_key given.
         assert response_list[0]["content_key"] == test_content_key_1
-        # We should not assume that a list price is fetchable if the
-        # content cant' be redeemed - the content may not be in any catalog for any policy.
-        assert response_list[0]["list_price"] is None
+        # When there's no redeemable policy, the price returned is the current
+        # fixed price fetched directly from catalog.
+        assert response_list[0]["list_price"] == {'usd': 50.0, 'usd_cents': 5000}
         assert len(response_list[0]["redemptions"]) == 0
         assert response_list[0]["has_successful_redemption"] is False
         assert response_list[0]["redeemable_subsidy_access_policy"] is None
@@ -2031,7 +2044,7 @@ class TestSubsidyAccessPolicyCanRedeemView(BaseCanRedeemTestMixin, APITestWithMo
 
         # Check the response for the second content_key given.
         assert response_list[1]["content_key"] == test_content_key_2
-        assert response_list[1]["list_price"] is None
+        assert response_list[1]["list_price"] == {'usd': 50.0, 'usd_cents': 5000}
 
         assert len(response_list[1]["redemptions"]) == 0
         assert response_list[1]["has_successful_redemption"] is False
@@ -2263,6 +2276,12 @@ class TestSubsidyAccessPolicyCanRedeemView(BaseCanRedeemTestMixin, APITestWithMo
             "content_price": 1234,
             "enroll_by_date": '2001-01-01T00:00:00Z',
         }
+        self.mock_catalog_get_and_cache_content_metadata.return_value = {
+            'normalized_metadata': {
+                'content_price': 12.34,  # normalized_metadata always serializes USD.
+            },
+            'normalized_metadata_by_run': {},
+        }
 
         with mock.patch(
             'enterprise_access.apps.subsidy_access_policy.content_metadata_api.get_and_cache_content_metadata',
@@ -2368,6 +2387,13 @@ class TestSubsidyAccessPolicyCanRedeemView(BaseCanRedeemTestMixin, APITestWithMo
             'content_price': 19900,
         }
         self.mock_get_content_metadata.return_value = mock_subsidy_content_data
+        self.mock_catalog_get_and_cache_content_metadata.return_value = {
+            'normalized_metadata': {
+                'content_price': 199.00,  # normalized_metadata always serializes USD.
+            },
+            'normalized_metadata_by_run': {},
+        }
+
         query_params = {'content_key': test_content_key}
         if lms_user_id_override:
             query_params['lms_user_id'] = lms_user_id_override
@@ -2793,6 +2819,12 @@ class TestAssignedSubsidyAccessPolicyCanRedeemView(BaseCanRedeemTestMixin, APITe
             "content_price": content_key_for_redemption_metadata_price,
         }
         self.mock_get_content_metadata.return_value = mock_get_subsidy_content_data
+        self.mock_catalog_get_and_cache_content_metadata.return_value = {
+            'normalized_metadata': {
+                'content_price': content_key_for_redemption_metadata_price / 100,
+            },
+            'normalized_metadata_by_run': {},
+        }
 
         # It's an unredeemable response, so mock out some admin users to return
         mock_lms_client.return_value.get_enterprise_customer_data.return_value = {
@@ -2816,7 +2848,10 @@ class TestAssignedSubsidyAccessPolicyCanRedeemView(BaseCanRedeemTestMixin, APITe
 
         # Check the response for the first content_key given.
         assert response_list[0]["content_key"] == content_key_for_redemption
-        assert response_list[0]["list_price"] is None
+        assert response_list[0]["list_price"] == {
+            'usd': content_key_for_redemption_metadata_price / 100.0,
+            'usd_cents': content_key_for_redemption_metadata_price,
+        }
         assert response_list[0]["redemptions"] == []
         assert response_list[0]["has_successful_redemption"] is False
         assert response_list[0]["redeemable_subsidy_access_policy"] is None
--- a/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
@@ -27,6 +27,7 @@ from enterprise_access.apps.api import filters, serializers, utils
 from enterprise_access.apps.api.mixins import UserDetailsFromJwtMixin
 from enterprise_access.apps.api_client.lms_client import LmsApiClient
 from enterprise_access.apps.content_assignments.api import AllocationException
+from enterprise_access.apps.content_metadata.api import get_and_cache_content_metadata
 from enterprise_access.apps.core.constants import (
     SUBSIDY_ACCESS_POLICY_ALLOCATION_PERMISSION,
     SUBSIDY_ACCESS_POLICY_READ_PERMISSION,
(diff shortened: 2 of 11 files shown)


enterprise_access/apps/subsidy_access_policy/subsidy_api.py
```diff
@@ -146,8 +146,15 @@ def get_redemptions_by_content_and_policy_for_learner(policies, lms_user_id):
...
-                result[content_key][subsidy_access_policy_uuid].append(redemption)
+            matching_policies = [
+                policy for policy in policies_with_subsidy
+                if str(policy.uuid) == subsidy_access_policy_uuid
+            ]
+            # We can assume there's at most one matching policy because the ``policies`` arg passed into this function
+            # should not contain duplicates.
+            matching_policy = matching_policies[0] if matching_policies else None
+            if matching_policy:
+                result[content_key][matching_policy].append(redemption)
```
pwnage101, line 157: I slightly refactored this function to return different mapping:

* old: dict(content_key -> dict(policy UUID -> list of redemptions)
* new: dict(content_key -> dict(**policy instance** -> list of redemptions)

The purpose is to avoid another DB call in downstream code to fetch the policy instance.  We already pulled it into memory here in this function, so just don't discard it yet.

enterprise_access/apps/content_metadata/tests/test_api.py
```diff
@@ -15,10 +16,15 @@ class TestContentMetadataApi(TestCase):
...
+        super().tearDown()
+        # Clear the cache to prevent cache bleed across tests.
+        TieredCache.dangerous_clear_all_tiers()
+
     @mock.patch('enterprise_access.apps.content_metadata.api.EnterpriseCatalogApiClient', autospec=True)
-    def test_get_and_cache_metadata_happy_path(self, mock_client_class):
+    def test_get_and_cache_catalog_content_metadata_happy_path(self, mock_client_class):
         content_keys = ['course+A', 'course+B']
-        customer_uuid = uuid4()
+        catalog_uuid = uuid4()
```
pwnage101, line 27: A repetitive change in this file is replacing customer_uuid with catalog_uuid.  It was just wrong before.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -785,25 +792,26 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
-                ) from exc
+            if resolved_policy:
+                list_price = resolved_policy.get_list_price(lms_user_id, content_key)
+            elif successful_redemptions:
+                # Get the policy used for redemption and use that to compute the price. If the redemption was the result
+                # of assignment, the historical assignment price might differ from the canonical price. We prefer to
+                # display the redeemed price to avoid confusion.
+                successfully_redeemed_policy = policy_by_redemption_uuid[successful_redemptions[0]['uuid']]
+                list_price = successfully_redeemed_policy.get_list_price(lms_user_id, content_key)
+            else:
```
pwnage101, line 803: Here's the red "else" in the corresponding diagram:

Screenshot 2025-02-26 at 4 50 18 PM

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -785,25 +792,26 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
+                # In the case where the learner cannot redeem and has never redeemed this content, bypass the subsidy
+                # metadata endpoint and go straight to the source (enterprise-catalog) to find price information. This
+                # is rare, but can happen in at least these cases:
+                # * Courses are searchable because enterprise has an assignment-based policy, but nothing was assigned.
+                # * Enterprise customers that leverage the API directly always expect a non-null price.
```
adamstankiewicz, line 808: nit: I believe this code path will happen more often than rarely, e.g.:
* any enterprise learner visiting course page whose customer only has subscriptions, not learner credit.
* any enterprise learner visiting course page whose customer only has legacy coupon codes, not learner credit.
* any enterprise learner visiting course page whose customer only has legacy offers, not learner credit.
* non-redeemable policies that are not assignment-based (e.g., learner exceeded per-learner spend limit, or no funds remaining)

enterprise_access/apps/api_client/enterprise_catalog_client.py
```diff
@@ -84,3 +90,39 @@ def get_content_metadata_count(self, catalog_uuid):
...
+        """
+        Fetch catalog-/customer-agnostic content metadata.
+
+        Arguments:
+            content_id (str): ID of content to fetch.
+
+        Returns:
+            dict: serialized content metadata, or None if not found.
+        """
+        query_params = {'content_identifiers': content_id}
```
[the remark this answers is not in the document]
pwnage101: Indeed, that was the original intent, which I forgot.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: Hm, I _think_ these are all example where the endpoint gets called, but we don't actually read the list_price from the response.  My examples include only when list_price actually drives the amount displayed. LMK if that makes sense.

adamstankiewicz: I see, yeah your clarification does make sense. Although, while the calling frontend wouldn't actually read the `list_price` from the response in these additional cases, this code path is still executing, regardless of whether or not it gets displayed, right?

That's where "rare" trips me up a bit, since this code path in the additional cases still executes, even though its data doesn't get displayed. By saying "rare", it makes me think that this code path is truly rarely executed (even in these additional use cases), despite the code path still executing in practice (though its `list_price` not consumed).

From an API-first perspective, does it make sense for the comments to largely assume how consumers use its response? For instance, if/when Learner Portal changes to _always_ use the `list_price` from this API across all use cases, the comment would become stale/misleading.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -785,25 +792,26 @@ def can_redeem(self, request, enterprise_customer_uuid):
...
+                course_metadata = get_and_cache_content_metadata(content_key, coerce_to_parent_course=True)
+                usd_cents = (
+                    course_metadata['normalized_metadata_by_run'].get(content_key, {}).get('content_price') or
+                    course_metadata['normalized_metadata'].get('content_price')
+                )
```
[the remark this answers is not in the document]
pwnage101: Thanks for pointing this out!  This should be fixed up now.

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
pwnage101: Okay, I overhauled the comment.  Please take another look!

enterprise_access/apps/subsidy_access_policy/content_metadata_api.py
```diff
@@ -111,21 +111,25 @@ def get_list_price_for_content(enterprise_customer_uuid, content_key, content_me
...
 
     # Note that the "content_price" key is guaranteed to exist, but the value may be None.
-    return list_price_dict_from_usd_cents(content_metadata['content_price'])
+    return make_list_price_dict(integer_cents=content_metadata['content_price'])
 
 
-def list_price_dict_from_usd_cents(list_price_integer_cents):
+def make_list_price_dict(decimal_dollars=None, integer_cents=None):
     """
     Helper to compute a list price dictionary given the non-negative price of the content in USD cents.
```
adamstankiewicz, line 119: nit: docstring could be updated, if this method now deals with either `decimal_dollars` or `integer_cents` (and not just "USD cents").

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -2263,6 +2276,12 @@ def test_can_redeem_policy_beyond_enrollment_deadline(self, mock_lms_client, moc
             "content_price": 1234,
             "enroll_by_date": '2001-01-01T00:00:00Z',
         }
+        self.mock_catalog_get_and_cache_content_metadata.return_value = {
+            'normalized_metadata': {
+                'content_price': 12.34,  # normalized_metadata always serializes USD.
```
adamstankiewicz, line 2281: nit: should this be an integer vs. float?

[Example](https://enterprise-catalog.edx.org/api/v1/content-metadata/TheEconomist+CRIT/?format=json)

```
{
  "normalized_metadata": {
    "start_date": "2025-03-10T00:00:00Z",
    "end_date": "2025-03-21T00:00:00Z",
    "enroll_by_date": "2025-03-16T00:00:00Z",
    "enroll_start_date": null,
    "content_price": 828
  },
}
```

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -2368,6 +2387,13 @@ def test_can_redeem_lms_user_id_override_for_staff(
             'content_price': 19900,
         }
         self.mock_get_content_metadata.return_value = mock_subsidy_content_data
+        self.mock_catalog_get_and_cache_content_metadata.return_value = {
+            'normalized_metadata': {
+                'content_price': 199.00,  # normalized_metadata always serializes USD.
```
adamstankiewicz, line 2392: nit: should this be an integer vs. float?

[Example](https://enterprise-catalog.edx.org/api/v1/content-metadata/TheEconomist+CRIT/?format=json)

```
{
  "normalized_metadata": {
    "start_date": "2025-03-10T00:00:00Z",
    "end_date": "2025-03-21T00:00:00Z",
    "enroll_by_date": "2025-03-16T00:00:00Z",
    "enroll_start_date": null,
    "content_price": 828
  },
}
```

adamstankiewicz (approved):
LGTM! Just left a handful of non-blocking nits.

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
pwnage101: The serializer field type is actually a float: https://github.com/openedx/enterprise-catalog/blob/a387d9c82368405b0cda7cfa06ea96c11c0e2c04/enterprise_catalog/apps/catalog/serializers.py#L153-L154

I guess floats are serialized as seemingly integers in json when the value is close to an integer.

adamstankiewicz: Gotcha, thanks for linking to the serializer!

enterprise_access/apps/subsidy_access_policy/content_metadata_api.py
```diff
@@ -111,21 +111,36 @@ def get_list_price_for_content(enterprise_customer_uuid, content_key, content_me
             raise ContentPriceNullException(f'Could not determine list price for {content_key}') from exc
 
     # Note that the "content_price" key is guaranteed to exist, but the value may be None.
-    return list_price_dict_from_usd_cents(content_metadata['content_price'])
+    return make_list_price_dict(integer_cents=content_metadata['content_price'])
 
 
-def list_price_dict_from_usd_cents(list_price_integer_cents):
+def make_list_price_dict(decimal_dollars=None, integer_cents=None):
```
iloveagent57, line 117: nit: maybe it's worth doing some type validation on these? for example, so that we don't accidentally add code like `make_list_price_dict(decimal_dollars=1000)` when we really mean `make_list_price_dict(integer_cents=1000)`.

iloveagent57 (approved):

enterprise_access/apps/subsidy_access_policy/content_metadata_api.py
pwnage101: Hm, not sure how one would differentiate since e.g. 1000 dollars is a believable price.

iloveagent57: Yeah, I just mean in the sense that it's not a believable `Decimal`. Maybe adding
```python
assert (integer_cents is None or isinstance(integer_cents, int))
assert (decimal_dollars is None or isinstance(decimal_dollars, (Decimal, float)))
```
...something like that, just as a foot-gun defense.

PR #651, opened 2025-03-03
feat: protect against decimal cents.

pwnage101:
This should hopefully surface buggy inputs before they get harder to track as they get passed along.

ENT-9660


iloveagent57 (approved):

pwnage101:
Note that I'm not protecting against integers passed to `decimal_dollars`.  This is because I've noticed that often decimal dollars are serialized in our API responses as integers.  I'm not sure why.  If access fetched an API which returned an integer dollar value, and it gets passed along as an integer via the `decimal_dollars` argument, we still want to treat it as dollars instead of throwing an error.

PR #654, opened 2025-03-04
fix: improve handling of enterprise customers with learner portal disabled

adamstankiewicz:
**Description:**

Ensures the Learner Portal BFF API only returns enterprise customers with the enterprise learner portal enabled.

**Jira:**
ENT-9937

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/bffs/tests/test_handlers.py
```diff
@@ -172,21 +172,58 @@ def test_load_and_process_without_learner_portal_enabled(self, mock_get_enterpri
...
+        mock_customer_without_learner_portal = {
+            **self.mock_enterprise_customer,
+            'enable_learner_portal': False,
+        }
         mock_get_enterprise_customers_for_user.return_value = {
             **self.mock_enterprise_learner_response_data,
-            'results': [{
-                **self.mock_enterprise_customer,
-                'enable_learner_portal': False,
-            }],
```
adamstankiewicz, line 180: [inform] This mocked data was incorrect; the `results` is a list of `EnterpriseCustomerUser` records, not `EnterpriseCustomer`.

PR #655, opened 2025-03-05
fix: [can-redeem] bogus content should not throw 500 response

pwnage101:
Previous behavior before we required a non-null list_price was to return 200 response with reason = "content_not_in_catalog" if the content key was totally fake.  Now that we are no longer tolerant of NOT including the price in the response, we must return some sort of non-200 response.

Before this PR, the can-redeem endpoint would return 500.  This fixes the response to return 422 Unprocessable to at least indicate that the requester did something wrong.


enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -2785,7 +2785,7 @@ def test_can_redeem_assigned_policy(
         # Only an errored assignment exists.
         {'has_cancelled_assignment': False, 'has_failed_assignment': True},
         # No assignment exists for the learner/content pair to check.
-        {'has_cancelled_assignment': False, 'has_failed_assignment': True},
+        {'has_cancelled_assignment': False, 'has_failed_assignment': False},
```
pwnage101, line 2788: This is not related to my changes, just noticed that we had a duplicate test case.

PR #661, opened 2025-03-17
feat: Added Learner Credit Request related models

zamanafzal:
**Description**:
This PR introduces updates related to LearnerCreditRequest and its associated models.

**Changes:**
1. Added browse_and_request Field in SubsidyAccessPolicy

> Enables or disables the Browse and Request feature.

2. Introduce LearnerCreditRequest Table

> Stores all learner credit requests and their states.

> Each request is linked to LearnerContentAssignment using a OneToOneField.

3. Added LearnerCreditRequestConfiguration Table

> Holds configuration settings for credit request processing.

**Jira:**
[ENT-10105
](https://2u-internal.atlassian.net/browse/ENT-10105)
**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/subsidy_request/models.py
```diff
@@ -306,3 +306,71 @@ def update_course_info_for_subsidy_request(sender, **kwargs):  # pylint: disable
...
+    )
+
+    history = HistoricalRecords()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=['user', 'enterprise_customer_uuid', 'course_id'],
+                name='unique_learner_course_request',
+                condition=models.Q(state__in=[SubsidyRequestStates.REQUESTED, SubsidyRequestStates.PENDING]),
```
zamanafzal, line 343: The UniqueConstraint in LearnerCreditRequest ensures that only one active request (REQUESTED or PENDING state) can exist for a combination of user, enterprise_customer_uuid, and course_id.

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -306,3 +307,80 @@ def update_course_info_for_subsidy_request(sender, **kwargs):  # pylint: disable
...
+    subsidy_access_policy = models.ForeignKey(
+        SubsidyAccessPolicy,
+        related_name="learner_credit_requests",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        help_text="The SubsidyAccessPolicy that this learner credit request is associated with.",
+    )
+    assignment = models.OneToOneField(
+        'content_assignments.LearnerContentAssignment',  # pylint: disable=all
```
zamanafzal, line 327: I've used String Reference to fix a circular import issue.

enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -158,6 +158,12 @@ class Meta:
             'it disappear from all frontends, effectively soft-deleting it. Default is False (deactivated).'
         ),
     )
+    browse_and_request = models.BooleanField(
+        default=False,
+        help_text=(
+            'If selected, it means browse and request is active on this policy.'
+        ),
+    )
```
sameenfatima78, line 166: nit: Can we rename it to 'bnr_enabled'? This makes it more specific and clearly defines the purpose of the flag."

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -306,3 +306,71 @@ def update_course_info_for_subsidy_request(sender, **kwargs):  # pylint: disable
...
+
+
+class LearnerCreditRequest(SubsidyRequest):
+    """
+    Stores information related to a learner credit request.
+
+    .. no_pii: This model has no PII
+    """
+
+    subsidy_access_policy = models.ForeignKey(
```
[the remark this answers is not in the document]
sameenfatima78: Question: Can we include `subsidy_access_policy` as a cached property instead of a model field as we discussed earlier?

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -306,3 +307,86 @@ def update_course_info_for_subsidy_request(sender, **kwargs):  # pylint: disable
...
+        related_name="learner_credit_requests",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        help_text="The SubsidyAccessPolicy that this learner credit request is associated with.",
+    )
+
+    assignment = models.OneToOneField(
+        'content_assignments.LearnerContentAssignment',  # pylint: disable=all
+        related_name="credit_request",
```
sameenfatima78, line 330: the related name could simply be called "requests"

enterprise_access/apps/subsidy_request/models.py
sameenfatima78: Question: Shouldn't the uniqueness be defined at a budget level instead of customer?

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -306,3 +307,86 @@ def update_course_info_for_subsidy_request(sender, **kwargs):  # pylint: disable
         subsidy_type,
         str(subsidy_request.uuid),
     )
+
+
+class LearnerCreditRequest(SubsidyRequest):
```
sameenfatima78, line 312: I think we should have an additional field to save `course_key`. I don't think the `SubsidyRequest` model currently stores it. It only has `course_run_id`.

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -306,3 +307,86 @@ def update_course_info_for_subsidy_request(sender, **kwargs):  # pylint: disable
...
+
+    def decline(self, reviewer, reason=None):
+        self.reviewer = reviewer
+        self.state = SubsidyRequestStates.DECLINED
+        self.decline_reason = reason
+        self.reviewed_at = localized_utcnow()
+        self.save()
+
+
+class LearnerCreditRequestConfiguration(TimeStampedModel):
```
sameenfatima78, line 370: Would it make sense to include a cached_property for `enterprise_customer_uuid`?

sameenfatima78 (commented):
Dropped a couple of suggestions.

enterprise_access/apps/subsidy_request/models.py
zamanafzal: `SubsidyRequest` has the `course_id` field.
https://github.com/openedx/enterprise-access/blob/de56f3b4d8ecc8188bb4b5671e5ca75f78cc11df/enterprise_access/apps/subsidy_request/models.py#L47

enterprise_access/apps/subsidy_request/models.py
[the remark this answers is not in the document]
zamanafzal: Since the relation is persistent and needs to be stored in the database and there is no other table from which we can get this information so I guess this is not possible to do it using a cached property.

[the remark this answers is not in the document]
iloveagent57: We should change the direction of the FK relationship to be defined on the SubsidyAccessPolicy model, which will require introducing a new model here in the `subsidy_request` module, called something like `LearnerCreditRequestCongifuration`.  See the diagram below - you'll end up with a OneToOneField on SubsidyAccessPolicy, very similar to the `SubsidyAccessPolicy.assignment_configuration` field.  Having the relationship go in that direciton is important for a couple reasons:
1. It's consistent with the direction of the relationship between policies and assignments.
2. It keeps things so that no app in this repo (aside from `/api`) depends on `subsidy_access_policy`, which helps keeps code dependencies clean.
image

Also, if you change things in this manner, you'll be able to get rid of the `SubsidyAccessPolicy.browse_and_request` boolean field. Instead, the existence of a `LearnerCreditRequestConfiguration` object associated with a policy can indicate whether B&R is enabled for a policy.  You can add a property to the SubsidyAccessPolicy to make that lookup easy.

sameenfatima78 (commented):
@zamanafzal One thing we should verify is whether the SubsidyRequest.state field accounts for all necessary lifecycle states for `LearnerCreditRequest` instances?

https://lucid.app/lucidchart/52c73279-b624-4531-b239-eaf17172050e/edit?viewport_loc=4489%2C682%2C2552%2C1430%2CWoEgvmhzd6iw&invitationId=inv_4bf7280c-7c9d-440f-ba99-fa7d1c45f985

enterprise_access/apps/subsidy_request/models.py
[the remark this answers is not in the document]
zamanafzal: Thank you for the detailed feedback @iloveagent57. The `LearnerCreditRequestConfiguration` model in the `subsidy_request` module is already designed with a `OneToOneField` relationship to `SubsidyAccessPolicy`. This approach was chosen because most `SubsidyAccessPolicy` instances do not have a corresponding `LearnerCreditRequestConfiguration`, resulting in this field often being empty.

However, we can still retrieve the related data using the reverse relationship: SubsidyAccessPolicy.learner_credit_request_config. If having an empty record is not a concern, we could consider defining the OneToOneField relation in `SubsidyAccessPolicy` instead of `LearnerCreditRequestConfiguration`. Please let me know your thoughts on this.

That's a great suggestion to get rid of the `SubsidyAccessPolicy.browse_and_request` boolean field and handle it with `LearnerCreditRequestConfiguration` association. I'll remove `browse_and_request` and handle it this way.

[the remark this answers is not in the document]
iloveagent57: > If having an empty record is not a concern, we could consider defining the OneToOneField relation in SubsidyAccessPolicy instead of LearnerCreditRequestConfiguration. Please let me know your thoughts on this
Ah, I didn't notice that `LearnerCreditRequestConfiguration` 1-1 field. Yes, let's define the 1-1 field on `SubsidyAccessPolicy` and have it point at `LearnerCreditRequestConfiguration`.  That will help us adhere to our guiding principle of having the `subsidy_access_policy` module depend on "everything", and have nothing depend on _it_. It's not a concern that many SAP records will have a null value for this 1-1 field.

> That's a great suggestion to get rid of the SubsidyAccessPolicy.browse_and_request boolean field and handle it with LearnerCreditRequestConfiguration association. I'll remove browse_and_request and handle it this way.
Excellent!

zamanafzal:
@sameenfatima78 That was a good callout. I don't see two states `expired` and `accepted` in `SubsidyRequestStates'. What do you suggest about this?

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -306,3 +307,78 @@ def update_course_info_for_subsidy_request(sender, **kwargs):  # pylint: disable
...
+        """
+        Return human-readable string representation.
+        """
+        if self.course_id:
+            return f'<LearnerCreditRequest for user {self.user} and course {self.course_id}>'
+        return f'<LearnerCreditRequest for user {self.user}>'
+
+    def approve(self, reviewer):
+        self.reviewer = reviewer
+        self.state = SubsidyRequestStates.PENDING
```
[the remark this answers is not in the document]
zamanafzal: Yes, thats right. Thanks for pointing this out.

sameenfatima78:
> @sameenfatima78 That was a good callout. I don't see two states `expired` and `accepted` in `SubsidyRequestStates'. What do you suggest about this?

@zamanafzal We also want to introduce a new state called "cancelled" to cover cases where an admin revokes an approved request. As for "Accepted," I believe we already have a state named "Approved" for that.

enterprise_access/apps/subsidy_request/constants.py
```diff
@@ -4,20 +4,36 @@
...
+    REVERSED = 'reversed'
+
+    LC_REQUEST_STATES = (
+        (ACCEPTED, "Accepted"),
+        (CANCELLED, "Cancelled"),
+        (EXPIRED, "Expired"),
+        (REVERSED, "Reversed"),
+    )
+
+    CHOICES = COMMON_STATES + LC_REQUEST_STATES
```
sameenfatima78, line 35: 👏 👏

sameenfatima78 (approved):
Looks good! Thank you!

PR #662, opened 2025-03-18
docs: Add ADR for stripe event consumption

pwnage101:
ENT-10075


docs/decisions/0025-stripe-event-consumption-delivery.rst
```diff
@@ -0,0 +1,108 @@
...
+
+     * Use `de-duplication <https://docs.stripe.com/webhooks#handle-duplicate-events>`_.
+
+   * After successful ingestion, respond with HTTP 200 to tell Stripe to stop retrying sending the event.
+
+#. **Pick:** Implement an asynchronous event picker to continuously try to pick events to handle/forward.
+
+   * Architecturally, this is an infinite loop via management command running in a dedicated container/pod.
+
+     * Alternatively, this might work as a Celery "beat" scheduled for every 10 seconds.  This could reduce
```
adamstankiewicz, line 44: [question] How would this alternative 10-second "beat" schedule compare to the infinite loop approach in terms of event delivery timeliness? Just pondering the tradeoff between potentially slower delivery and reduced code/infra.

docs/decisions/0025-stripe-event-consumption-delivery.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Context
+=======
+
+We intend to utilize Stripe to manage subscription and payment lifecycles for the Enterprise Self-Service Purchasing
+feature. We will need a robust way to ingest events from Stripe, handle those events, and forward them to SalesForce.
+For example, when stripe emits an ``invoice.paid`` event, that needs to be communicated to SalesForce in order to
+trigger the creation/update of Customers, Opportunities, etc. Furthermore, that same event needs to be communicated to
+our edX Enterprise backend so that we can immediately followup with the customer to confirm that we have begun to
+provision their product.
```
adamstankiewicz, line 16: [clarification] Is it this proposed event consumption/delivery system that will be communicating to the edX Enterprise backend? Or are we relying on SalesForce's handling of the forwarded events to trigger a request to the Provisioning API to communicate with the edX Enterprise backend?

pwnage101: Both!  State-altering actions need to be handled by SalesForce calling our provisioning API.  However, we also may need to perform non-state-altering actions (e.g. product pending emails) and actions which should be triggered by events that we won't even send to SalesForce (e.g. subscription canceled emails), direct and immediate ingestion/handling of events into the edX backend would be required.

My proposed SSP architecture isolates the domain of recurring payments to Stripe + edX backend, and SalesForce would not be aware that the invoices it receives are part of a schedule or attached to a Stripe subscription. Hence, not all events would be delivered to SalesForce, only the invoice.paid events.

adamstankiewicz: Got it, thanks for the clarification. This makes sense :)

docs/decisions/0025-stripe-event-consumption-delivery.rst
adamstankiewicz: Ah, good find on the additional downsides. A 1-second sleep with the infinite loop seems reasonable!

adamstankiewicz (approved):
LGTM!

docs/decisions/0025-stripe-event-consumption-delivery.rst
```diff
@@ -0,0 +1,133 @@
...
+Alternatives Considered
+=======================
+
+Rejected
+========
+
+* Queue event handler immediately as part of the ingestion, instead of via an asynchronous picker.
+
+  * This is too sensitive to errors during ingestion, which could result in dropped events.
+
```
iloveagent57, line 96: Why not queue an _async_ task immediately as part of ingestion? What kind of errors is this sensitive to that could drop events?  Couldn't we do something like this to avoid those errors?
```python
try:
    with transaction.atomic():
        persist_the_event_data()
        transaction.on_commit(handle_the_event.apply_async)
    return Response(200)
except:
    raise # return error response
```

iloveagent57: If we enqueued an async task on ingestion, we could additionally run a k8s cronjob to perform the pick logic you defined above. This would serve to pick up any ingested-but-unprocessed events that may otherwise be orphaned due to celery malfunctions, etc.

pwnage101: Hm, I think you're right! I probably mis-remembered the real motivation which was just that it would be less code to write/test. As long as we attempt to persist the event & COMMIT before attempting to queue it, then we can't "drop" the event even if queueing the handler crashes.

PR #665, opened 2025-03-20
fix: update integration warning logic to check active LMS integrations

ahmad-hassan-dev:
**Description:**
Adjust the display logic for the Integration Warning modal in the Learner Portal. Previously, the modal displayed when search was enabled and an identity provider was configured. This resulted in confusion for customers who did not have an LMS integration but saw the modal due to having SSO enabled.

Changes:
- Update the condition to display the Integration Warning modal only if the enterprise customer has an active LMS integration.
- Use the active_integrations field from the BFF or direct LMS API call as the new condition.
- Ensure customers without LMS integrations do not see the modal, regardless of SSO setup.

**Jira:**
[ENT-10068 - IntegrationWarning showing for a customer with search enabled and no identity provider.](https://2u-internal.atlassian.net/browse/ENT-10068)

### DIFF
--- a/enterprise_access/apps/bffs/handlers.py
+++ b/enterprise_access/apps/bffs/handlers.py
@@ -206,11 +206,12 @@ class BaseLearnerPortalHandler(BaseHandler, BaseLearnerDataMixin):
         """
         # Learner Portal is enabled, so transform the enterprise customer data.
         identity_provider = enterprise_customer.get("identity_provider")
+        active_integrations = enterprise_customer.get("active_integrations")
         disable_search = bool(
             not enterprise_customer.get("enable_integrated_customer_learner_portal_search", False) and
             identity_provider
         )
-        show_integration_warning = bool(not disable_search and identity_provider)
+        show_integration_warning = bool(not disable_search and active_integrations)
 
         return {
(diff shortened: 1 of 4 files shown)


sameeramin (approved):
LGTM 👍🏼

enterprise_access/apps/bffs/handlers.py
```diff
@@ -206,11 +206,12 @@ def transform_enterprise_customer(self, enterprise_customer):
         """
         # Learner Portal is enabled, so transform the enterprise customer data.
         identity_provider = enterprise_customer.get("identity_provider")
+        active_integrations = enterprise_customer.get("active_integrations")
```
zamanafzal, line 209: Can we cover the test case for this?

sameenfatima78 (approved):
Looks good once unit test is added!

enterprise_access/apps/bffs/handlers.py
ahmad-hassan-dev: Hey @zamanafzal , @adamstankiewicz, I’ve updated the tests based on the new show_integration_warning logic.

zamanafzal (approved):

PR #666, opened 2025-03-21
feat: integrate algolia api keys into the bff base response

brobro10000:
**Description:**
Adds the algolia secured api key response from enterprise catalog to the root BFF response. This involves updating the serializer, context and response builder.

**Jira:**
[ENT-10085](https://2u-internal.atlassian.net/browse/ENT-10085)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/v1/tests/test_bff_views.py
+++ b/enterprise_access/apps/api/v1/tests/test_bff_views.py
@@ -160,6 +160,8 @@ class TestLearnerPortalBFFViewSet(TestHandlerContextMixin, MockLicenseManagerMet
                     },
                 },
             ],
+            'secured_algolia_api_key': self.mock_secured_algolia_api_key,
+            'catalog_uuids_to_catalog_query_uuids': self.mock_catalog_uuids_to_catalog_query_uuids,
             'active_enterprise_customer': self.expected_enterprise_customer,
             'staff_enterprise_customer': None,
             'should_update_active_enterprise_customer_user': self.mock_should_update_active_enterprise_customer_user,
@@ -232,6 +234,7 @@ class TestLearnerPortalBFFViewSet(TestHandlerContextMixin, MockLicenseManagerMet
     def test_dashboard_empty_state_with_permissions(
         self,
         mock_get_enterprise_customers_for_user,
+        mock_get_secured_algolia_api_key_for_user,
         mock_get_subscription_licenses_for_learner,
         mock_get_default_enrollment_intentions_learner_status,
         mock_get_enterprise_course_enrollments,
@@ -256,6 +259,7 @@ class TestLearnerPortalBFFViewSet(TestHandlerContextMixin, MockLicenseManagerMet
             'context': role_context,
         }])
         mock_get_enterprise_customers_for_user.return_value = self.mock_enterprise_learner_response_data
+        mock_get_secured_algolia_api_key_for_user.return_value = self.mock_secured_algolia_api_key_response
         mock_get_subscription_licenses_for_learner.return_value = self.mock_subscription_licenses_data
         mock_get_default_enrollment_intentions_learner_status.return_value =\
             self.mock_default_enterprise_enrollment_intentions_learner_status_data
@@ -284,6 +288,7 @@ class TestLearnerPortalBFFViewSet(TestHandlerContextMixin, MockLicenseManagerMet
     def test_dashboard_with_subscriptions(
         self,
         mock_get_enterprise_customers_for_user,
+        mock_get_secured_algolia_api_key_for_user,
         mock_get_default_enrollment_intentions_learner_status,
         mock_get_subscription_licenses_for_learner,
         mock_get_enterprise_course_enrollments,
@@ -296,6 +301,7 @@ class TestLearnerPortalBFFViewSet(TestHandlerContextMixin, MockLicenseManagerMet
             'context': self.mock_enterprise_customer_uuid,
         }])
         mock_get_enterprise_customers_for_user.return_value = self.mock_enterprise_learner_response_data
+        mock_get_secured_algolia_api_key_for_user.return_value = self.mock_secured_algolia_api_key_response
         mock_subscription_licenses_data = {
             'customer_agreement': self.mock_customer_agreement,
             'results': [self.mock_subscription_license],
@@ -337,6 +343,7 @@ class TestLearnerPortalBFFViewSet(TestHandlerContextMixin, MockLicenseManagerMet
     def test_dashboard_with_subscriptions_license_activation(
         self,
         mock_get_enterprise_customers_for_user,
+        mock_get_secured_algolia_api_key_for_user,
         mock_get_default_enrollment_intentions_learner_status,
         mock_get_subscription_licenses_for_learner,
         mock_get_enterprise_course_enrollments,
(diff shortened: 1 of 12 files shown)


enterprise_access/utils.py
```diff
@@ -285,3 +287,43 @@ def get_course_run_metadata_for_assignment(assignment, content_metadata):
 
     # For course-based assignments, return metadata for the advertised course run
     return get_advertised_course_run_metadata(content_metadata)
+
+
+def time_until(date_str: str, unit: str = 'seconds', exact: bool = False):
```
iloveagent57, line 292: Django has some nice utilities around parsing dates in a timezone-aware way, and so on:
https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.dateparse.parse_datetime
You could achieve what you're doing here with something like:
```python
from django.utils import dateparse, timezone 
delta = dateparse.parse_datetime(date_str) - timezone.now() # result will be a timedelta object
delta_seconds = delta.seconds
```

enterprise_access/apps/bffs/api.py
```diff
@@ -96,6 +103,43 @@ def get_and_cache_enterprise_customer(
...
+    client = EnterpriseCatalogUserV1ApiClient(request)
+    response_payload = client.get_secured_algolia_api_key(
+        enterprise_customer_uuid=enterprise_customer_uuid,
+    )
+
+    # Calculate timeout based on response_payload's valid_until field, it should be the
+    # valid until - the smallest unit of time
+    algolia_response = response_payload.get('algolia', {})
+    if valid_until_date := algolia_response.get('valid_until'):
+        timeout = time_until(valid_until_date, 'seconds') - EPSILON_SECONDS
```
iloveagent57, line 138: we probably want a larger buffer here, maybe 10 seconds?  You could also define `epsilon_seconds` as a simple variable in this function - it's just a little buffer we're building in, it's unlikely that we'd ever have to adjust it.

enterprise_access/apps/bffs/api.py
```diff
@@ -381,3 +425,17 @@ def transform_enterprise_customer_users_data(data, request, enterprise_customer_
...
+def transform_secured_algolia_api_key_response(secured_algolia_api_key_response):
+    """
+    Retrieve the secured algolia api key
+    """
+    catalog_uuids_to_catalog_query_uuids = secured_algolia_api_key_response.get(
+        'catalog_uuids_to_catalog_query_uuids',
+        {}
+    )
+    if algolia := secured_algolia_api_key_response.get('algolia', {}):
+        secured_algolia_api_key = algolia.get('secured_api_key', None)
```
iloveagent57, line 439: nit: `dict.get()` returns `None` by default if the key wasn't found and no default was provided, you don't have to specify it with the second argument here.

iloveagent57 (commented):
Couple small comments, this is looking good!

enterprise_access/utils.py
```diff
@@ -285,3 +289,81 @@ def get_course_run_metadata_for_assignment(assignment, content_metadata):
...
+            result = time_delta_seconds / 3600
+        case 'days':
+            result = time_delta_seconds / 86400
+        case _:
+            raise ValueError("Invalid unit. Use 'seconds', 'minutes', or 'days'.")
+
+    return result if exact else math.floor(result)
+
+
+def determine_timeout_offset(
```
adamstankiewicz, line 339: [suggestion] I feel this `determine_timeout_offset ` method and the `time_until` method could be simplified here. Would something like the following be a bit cleaner?

```python
def determine_timeout_offset(date_str, epsilon_seconds = 10):
    parsed_time = dateparse.parse_datetime(date_str)
    if not parsed_time:
      raise ValueError(f"Invalid date string: {date_str}")
    now = timezone.now()
    timeout = max(0, (parsed_time - now).total_seconds() - epsilon_seconds)
    return timeout
```

Where `timeout` will always be an int in seconds, `>= 0`.

enterprise_access/utils.py
```diff
@@ -285,3 +289,81 @@ def get_course_run_metadata_for_assignment(assignment, content_metadata):
...
+        valid_until_date_str: str,
```
adamstankiewicz, line 340: [consideration] If creating a generic util method, I believe its args should be agnostic to the semantic meaning of the date string, i.e. should not reference `valid_until`.

enterprise_access/utils.py
```diff
@@ -285,3 +289,81 @@ def get_course_run_metadata_for_assignment(assignment, content_metadata):
...
+        epsilon_seconds: int = 10,
+        timeout: int = DEFAULT_CACHE_TIMEOUT
+) -> Union[float, int]:
+    """
+    Determines the effective timeout duration based on a 'valid until' UTC timestamp,
+    a fallback timeout in seconds, and an epsilon buffer.
+
+    This function uses `time_until` to calculate the time (in seconds) until the provided
+    `valid_until_date_str`. If the result is negative (i.e., the time is in the past), the
+    fallback `timeout` value is used instead.
```
adamstankiewicz, line 350: I'm not quite following why "[i]f the result is negative (i.e., the time is in the past), the fallback `timeout` value is used instead."

Can you help me understand why if the `valid_until_date_str` is already in the past, we'd want to fallback to the `DEFAULT_CACHE_TIMEOUT` of 5 minutes? If it's already in the past (which shouldn't really happen in practice for this specific use case), I think we'd want the timeout to be `0` to force a re-fetch to get a new `valid_until`? 🤔

enterprise_access/utils.py
brobro10000: updated

enterprise_access/utils.py
brobro10000: Done

enterprise_access/apps/api/v1/tests/test_bff_views.py
```diff
@@ -404,19 +412,21 @@ def test_dashboard_with_subscriptions_license_activation(
...
+            mock_get_secured_algolia_api_key_for_user,
+            mock_get_enterprise_course_enrollments,
+            mock_get_enterprise_customer_data,
+            mock_auto_apply_license,
+            identity_provider,
+            is_staff_request_user,
+            has_plan_for_auto_apply,
+            has_existing_activated_license,
+            has_existing_revoked_license,
+            auto_apply_with_universal_link,
```
adamstankiewicz, line 428: nit: not sure these extra indents are necessary?

enterprise_access/apps/bffs/context.py
```diff
@@ -197,6 +207,17 @@ def _initialize_common_context_data(self):
         if not self.enterprise_customer_uuid:
             self._enterprise_customer_uuid = self.enterprise_customer.get('uuid')
 
+        # Initialize the secured algolia api keys metadata derived from enterprise catalog
+        try:
+            self._initialize_secured_algolia_api_keys()
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(
```
adamstankiewicz, line 214: Two questions/considerations:
1. `self. _initialize_secured_algolia_api_keys()` itself wraps its entire function body in a try/except, where I don't believe an error would not be surfaced to this try/except given it was already caught within `self. _initialize_secured_algolia_api_keys()`. Given this, would it make sense to keep only 1 layer of try/except for this method?
2. When handling an error related to `self. _initialize_secured_algolia_api_keys()`, do we want the error to be returned in the BFF API response in its `errors` list?

iloveagent57 (approved):

adamstankiewicz (commented):
Looking really good, just had a couple questions/considerations around error handling before leaving a 👍

enterprise_access/apps/bffs/context.py
adamstankiewicz: [clarification] For (2), if we want the error to be surfaced in the BFF API response itself, `self.add_error(...)`.

adamstankiewicz (approved):

PR #671, opened 2025-03-31
feat: provisioning workflow-based app

iloveagent57:
**Jira:**
ENT-9971

**Testing**
1. Run migrations.
2. Observe that http://localhost:18270/admin/provisioning/ exists.
3. Make a request to the provisioning endpoint:
```
curl --location 'http://localhost:18270/api/v1/provisioning' \
--header 'Authorization: JWT [YOUR JWT HERE]' \
--header 'Content-Type: application/json' \
--data-raw '{
    "enterprise_customer": {
        "name": "Test customer B",
        "country": "US",
        "slug": "test-customer-b"
    },
    "pending_admins": [
        {
            "user_email": "test-admin-1@foo.com"
        }
    ]
}'
```
4. This should result in a customer and admin users being created (or fetched if they already exist).
5. There should be successful worfklow and step records corresponding to your provisioning request, and they should have references to/from each other.
image
image
image

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/workflow/serialization.py
```diff
@@ -0,0 +1,38 @@
+"""
+Workflow Steps do work on objects, not on dicts.  The BaseInputOutput
+class below can be subclassed by attrs classes, so that every input/output
+class utilized by workflow steps can be counted on to have a from_dict()
+classmethod to generate an object from a dict, and a to_dict() method
+to turn an instance of an input/output class back into a dict (in other words
+to structure and unstructure, or to deserialize and then serialize).
+"""
```
iloveagent57, line 8: I need to write up an ADR about "why attrs/cattrs?" still, but here's some justification from the attrs docs: https://www.attrs.org/en/stable/why.html#dicts

enterprise_access/apps/provisioning/models.py
```diff
@@ -0,0 +1,191 @@
+"""
+Workflow models for the customer-and-subsidy-provisioning business domain.
+"""
+from uuid import UUID
+
+from attrs import define, field, validators
+from django_countries import countries
```
adamstankiewicz, line 7: [curious] Does `django_countries` correlate with the same list of countries returned by the `https://courses.edx.org/api/user/v2/account/registration/` REST API?

[context] The Enterprise Checkout MFE plans to utilize the countries provided by this API in the "Country" field during the self-service trial sign-up.

Example (partial) response:

```
{
  "method": "post",
  "submit_url": "/api/user/v2/account/registration/",
  "fields": [
    {
      "name": "country",
      "label": "Country or Region of Residence",
      "type": "select",
      "defaultValue": "",
      "instructions": "The country or region where you live.",
      "exposed": true,
      "required": true,
      "restrictions": {},
      "errorMessages": {
        "required": "Select your country or region of residence"
      },
      "options": [
        {
          "value": "",
          "name": "--",
          "default": true
        },
        {
          "value": "AF",
          "name": "Afghanistan",
          "default": false
        },
        {
          "value": "AX",
          "name": "Åland Islands",
          "default": false
        },
        ...
    }
}
```

enterprise_access/apps/provisioning/models.py
```diff
@@ -0,0 +1,191 @@
...
+
+@define
+class GetCreateEnterpriseAdminUsersInput(BaseInputOutput):
+    """
+    The input object to be used for the fetching or creating
+    of enterprise admin users.
+    """
+    KEY = 'create_enterprise_admin_users_input'
+
+    # enterprise_customer_uuid: UUID = field(validator=validators.instance_of(UUID))
```
adamstankiewicz, line 94: [question] if this is no longer needed, should it be removed? Or does it need to be uncommented?

enterprise_access/apps/provisioning/models.py
```diff
@@ -0,0 +1,191 @@
...
+    user_emails: list[str] = field(validator=validators.instance_of(list))
```
adamstankiewicz, line 95: [consideration] Does there need to be any validation this list actually contains valid email addresses vs. just a list of strings?

enterprise_access/apps/workflow/serialization.py
adamstankiewicz: Thanks for the link to the docs. An ADR about "why attrs/cattrs?" would be helpful!

enterprise_access/apps/provisioning/models.py
iloveagent57: Yup, it uses the same exact package to validate the provided country code during registration: https://github.com/openedx/edx-platform/blob/52f2231f2070f857b4802482a068a57d3dbf26ab/openedx/core/djangoapps/user_authn/views/register.py#L24

adamstankiewicz: Great! Thanks for the confirmation and linking to the source :)

enterprise_access/apps/provisioning/models.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def get_create_customer_step(self):
+        return GetCreateCustomerStep.objects.filter(
+            workflow_record_uuid=self.uuid,
+        ).first()
+
+    def get_create_enterprise_admin_users_step(self):
+        return GetCreateEnterpriseAdminUsersStep.objects.filter(
+            workflow_record_uuid=self.uuid,
+        ).first()
```
pwnage101, line 185: Seemingly unused methods.  Remove?

enterprise_access/apps/workflow/serialization.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@CONVERTER.register_structure_hook
+def uuid_structure_hook(val: str, _) -> uuid.UUID:
+    return uuid.UUID(val)
+
+
+@CONVERTER.register_unstructure_hook
+def uuid_unstructure_hook(val: uuid.UUID) -> str:
+    return str(val)
```
pwnage101, line 24: Nit: Since this is at least my first foray into attrs/cattrs and likely others too, maybe add a code comment here stating that we need to teach cattrs how to structure/destructure UUIDs because it happens to not be supported by default: https://catt.rs/en/stable/defaulthooks.html

enterprise_access/apps/provisioning/models.py
pwnage101: I'd argue that email format validation could just be the domain of the DRF serializer layer, which in turn feeds this input pre-cleaned.  That way, most errors are as close to the surface as possible.

pwnage101 (approved):
The rest LGTM

enterprise_access/apps/provisioning/models.py
adamstankiewicz: Makes sense!

enterprise_access/apps/provisioning/models.py
iloveagent57: These are used in the admin classes.

enterprise_access/apps/api/v1/views/provisioning.py
```diff
@@ -50,37 +50,26 @@ def create(self, request, *args, **kwargs):
...
             ) from exc
 
         response_serializer = serializers.ProvisioningResponseSerializer({
-            'enterprise_customer': created_customer,
-            'customer_admins': customer_admins,
+            'enterprise_customer': workflow.customer_output_dict(),
+            'customer_admins': workflow.admin_users_output_dict(),
         })
         return Response(
             response_serializer.data,
```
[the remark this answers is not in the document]
iloveagent57: They'll always create a new workflow model instance, even if that workflow does not actually produce downstream business records like `EnterpriseCustomers`.

PR #672, opened 2025-03-31
fix: [can-redeem] Do not 422 just because one run is restricted

pwnage101:
ENT-10254


enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -2998,4 +3002,97 @@ def test_can_redeem_content_not_in_catalog_service_422_unprocessable(
             query_params = {'content_key': [content_key_for_redemption]}
             response = self.client.get(self.subsidy_access_policy_can_redeem_endpoint, query_params)
 
+        assert response.status_code == status.HTTP_200_OK
```
pwnage101, line 3005: This test case now returns 200 OK instead of 422 UNPROCESSABLE, which should address what we're seeing in the bug ticket.

enterprise_access/apps/api/v1/tests/test_subsidy_access_policy_views.py
```diff
@@ -2998,4 +3002,97 @@ def test_can_redeem_content_not_in_catalog_service_422_unprocessable(
...
+        assert response_list[0]["list_price"] is None
+        assert len(response_list[0]["redemptions"]) == 0
+        assert response_list[0]["has_successful_redemption"] is False
+        assert response_list[0]["redeemable_subsidy_access_policy"] is None
+        assert response_list[0]["can_redeem"] is False
+        assert response_list[0]["reasons"][0]["reason"] == REASON_CONTENT_NOT_IN_CATALOG
+
+    @mock.patch('enterprise_access.apps.content_assignments.api.get_and_cache_content_metadata')
+    @mock.patch('enterprise_access.apps.subsidy_access_policy.subsidy_api.get_and_cache_transactions_for_learner')
+    def test_can_redeem_content_malformed_from_downstream_subsidy_call_422_unprocessable(
```
pwnage101, line 3021: I added two more test cases to cover additional cases where we still return 422 UNPROCESSABLE.

iloveagent57 (approved):

adamstankiewicz (approved):

PR #674, opened 2025-03-31
fix: learner profile error response

jesperhodge:
**Description:**
Fixes a bug where the Learner Portal endpoint does not return error messages in the expected format.

**Important:**
The code was generated with some help of AI, which was reviewed by me. However the review here should also have extra scrutiny that there's no weirdness in the code.

**Jira:**
[ENT-10214](https://2u-internal.atlassian.net/browse/ENT-10214)

**How to test:**
1. test that existing API response has not been broken
    - Grab a suitable enterprise customer and learner
    - Go to /api-docs/
    - Execute the /v1/admin-view/learner_profile/ endpoint with the required arguments
    - Make sure expected results are returned
2. test that error response is now as expected
    - Fabricate HTTP errors when the `EnterpriseAccess` code requests subscriptions, group memberships, and enrollments
        - To do this, you can go to the file enterprise_access/apps/admin_portal_learner_profile/api.py in the repo locally, and in the try block for each method, raise an `requests.exceptions.HTTPError`
    - The correct response per field should look like `  "subscriptions": { "error": "Failed to fetch subscriptions" },`
    - The error message should not just be "error" (that happens if the response is serialized incorrectly)
    - The error message for the "enrollments" field should be a string, not be an array of individual characters
    

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/admin_portal_learner_profile/serializers.py
+++ b/enterprise_access/apps/admin_portal_learner_profile/serializers.py
@@ -40,6 +40,7 @@ class ErrorOrField(serializers.Field):
         return self.base_field.to_representation(value)
 
 
+# pylint: disable=abstract-method
 class AdminLearnerProfileRequestSerializer(serializers.Serializer):
     """Serializer for validating admin learner profile query parameters."""
     user_email = serializers.EmailField(required=True, help_text="The email address of the learner.")
@@ -47,6 +48,7 @@ class AdminLearnerProfileRequestSerializer(serializers.Serializer):
     enterprise_customer_uuid = serializers.UUIDField(required=True, help_text="The UUID of the enterprise customer.")
 
 
+# pylint: disable=abstract-method
 class AdminLearnerProfileResponseSerializer(serializers.Serializer):
     """Serializer for structuring the admin learner profile response."""
     subscriptions = ErrorOrField(
--- a/enterprise_access/apps/admin_portal_learner_profile/serializers.py
+++ b/enterprise_access/apps/admin_portal_learner_profile/serializers.py
@@ -8,7 +8,38 @@ from rest_framework import serializers
 logger = logging.getLogger(__name__)
 
 
-# pylint: disable=abstract-method
+class ErrorOrField(serializers.Field):
+    """
+    A field that wraps an existing field (base_field) to allow either its normal value
+    or an error dictionary in the form {"error": "error text"}.
+    """
+    def __init__(self, base_field, **kwargs):
+        self.base_field = base_field
+        super().__init__(**kwargs)
+
+    def to_internal_value(self, data):
+        '''
+        Validate the input data. If it's an error dictionary, return it as-is.'
+        '''
+        # If the value is an error dictionary, validate its structure.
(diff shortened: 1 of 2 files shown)


enterprise_access/apps/admin_portal_learner_profile/serializers.py
```diff
@@ -8,21 +8,66 @@
...
+        '''
+        # If the value is already an error dict, return as-is.
+        if isinstance(value, dict) and "error" in value:
+            return value
+
+        # Otherwise, let the base field handle serialization.
+        return self.base_field.to_representation(value)
+
+
 class AdminLearnerProfileRequestSerializer(serializers.Serializer):
```
marlonkeating, line 43: nit: I don't know that your change introduced this, but there are several pylint failures for not overriding create/update methods of `BaseSerializer` for the two serializer classes.

> enterprise_access/apps/admin_portal_learner_profile/serializers.py:43:0: W0223: Method 'create' is abstract in class 'BaseSerializer' but is not overridden in child class 'AdminLearnerProfileRequestSerializer' (abstract-method)

enterprise_access/apps/admin_portal_learner_profile/serializers.py
```diff
@@ -8,21 +8,66 @@
 logger = logging.getLogger(__name__)
 
 
-# pylint: disable=abstract-method
+class ErrorOrField(serializers.Field):
```
marlonkeating, line 11: Can we get unit tests for this?

enterprise_access/apps/admin_portal_learner_profile/serializers.py
```diff
@@ -8,21 +8,66 @@
...
 
-# pylint: disable=abstract-method
 class AdminLearnerProfileResponseSerializer(serializers.Serializer):
     """Serializer for structuring the admin learner profile response."""
-    subscriptions = serializers.ListField(help_text="Details of the learner's subscription licenses.", required=False)
-    group_memberships = serializers.ListField(help_text="Flex group memberships for the learner.", required=False)
-    enrollments = serializers.DictField(
-        child=serializers.ListField(),
-        help_text="Course enrollments for the learner.",
+    subscriptions = ErrorOrField(
```
marlonkeating, line 52: Can we get unit tests with error scenarios for the `subscriptions`, `group_memberships`, `enrollments` fields?

marlonkeating (changes requested):
Nice abstraction for field errors!  It just needs some unit tests, as far as I can see.

enterprise_access/apps/admin_portal_learner_profile/serializers.py
jesperhodge: The tests I just added for the error responses for the `subscriptions` and other fields should include this in their test coverage sufficiently.

enterprise_access/apps/admin_portal_learner_profile/serializers.py
jesperhodge: Good catch

marlonkeating (approved):

PR #680, opened 2025-04-10
docs: Self-Service Pricing ADR

pwnage101:
ENT-10076


docs/decisions/0027-self-service-pricing.rst
```diff
@@ -0,0 +1,44 @@
...
+the same time we can't tolerate price discrepancies within the same checkout journey, so
+prices need to propagate to every system very quickly.
+
+Decision
+========
+
+We will define Stripe products and prices via Terraform.  These prices will be made
+available via an unauthenticated billing API endpoint within enterprise-access (URL TBD)
+which dynamically fetches the prices from Stripe.  Frontends, in turn, may call this
+API endpoint to retrieve the price pre-authenticated.
```
adamstankiewicz, line 23: [consideration/clarification] For the checkout frontend in particular, the current thinking is to have a BFF-style API within enterprise-access that returns not only this price information from Stripe, but also provides additional metadata from other services (e.g., the LMS form validation API for its metadata about the form fields related to logistration, etc.).

In this scenario, I don't believe the checkout frontend would be calling this proposed unauthenticated billing API endpoint directly to `GET` price info on initial page load, but rather rely on the underlying functionality (e.g., Python API/helper method, serializers, etc.) to surface the same price information from Stripe, but composed in the BFF-style response rather than doing API orchestration on the checkout frontend.

I'm assuming this BFF-style endpoint is not quite what you're proposing here with the "unauthenticated billing API endpoint", but rather more of a standalone API endpoint?

Not sure if it's worth mentioning the intent of how this might be consumed within such a BFF-style checkout frontend-specific endpoint in this ADR, beyond perhaps a mention of extensibility of the price info beyond just this proposed standalone, API endpoint.

Hopefully this made sense!

pwnage101: Thanks for the clarification. I ultimately want to defer to your proposal in this case.

Is it possible that price information would ever be surfaced prior to the registration step, in which case BFF endpoints would be inaccessible?  If not, then I agree that price (possibly in the form of serialized Stripe Price objects) can simply be included within a BFF endpoint for this service/route.  I'm happy as long as we continue to consolidate Stripe API calls behind a singular python API.

That said, if price will ever be surfaced by whatever marketing page funnels people into the checkout flow (think "Pricing" page), then I _think_ a dedicated unauthenticated prices endpoint would still need to be part of the plan.

adamstankiewicz: > I'm happy as long as we continue to consolidate Stripe API calls behind a singular python API.

Yeah, this makes sense to me; python API can be used for different use cases (BFF API vs. dedicated API), but have the same underlying functionality.

> That said, if price will ever be surfaced by whatever marketing page funnels people into the checkout flow (think "Pricing" page), then I think a dedicated unauthenticated prices endpoint would still need to be part of the plan.

Agreed! I think the dedicated unauthenticated prices endpoint still will play a role for the marketing page price display, assuming we do end up showing price upfront there. 

FWIW, I can't find any current price information displayed on [business.edx.org](https://business.edx.org/) right now, but there is some price displayed on [campus.edx.org](https://campus.edx.org/) (e.g., https://campus.edx.org/subscriptions). I haven't heard/seen anything about design/copy changes for the marketing pages yet, but I think it's a fair assumption we'll likely want to surface price info on the marketing pages, where we'd still need this dedicated prices endpoint.

pwnage101: whoops, I forgot that https://business.edx.org/business/book-demo doesn't actually have pricing info, just a license count drop-down:
Screenshot 2025-04-10 at 2 33 22 PM

docs/decisions/0027-self-service-pricing.rst
```diff
@@ -0,0 +1,53 @@
...
+We will define Stripe products and prices via Terraform.  These prices will be made
+available via a central python API method within enterprise-access which
+dynamically fetches the prices from Stripe.  This python API in turn will drive
+the inclusion of price data within various REST APIs:
+
+* GET https://enterprise-access.edx.org/api/v1/billing/prices
+
+  * Unauthenticated endpoint intended for delivery of pricing to Marketing pages.
+
+* POST https://enterprise-access.edx.org/api/v1/bffs/checkout/checkout
```
adamstankiewicz, line 29: The checkout MFE will primarily utilize an **unauthenticated `GET`** to serve the price information (pre-logistration) and metadata needed for basic client-side form validation of the logistration related form fields on initial page load. A possible API route for the above endpoint serving prices for the checkout MFE might be `/api/v1/bffs/checkout/validation/` or similar.

There would be a separate, **authenticated `POST`** used by the checkout MFE for when the fully validated form is submitted as an authenticated user with the metadata required for Stripe and customer provisioning, perhaps the `/api/v1/bffs/checkout/checkout/` path as you've proposed above. That said, I'm not sure if the `POST` to perform the checkout itself would necessarily need to respond with pricing information itself as we need it before form submission (aka the `GET`).

adamstankiewicz (approved):
Left a comment to clarify the BFF endpoint listed that would serve pricing information, but LGTM otherwise!

docs/decisions/0027-self-service-pricing.rst
```diff
@@ -0,0 +1,104 @@
...
+available via a central python API method within enterprise-access which
+dynamically fetches the prices from Stripe, possibly internationalizes it, then
+serializes it in a way that is consumable by frontends.  This python API in
+turn will drive the inclusion of price data within various REST APIs:
+
+* GET https://enterprise-access.edx.org/api/v1/billing/prices
+
+  * Unauthenticated endpoint intended for delivery of pricing information to marketing pages.
+
+* GET https://enterprise-access.edx.org/api/v1/bffs/checkout/checkout
```
adamstankiewicz, line 30: Looking back at https://github.com/openedx/enterprise-access/pull/680#discussion_r2047979871, I think the endpoint to serve the pricing information (and other logistration field validation rules, etc.) as an unauthenticated user would be:

```
https://enterprise-access.edx.org/api/v1/bffs/checkout/validation/
```

I feel the `checkout` endpoint listed here would be the authenticated `POST` to submit the form / checkout session.

adamstankiewicz (approved):
LGTM, with just one comment about the checkout MFE validation endpoint route.

PR #684, opened 2025-04-17
fix: ensure get_assignments_serializer is annotated as a list of assignments

adamstankiewicz:
**Description:**

https://enterprise-access.edx.org/api/schema/redoc/#tag/Subsidy-Access-Policy-Redemption/operation/api_v1_policy_redemption_credits_available_list

The `learner_content_assignments` property within the credits-available API response is annotated as a single assignment object vs. a list of assignments:

image

As such, the auto-generated TypeScript definitions within Learner Portal, consuming this service's Open API schema are incorrectly suggesting `learnerContentAssignments` is an object vs. an array, blocking the ability to properly type a service function.

**Jira:**

N/A

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


brobro10000 (approved):
Inspiring 🥇

PR #686, opened 2025-04-22
fix: [assignment email] show actual course start date for courses which have not started

pwnage101:
Emails to learners about course assignments previously displayed the current date as the course "start date" in this case:

* course is self-paced.
* end date is far enough in the future that there's enough time to complete.
* start date is in the future.

Displaying the current date rather than the actual course start date in this specific case was unintentional, and does not match the corresponding logic in the learner-portal.

ENT-10263

### DIFF
--- a/enterprise_access/apps/content_assignments/utils.py
+++ b/enterprise_access/apps/content_assignments/utils.py
@@ -10,47 +10,57 @@ from enterprise_access.apps.content_assignments.constants import (
     BRAZE_TIMESTAMP_FORMAT,
     START_DATE_DEFAULT_TO_TODAY_THRESHOLD_DAYS
 )
+from enterprise_access.utils import localized_utcnow
 
 
-def is_within_minimum_start_date_threshold(curr_date, start_date):
+def is_within_minimum_start_date_threshold(
+    curr_date: datetime,
+    start_date: datetime,
+) -> bool:
     """
     Checks if today's date were set to a certain number of days in the past,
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/content_assignments/utils.py
```diff
@@ -44,13 +53,15 @@ def get_self_paced_normalized_start_date(start_date, end_date, course_metadata):
...
     pacing_type = course_metadata.get('pacing_type', {}) or None
     weeks_to_complete = course_metadata.get('weeks_to_complete', {}) or None
     if not (start_date and end_date and pacing_type and weeks_to_complete):
         return curr_date.strftime(BRAZE_TIMESTAMP_FORMAT)
-    if pacing_type == "self_paced":
-        if has_time_to_complete(curr_date, end_date, weeks_to_complete) or \
-                is_within_minimum_start_date_threshold(curr_date, start_date):
+    start_date_datetime = parser.parse(start_date).astimezone(UTC)
+    end_date_datetime = parser.parse(end_date).astimezone(UTC)
+    if pacing_type == "self_paced" and start_date_datetime < curr_date:
```
pwnage101, line 63: This one line is the meat of the fix, the rest are code cleanup / readability improvement.

brobro10000 (approved):
Great job refactoring to `parser.parse` for the datetime conversion 👍🏽

adamstankiewicz (approved):

PR #687, opened 2025-04-23
feat: Register Learner Credit Request models in admin

zamanafzal:
**Description:**
Register the `LearnerCreditRequestConfiguration` and `LearnerCreditRequest` model in admin. 

**Jira:**
ENT-10292

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/subsidy_request/admin.py
+++ b/enterprise_access/apps/subsidy_request/admin.py
@@ -165,3 +165,76 @@ class SubsidyRequestCustomerConfigurationAdmin(DjangoQLSearchMixin, admin.ModelA
         obj.changed_by = current_user
 
         super().save_model(request, obj, form, change)
+
+
+@admin.register(models.LearnerCreditRequest)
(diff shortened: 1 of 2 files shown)


sameeramin (approved):
LGTM 👍🏼

sameenfatima78 (approved):

enterprise_access/apps/subsidy_request/admin.py
```diff
@@ -165,3 +165,76 @@ def save_model(self, request, obj, form, change):
...
+    """ Admin configuration for the LearnerCreditRequestConfiguration model. """
+
+    list_display = (
+        'uuid',
+        'active',
+        'created',
+        'modified',
+    )
+
+    search_fields = ('uuid',)
```
sameenfatima78, line 218: nit: can we also include learner email here?

zamanafzal: I think no, as this is the `LearnerCreditRequestConfiguration` model and doesn't have an email field.

PR #689, opened 2025-04-28
feat: Create CRUD APIs for LearnerCreditRequest

ahmad-hassan-dev:
**Description:**
Added the `LearnerCreditRequestViewSet`. The new viewset follows the same pattern as existing subsidy request viewsets (License and Coupon Code) while adding specific functionality for Learner Credit requests.

Key features:

- Creates a new LearnerCreditRequestViewSet extending SubsidyRequestViewSet
- Adds validation specific to Learner Credit requests:
- - Validates policy UUID and checks if Browse & Request is active for the policy
- - Prevents duplicate requests for the same course and policy
- Includes proper permission checks and enterprise filtering
- Adds OpenAPI schema documentation for all endpoints

The viewset supports:

> Creating new Learner Credit requests
> Listing/retrieving requests (with enterprise filtering)
> Getting request overview counts by state

**Jira:**
[ENT-10182 - Create CRUD APIs for LearnerCreditRequest](https://2u-internal.atlassian.net/browse/ENT-10182)


enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +691,126 @@ def decline(self, *args, **kwargs):
...
+)
+class LearnerCreditRequestViewSet(SubsidyRequestViewSet):
+    """
+    Viewset for learner credit requests.
+    """
+
+    queryset = LearnerCreditRequest.objects.order_by("-created")
+    serializer_class = serializers.LearnerCreditRequestSerializer
+
+    subsidy_type = SubsidyTypeChoices.LEARNER_CREDIT
```
ahmad-hassan-dev, line 720: `subsidy_type` is sourced from `SubsidyRequestCustomerConfiguration`, but we are not proceeding with that approach. Instead, we will rely on `LearnerCreditRequestConfiguration`. Therefore, I will be removing `subsidy_type` and its related code from the viewset.

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +691,126 @@ def decline(self, *args, **kwargs):
...
+
+    def _validate_subsidy_request(self):
+        """
+        Validate request creation:
+        - Inherit basic validation from parent.
+        - Ensure policy_uuid is provided and valid.
+        - Ensure Browse & Request is active for the policy.
+        - No duplicate PENDING/REQUESTED requests for the same course and policy.
+        """
+        super()._validate_subsidy_request()
```
ahmad-hassan-dev, line 747: Since the parent method contains conditions related to SubsidyRequestCustomerConfiguration, I will remove the parent method call. This will allow us to add only the conditions related to LearnerCreditRequest.

enterprise_access/apps/api/serializers/subsidy_requests.py
```diff
@@ -121,3 +123,34 @@ def update(self, instance, validated_data):
...
+class LearnerCreditRequestSerializer(SubsidyRequestSerializer):
+    """
+    Serializer for the `LearnerCreditRequest` model.
+    """
+
+    learner_credit_request_config = serializers.PrimaryKeyRelatedField(
+        queryset=LearnerCreditRequestConfiguration.objects.all(),
+        required=False,
+        allow_null=True,
+    )
```
sameenfatima78, line 148: Question: Curious why we added this field, given that we'll only be getting policy uuid in the request.

enterprise_access/apps/api/v1/views/browse_and_request.py
sameenfatima78, line 747: nit: this check can be removed for LC requests as we discussed.

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +691,126 @@ def decline(self, *args, **kwargs):
...
+        ).exists():
+            error_msg = (
+                f"You already have an active learner credit request for course {course_id} "
+                f"under policy UUID: {policy_uuid}."
+            )
+            raise SubsidyRequestCreationError(
+                error_msg, status.HTTP_422_UNPROCESSABLE_ENTITY
+            )
+
+    def create(self, request, *args, **kwargs):
```
sameenfatima78, line 791: nit: add necessary permission on this method.

sameenfatima78 (commented):
Looks good, left some comments

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +692,127 @@ def decline(self, *args, **kwargs):
...
+        queryset = super().get_queryset()
+        enterprise_uuid = self.request.query_params.get("enterprise_customer_uuid")
+
+        # Non-admin users can only see their own requests
+        if not (self.request.user.is_staff or self.request.user.is_superuser):
+            queryset = queryset.filter(user=self.request.user)
+
+        # Filter by enterprise if provided
+        if enterprise_uuid:
+            queryset = queryset.filter(enterprise_customer_uuid=enterprise_uuid)
```
zamanafzal, line 736: I think enterprise_uuid should be a required field as we want to show the `LearnerCreditRequests` for a specific enterprise to the enterprise admins.

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +692,127 @@ def decline(self, *args, **kwargs):
...
+        """
+        policy_uuid = self.request.data.get("policy_uuid")
+        if not policy_uuid:
+            raise SubsidyRequestCreationError(
+                "policy_uuid is required.", status.HTTP_400_BAD_REQUEST
+            )
+
+        try:
+            policy = SubsidyAccessPolicy.objects.get(uuid=policy_uuid)
+            self.request.validated_policy = (
```
zamanafzal, line 755: Can we please make it more readable like this
`self.request.validated_policy = policy`  # Store validated policy for use in create

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +692,127 @@ def decline(self, *args, **kwargs):
...
+                policy  # Store validated policy for use in create
+            )
+        except SubsidyAccessPolicy.DoesNotExist:
+            raise SubsidyRequestCreationError(
+                f"Invalid policy_uuid: {policy_uuid}.", status.HTTP_400_BAD_REQUEST
+            )
+
+        if (
```
zamanafzal, line 763: nit: We already have a property `bnr_enabled` in `SubsidyAccessPolicy` model   for this.

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +692,127 @@ def decline(self, *args, **kwargs):
...
+            not policy.learner_credit_request_config
+            or not policy.learner_credit_request_config.active
+        ):
```
zamanafzal, line 766: Do we need to check the `policy.retired` here?

enterprise_access/apps/api/serializers/subsidy_requests.py
```diff
@@ -121,3 +123,34 @@ def update(self, instance, validated_data):
...
+
+class LearnerCreditRequestConfigurationSerializer(serializers.ModelSerializer):
+    """
+    Serializer for the `LearnerCreditRequestConfiguration` model.
+    """
+
+    class Meta:
+        model = LearnerCreditRequestConfiguration
+        fields = "__all__"
+        read_only_fields = ["uuid", "active", "created", "modified"]
```
zamanafzal, line 136: As we plan to allow activating or deactivating LearnerCreditRequestConfiguration via the admin portal in future, I think the active field should not be included in read_only_fields.

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: Done.

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: Done.

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: Done.

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: Done.

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: Most of the checks will be handled in the `can_request` method, so they don’t apply here.

enterprise_access/apps/api/serializers/subsidy_requests.py
ahmad-hassan-dev: Done.

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: Done.

enterprise_access/apps/api/serializers/subsidy_requests.py
sameenfatima78: good catch!

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +692,105 @@ def decline(self, *args, **kwargs):
...
+    serializer_class = serializers.LearnerCreditRequestSerializer
+
+    subsidy_type = SubsidyTypeChoices.LEARNER_CREDIT
+
+    def _validate_subsidy_request(self):
+        """
+        Validate request creation:
+        - Ensure policy_uuid is provided and valid.
+        - Ensure Browse & Request is active for the policy.
+        - No duplicate PENDING/REQUESTED requests for the same course and policy.
```
sameenfatima78, line 728: I think these two checks are already going to be covered in "can_request" as we'll need to handle the logic on frontend to show appropriate CTA. Maybe we can skip these validations here?

sameenfatima78 (approved):
Looks good, just one small suggestion.

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: I believe we should keep these checks in place—they help make our API calls safer by preventing record creation with an invalid policy and avoiding duplicate entries.

Additionally, this maintains consistency across the codebase, as similar duplication checks exist in other viewsets. In the parent validation method, there's also a check to ensure that the configuration for the policy is enabled, done through the subsidy_request's validation logic. Since we're not using SubsidyRequestCustomerConfiguration here and instead have a separate LearnerCreditRequestConfiguration, we’re not calling that logic, but added the same check here.

enterprise_access/apps/api/serializers/subsidy_requests.py
```diff
@@ -121,3 +123,34 @@ def update(self, instance, validated_data):
...
+    learner_credit_request_config = serializers.PrimaryKeyRelatedField(
+        queryset=LearnerCreditRequestConfiguration.objects.all(),
+        required=False,
+        allow_null=True,
+    )
+
+    class Meta:
+        model = LearnerCreditRequest
+        fields = SubsidyRequestSerializer.Meta.fields + [
+            "learner_credit_request_config"
```
zamanafzal, line 153: Noticed the `assignment` isn’t included in the serializer — was this intentional?

zamanafzal (approved):
LGTM

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -690,6 +692,105 @@ def decline(self, *args, **kwargs):
...
+        tags=['Learner Credit Requests'],
+        summary='Learner credit request overview.',
+    ),
+)
+class LearnerCreditRequestViewSet(SubsidyRequestViewSet):
+    """
+    Viewset for learner credit requests.
+    """
+
+    queryset = LearnerCreditRequest.objects.order_by("-created")
```
zamanafzal, line 718: One more question: Should we show the learner requests with the status errored or expired?

ahmad-hassan-dev: I think we should return the LearnerCreditRequest even if its status is error or expired. According to the flowchart, if an admin attempts to approve a request and an error occurs during that process, the request status should be updated to error. However, the Approve button should still be visible to the admin for retrying the action.

At the same time, if a LearnerCreditRequest is in the error state, we still want the learner to see it as if it's in the requested state from their perspective. We can handle this by mapping the error status to requested when presenting the data to the learner. This way, even if the request failed internally, the learner will still see it as pending, while the admin is informed that an error occurred and can take corrective action.

enterprise_access/apps/api/serializers/subsidy_requests.py
ahmad-hassan-dev: Initially yes, But Adding it make more sense so added that into the serializer.

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: For now, as discussed, we'll include data with the 'expired' status in the response.

PR #696, opened 2025-05-08
feat: expose `display_reason` derived from `reasons`

brobro10000:
**Description:**
Exposes a net new field from the `SubsidyAccessPolicyCanRedeemElementResponseSerializer`, `display_reason`. The intent of this new field is to isolate a failed redemption reason to the client without naively picking the zero indexed value. 

The logic within `evaluate_policies` was also updated to bring sorting logic implemented downstream, higher in the call stack to correctly sort policies for both redeemable and non-redeemable policies.

Can-redeem response:
Screenshot 2025-05-09 at 1 01 30 PM

**Jira:**
[ENT-10268](https://2u-internal.atlassian.net/browse/ENT-10268)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/serializers/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/serializers/subsidy_access_policy.py
@@ -694,6 +694,10 @@ class SubsidyAccessPolicyCanRedeemElementResponseSerializer(serializers.Serializ
             "List of reasons why each of the enterprise's subsidy access policies are not redeemable, grouped by reason"
         )
     )
+    display_reason = serializers.CharField(
+        allow_null=True,
+        help_text="A single, user-facing description of the most salient reason for non-redeemability.",
+    )
 
 
 # pylint: disable=abstract-method
--- a/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
+++ b/enterprise_access/apps/api/v1/views/subsidy_access_policy.py
@@ -70,6 +70,7 @@ from enterprise_access.apps.subsidy_access_policy.subsidy_api import (
     get_and_cache_subsidy_learners_aggregate_data,
     get_redemptions_by_content_and_policy_for_learner
 )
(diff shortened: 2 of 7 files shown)


enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -694,6 +694,10 @@ class SubsidyAccessPolicyCanRedeemElementResponseSerializer(serializers.Serializ
             "List of reasons why each of the enterprise's subsidy access policies are not redeemable, grouped by reason"
         )
     )
+    display_reason = serializers.CharField(
+        allow_null=True,
+        help_text="A single reason user text derived from 'reasons' to be displayed in the interface",
```
iloveagent57, line 699: Suggested alternative: "A single, user-facing description of the most salient reason for non-redeemability."

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -500,8 +505,14 @@ def evaluate_policies(self, enterprise_customer_uuid, lms_user_id, content_key,
         """
         redeemable_policies = []
         non_redeemable_policies = defaultdict(list)
-        all_policies_for_enterprise = self.get_queryset()
-        for policy in all_policies_for_enterprise:
+        # Sort policies by:
+        # - priority (of type)
+        # - expiration, sooner to expire first
+        # - balance, lower balance first
+        all_sorted_policies_for_enterprise = sort_subsidy_access_policies_for_redemption(
```
iloveagent57, line 512: nit: consider adding a note to the docstring that the redeemable policies returned by this method are sorted in a meaningful way.

iloveagent57 (approved):

adamstankiewicz (approved):

PR #705, opened 2025-05-19
feat: expand `display_reason` to include addtl metadata

brobro10000:
**Description:**
Changes `display_reason` to return a the minimal metadata required from `reasons` to be usable by the client. 

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


iloveagent57 (approved):

enterprise_access/apps/api/serializers/subsidy_access_policy.py
```diff
@@ -672,17 +672,24 @@ def get_group_associations(self, obj):
...
-    can_redeem endpoint.
+    Response serializer used to document the structure of a "display_reason" highlighting a single reason
+    for policy non-redeemability
     """
     reason = serializers.CharField(
         help_text="reason code (in camel_case) for why the following subsidy access policies are not redeemable."
     )
     user_message = serializers.CharField(
         help_text="Description of why the following subsidy access policies are not redeemable."
     )
```
pwnage101, line 685: Why not just return a SubsidyAccessPolicyCanRedeemReasonResponseSerializer under `display_reason`?  The same frontend code and backend view code should work, and we'll at least get a clue as to which policy was selected for the display_reason.

brobro10000: Updated 👍🏽

pwnage101 (approved):

enterprise_access/apps/api/v1/views/subsidy_access_policy.py
```diff
@@ -186,8 +186,9 @@ def _get_reasons_for_no_redeemable_policies(enterprise_customer_uuid, non_redeem
             },
             "policy_uuids": [policy.uuid for policy in policies],
         })
-        if not display_reason:
-            display_reason = user_message
+
+    if len(reasons) > 0:
+        display_reason = reasons[0]
```
pwnage101, line 191: Nit: arbitrarily zero-indexing without a code comment feels like bad practice.

PR #736, opened 2025-06-12
feat: added Decline endpoint in a LearnerCreditRequestViewset

jajjibhai008:
**Description:**

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/v1/views/browse_and_request.py
+++ b/enterprise_access/apps/api/v1/views/browse_and_request.py
@@ -5,8 +5,8 @@ import logging
 
 from celery import chain
 from django.conf import settings
-from django.core.exceptions import ObjectDoesNotExist
-from django.db import transaction
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
+from django.db import DatabaseError, IntegrityError, transaction
 from django.db.models import Count
 from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import extend_schema, extend_schema_view
@@ -45,12 +45,23 @@ from enterprise_access.apps.api_client.ecommerce_client import EcommerceApiClien
 from enterprise_access.apps.api_client.license_manager_client import LicenseManagerApiClient
 from enterprise_access.apps.core import constants
 from enterprise_access.apps.subsidy_access_policy.models import SubsidyAccessPolicy
-from enterprise_access.apps.subsidy_request.constants import SegmentEvents, SubsidyRequestStates, SubsidyTypeChoices
+from enterprise_access.apps.subsidy_request.constants import (
+    LearnerCreditRequestActionErrorReasons,
+    SegmentEvents,
+    SubsidyRequestStates,
+    SubsidyTypeChoices
+)
 from enterprise_access.apps.subsidy_request.models import (
     CouponCodeRequest,
     LearnerCreditRequest,
+    LearnerCreditRequestActions,
     LicenseRequest,
     SubsidyRequestCustomerConfiguration
+)
+from enterprise_access.apps.subsidy_request.utils import (
+    get_action_choice,
+    get_error_reason_choice,
+    get_user_message_choice
 )
 from enterprise_access.apps.track.segment import track_event
 from enterprise_access.utils import get_subsidy_model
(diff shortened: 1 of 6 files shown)


enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -797,6 +812,108 @@ def create(self, request, *args, **kwargs):
...
+
+        # Handle post-decline operations
+        serialized_request = serializers.LearnerCreditRequestSerializer(learner_credit_request).data
+        lms_user_id = serialized_request["lms_user_id"]
+
+        if send_notification:
+            logger.info(
+                f"TODO: Send decline notification email for learner credit request {learner_credit_request_uuid}"
+            )
+        if disassociate_from_org:
```
jajjibhai008, line 900: I didn't write this disassociation functionality inside atomic block because we are not doing so for coupon codes decline functionality.

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -797,6 +812,108 @@ def create(self, request, *args, **kwargs):
...
+        try:
+            learner_credit_request = LearnerCreditRequest.objects.get(uuid=learner_credit_request_uuid)
+        except LearnerCreditRequest.DoesNotExist:
+            return Response(
+                f"Learner Credit Request with UUID {learner_credit_request_uuid} not found.",
+                status=status.HTTP_404_NOT_FOUND,
+            )
+
+        # State validation with action logging
+        if learner_credit_request.state not in [SubsidyRequestStates.REQUESTED]:
```
zamanafzal, line 854: Why are we checking the state here?

jajjibhai008: Just a double check to avoid any anomaly.

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -797,6 +812,108 @@ def create(self, request, *args, **kwargs):
...
+
+        # Perform the decline operation
+        try:
+            with transaction.atomic():
+                learner_credit_request.decline(self.user)
+                LearnerCreditRequestActions.create_action(
+                    learner_credit_request=learner_credit_request,
+                    recent_action=get_action_choice(SubsidyRequestStates.DECLINED),
+                    status=get_user_message_choice(SubsidyRequestStates.DECLINED),
+                )
```
sameenfatima78, line 877: We are calling this block multiple times across different branches in the code.

How about we create the action instance before entering the atomic block and then in case of an error simply update that same instance with the error info?

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -797,6 +812,108 @@ def create(self, request, *args, **kwargs):
...
+                "You must provide a subsidy request UUID to be declined.", status=status.HTTP_400_BAD_REQUEST
+            )
+
+        try:
+            validate_uuid(learner_credit_request_uuid)
+        except ParseError:
+            return Response(
+                f"Subsidy Request UUID provided ({learner_credit_request_uuid}) is not a valid UUID",
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
sameenfatima78, line 842: we can handle these checks in a serializer

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -797,6 +812,108 @@ def create(self, request, *args, **kwargs):
...
+                status=get_user_message_choice(SubsidyRequestStates.REQUESTED),
+                error_reason=get_error_reason_choice(LearnerCreditRequestActionErrorReasons.FAILED_DECLINE),
+                traceback=f"Cannot decline request in state: {learner_credit_request.state}",
+            )
+
+            return Response(
+                f"Learner Credit Request with UUID {learner_credit_request_uuid} cannot be declined. "
+                f"Current state: {learner_credit_request.state}",
+                status=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            )
```
sameenfatima78, line 867: we can skip this case as it is already going to be handled on frontend

jajjibhai008: It was just a double check to avoid any anomaly.

sameenfatima78 (approved):

PR #751, opened 2025-06-20
feat: added cancel api endpoint for learner credit request

omar-sarfraz:
**Description:**
Created the cancel api for learner credit request. It uses the existing `cancel_assignments` function to cancel the associated assignment and then updates the request state.

**Jira:**
[ENT-10357](https://2u-internal.atlassian.net/browse/ENT-10357)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/v1/views/browse_and_request.py
+++ b/enterprise_access/apps/api/v1/views/browse_and_request.py
@@ -43,6 +43,7 @@ from enterprise_access.apps.api.utils import (
 )
 from enterprise_access.apps.api_client.ecommerce_client import EcommerceApiClient
 from enterprise_access.apps.api_client.license_manager_client import LicenseManagerApiClient
+from enterprise_access.apps.content_assignments import api as assignments_api
 from enterprise_access.apps.core import constants
 from enterprise_access.apps.subsidy_access_policy.api import approve_learner_credit_request_via_policy
 from enterprise_access.apps.subsidy_access_policy.exceptions import SubisidyAccessPolicyRequestApprovalError
@@ -732,6 +733,10 @@ class CouponCodeRequestViewSet(SubsidyRequestViewSet):
         summary='Decline a learner credit request.',
         request=serializers.LearnerCreditRequestDeclineSerializer,
     ),
+    cancel=extend_schema(
+        tags=['Learner Credit Requests'],
+        summary='Learner credit request cancel endpoint.',
+    )
 )
 class LearnerCreditRequestViewSet(SubsidyRequestViewSet):
     """
@@ -901,6 +906,66 @@ class LearnerCreditRequestViewSet(SubsidyRequestViewSet):
             lc_request_action.traceback = format_traceback(exc)
             lc_request_action.save()
             return Response({"detail": error_msg}, exc.status_code)
+
+    @permission_required(
+        constants.REQUESTS_ADMIN_ACCESS_PERMISSION,
+        fn=get_enterprise_uuid_from_request_data,
+    )
+    @action(
+        detail=False,
+        url_path='cancel',
+        methods=['post'],
+        serializer_class=serializers.LearnerCreditRequestCancelSerializer
+    )
+    def cancel(self, request, *args, **kwargs):
+        """
+        Cancel a learner credit request.
+        """
+        serializer = serializers.LearnerCreditRequestCancelSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        learner_credit_request = serializer.get_learner_credit_request()
+
+        error_msg = None
+        lc_action = LearnerCreditRequestActions.create_action(
+            learner_credit_request=learner_credit_request,
+            recent_action=get_action_choice(SubsidyRequestStates.CANCELLED),
+            status=get_user_message_choice(SubsidyRequestStates.CANCELLED),
+        )
+
+        try:
+            with transaction.atomic():
+                response = assignments_api.cancel_assignments([learner_credit_request.assignment], False)
+                if response.get('non_cancelable'):
+                    error_msg = (
+                        f"Failed to cancel associated assignment with uuid: {learner_credit_request.assignment.uuid}"
+                        f" for request: {learner_credit_request.uuid}."
+                    )
+                    lc_action.error_reason = get_error_reason_choice(
+                        LearnerCreditRequestActionErrorReasons.FAILED_CANCELLATION
+                    )
+                    lc_action.status = get_user_message_choice(SubsidyRequestStates.APPROVED)
+                    lc_action.traceback = error_msg
+                    lc_action.save()
+                    return Response(error_msg, status=status.HTTP_422_UNPROCESSABLE_ENTITY)
+
+                learner_credit_request.cancel(self.user)
+                lc_action.save()
+
+                # TODO: add logic to send cancellation email
+
+            serialized_request = serializers.LearnerCreditRequestSerializer(learner_credit_request).data
+            return Response(serialized_request, status=status.HTTP_200_OK)
+        except (ValidationError, IntegrityError, DatabaseError) as exc:
+            error_msg = format_traceback(exc)
+            logger.exception(error_msg)
+            lc_action.error_reason = get_error_reason_choice(
+                LearnerCreditRequestActionErrorReasons.FAILED_CANCELLATION
+            )
+            lc_action.status = get_user_message_choice(SubsidyRequestStates.APPROVED)
+            lc_action.traceback = error_msg
+            lc_action.save()
+            return Response(status=status.HTTP_422_UNPROCESSABLE_ENTITY)
 
     @permission_required(
         constants.REQUESTS_ADMIN_ACCESS_PERMISSION,
--- a/enterprise_access/apps/subsidy_request/models.py
+++ b/enterprise_access/apps/subsidy_request/models.py
@@ -386,6 +386,12 @@ class LearnerCreditRequest(SubsidyRequest):
         self.reviewed_at = localized_utcnow()
         self.save()
 
+    def cancel(self, reviewer):
+        self.state = SubsidyRequestStates.CANCELLED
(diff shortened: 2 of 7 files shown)


enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -797,6 +816,148 @@ def create(self, request, *args, **kwargs):
...
+            return Response("Learner credit request not found.", status=status.HTTP_404_NOT_FOUND)
+
+        error_msg = None
+        lc_action = LearnerCreditRequestActions.create_action(
+            learner_credit_request=request,
+            recent_action=get_action_choice(SubsidyRequestStates.CANCELLED),
+            status=get_user_message_choice(SubsidyRequestStates.APPROVED),
+            error_reason=get_error_reason_choice(LearnerCreditRequestActionErrorReasons.FAILED_CANCELLATION),
+            traceback=error_msg,
+        )
```
sameenfatima78, line 853: [question] why are we creating an action with error reason and traceback here?

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -797,6 +816,148 @@ def create(self, request, *args, **kwargs):
...
+
+        if request.state != SubsidyRequestStates.APPROVED:
+            error_msg = "Learner credit request cannot be cancelled because it is not in an APPROVED state."
+            lc_action.traceback = error_msg
+            lc_action.save()
+            return Response(error_msg, status=status.HTTP_400_BAD_REQUEST)
```
sameenfatima78, line 859: This should be handled on frontend

enterprise_access/apps/subsidy_request/models.py
```diff
@@ -386,6 +386,10 @@ def decline(self, reviewer, reason=None):
         self.reviewed_at = localized_utcnow()
         self.save()
 
+    def cancel(self):
+        self.state = SubsidyRequestStates.CANCELLED
```
sameenfatima78, line 390: shouldn't we update the reviewed_at field as well?

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -797,6 +816,148 @@ def create(self, request, *args, **kwargs):
...
+
+                request.cancel()
+                # TODO: add logic to send cancellation email
+
+                lc_action.status = get_user_message_choice(SubsidyRequestStates.CANCELLED)
+                lc_action.error_reason = None
+                lc_action.traceback = None
+                lc_action.save()
+
+            return Response(status=status.HTTP_200_OK)
```
sameenfatima78, line 881: In all Response statements, we should be sending the information of the request that was processed. We'd need that info to handle frontend checks like showing failure state, etc.

sameenfatima78 (changes requested):

enterprise_access/apps/api/v1/views/browse_and_request.py
omar-sarfraz: `error_reason` remains same for all the failure cases so I've added it once here. `traceback` is optional as we are updating it for all failure cases but it should not create any problem as we are setting both these values to `None` for successful cancellation.

enterprise_access/apps/api/v1/views/browse_and_request.py
omar-sarfraz: I thought it's good to have this validation in backend as well. Should I remove it?

enterprise_access/apps/api/v1/views/browse_and_request.py
omar-sarfraz: It was handled like this in cancel assignments api endpoint. Wouldn't the frontend work the same way as assignments for these requests?

enterprise_access/apps/api/v1/views/browse_and_request.py
sameenfatima78: It's safe to remove it since we will only show Cancel CTA on an approved request.

enterprise_access/apps/api/v1/views/browse_and_request.py
sameenfatima78: @jajjibhai008 can add more to this since he's working on frontend

enterprise_access/apps/api/v1/views/browse_and_request.py
sameenfatima78: Yes, but it should only be populated when we encounter an error. Otherwise, we should leave these two fields null because on the frontend side, the condition to show failed chip is dependent on the value we reason in error_reason. Again, I would defer it to @jajjibhai008 to share his input here.

omar-sarfraz: I've updated the code to only set these 2 values in case of an error.

enterprise_access/apps/api/v1/views/browse_and_request.py
jajjibhai008: @omar-sarfraz it's better to return the updated object as well.

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -902,6 +903,73 @@ def approve(self, request, *args, **kwargs):
             lc_request_action.save()
             return Response({"detail": error_msg}, exc.status_code)
 
+    @extend_schema(
+        tags=['Learner Credit Requests'],
+        summary='Learner credit request cancel endpoint.',
+    )
```
sameenfatima78, line 909: to keep things consistent, let's move this to the top of the viewset (where other endpoints are defined)

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -902,6 +903,73 @@ def approve(self, request, *args, **kwargs):
...
+    @permission_required(
+        constants.REQUESTS_ADMIN_ACCESS_PERMISSION,
+        fn=get_enterprise_uuid_from_request_data,
+    )
+    @action(
+        detail=False,
+        url_path='cancel',
+        methods=['post'],
+        serializer_class=serializers.LearnerCreditRequestCancelSerializer
```
sameenfatima78, line 918: this too

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -902,6 +903,73 @@ def approve(self, request, *args, **kwargs):
...
+    )
+    def cancel(self, request, *args, **kwargs):
+        """
+        Cancel a learner credit request.
+        """
+        serializer = serializers.LearnerCreditRequestCancelSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        request_uuid = serializer.data['request_uuid']
+
+        request = self.get_queryset().filter(uuid=request_uuid).first()
```
sameenfatima78, line 928: this can be written as 
```python
self.get_queryset().get(uuid=request_uuid)

File: enterprise_access/apps/api/v1/views/browse_and_request.py [hunk 8]
```diff
@@ -902,6 +903,73 @@ def approve(self, request, *args, **kwargs):
...
+
+        if not request:
+            return Response("Learner credit request not found.", status=status.HTTP_404_NOT_FOUND)
+
+        error_msg = None
+        lc_action = LearnerCreditRequestActions.create_action(
+            learner_credit_request=request,
+            recent_action=get_action_choice(SubsidyRequestStates.CANCELLED),
+            status=get_user_message_choice(SubsidyRequestStates.APPROVED),
```
sameenfatima78, line 937: status should be set to CANCELLED

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -902,6 +903,73 @@ def approve(self, request, *args, **kwargs):
...
+            serialized_request = serializers.LearnerCreditRequestSerializer(request).data
+            return Response(serialized_request, status=status.HTTP_200_OK)
+        except (ValidationError, IntegrityError, DatabaseError) as exc:
+            error_msg = f"Error cancelling learner credit request with uuid: {request_uuid}: {str(exc)}"
+            logger.exception(error_msg)
+            lc_action.error_reason = get_error_reason_choice(
+                LearnerCreditRequestActionErrorReasons.FAILED_CANCELLATION
+            )
+            lc_action.traceback = error_msg
+            lc_action.save()
```
sameenfatima78, line 970: this is the point where lc_action.status should be changed back to `APPROVED`

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -902,6 +903,73 @@ def approve(self, request, *args, **kwargs):
...
+
+            serialized_request = serializers.LearnerCreditRequestSerializer(request).data
+            return Response(serialized_request, status=status.HTTP_200_OK)
+        except (ValidationError, IntegrityError, DatabaseError) as exc:
+            error_msg = f"Error cancelling learner credit request with uuid: {request_uuid}: {str(exc)}"
+            logger.exception(error_msg)
+            lc_action.error_reason = get_error_reason_choice(
+                LearnerCreditRequestActionErrorReasons.FAILED_CANCELLATION
+            )
+            lc_action.traceback = error_msg
```
sameenfatima78, line 969: nit: use `format_traceback(exc)`

enterprise_access/apps/api/v1/views/browse_and_request.py
omar-sarfraz: It's necessary to define the serializer class here because the `api-docs` display the whole LearnerCreditRequest schema as input to the api endpoint if we don't define the serializer class here. By defining it here, we ensure that the `api-docs` only show `request_uuid` as expected input.

If serializer class is defined on here api itself: 
image

If serializer is defined on the viewset level:
image

sameenfatima78 (approved):

PR #756, opened 2025-06-23
feat: add slug reservation mechanism

iloveagent57:
# Add Enterprise Slug Reservation System for Self-Service Checkout

## Problem
During the delay between creating a Stripe checkout session and actually provisioning the EnterpriseCustomer (via Stripe → Salesforce → Provisioning API), multiple users could reserve the same enterprise slug, leading to downstream failures.

This could happen with:
- Eager admins double-clicking submit buttons
- Frontend bugs causing duplicate requests
- Bad actors intentionally spamming the same slug

### Django Model (`EnterpriseSlugReservation`)
- **One reservation per user** - prevents spam from individual users, but the update-or-create business logic allows users to replace their own reservation, preventing them from getting locked out of the flow.
- **Unique slug constraint** - prevents conflicts between different users  
- **Expiration that matches stripe checkout expiry** - automatically cleans up abandoned reservations
- **Stripe session tracking** - links reservations to checkout sessions for debugging

### API Integration
- Updated `create_free_trial_checkout_session()` to reserve slugs before creating Stripe sessions
- Enhanced validation to check for existing reservations (excluding user's own)
- Backwards compatibility - can work with or without user parameter
- Stores reservation ID in Stripe metadata for tracking

### Management Command
- `cleanup_slug_reservations` command for automated cleanup
- Supports dry-run mode for testing.
- We'll have to hook this up to a cronjob.

https://2u-internal.atlassian.net/browse/ENT-10533

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/customer_billing/models.py
```diff
@@ -0,0 +1,160 @@
...
+    def is_expired(self):
+        """Check if this reservation has expired."""
+        if self.expires_at:
+            return timezone.now() > self.expires_at
+        return None
+
+    @classmethod
+    def get_reservation_duration(cls):
+        """Get reservation duration from settings or default to 30 minutes."""
+        return getattr(settings, 'ENTERPRISE_SLUG_RESERVATION_MINUTES', 30)
```
pwnage101, line 64: According to stripe's AI assistant: "When a Checkout Session is created, Stripe automatically sets the expires_at timestamp to 24 hours in the future." -- so maybe bump this default from 30 minutes to 24 hours?

enterprise_access/apps/api/v1/views/customer_billing.py
```diff
@@ -164,7 +164,10 @@ def create_checkout_session(self, request, *args, **kwargs):
             f'stripe_price_id="{validated_data["stripe_price_id"]}"'
         )
         try:
-            session = create_free_trial_checkout_session(**serializer.validated_data)
+            session = create_free_trial_checkout_session(
+                request.user,
+                **serializer.validated_data,
+            )
```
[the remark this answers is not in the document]
pwnage101: Oh, I see that you want to pass a User object, not a serialized user_id which then spends a DB call to be converted back to an object. You could still adapt downstream logic to exclude via string lms_user_id ([which is indexed](https://github.com/openedx/enterprise-access/blob/8695239c5d93c93a29b29b9bc847654c2123b3ad/enterprise_access/apps/core/models.py#L41)):

```python
query.exclude(user__lms_user_id=exclude_lms_user_id)
```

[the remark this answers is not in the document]
iloveagent57: yeah, we're now passing around like 3 different pointers to a user identity, which isn't ideal. I like having a FK on the reservation model directly to the user object, so maybe I'll try to bake some flexibiilty into the validation layer (like the caller can provide either an lms_user_id or a User object).

docs/decisions/0029_slug_reservations.rst
```diff
@@ -0,0 +1,114 @@
...
+2. Stripe payment processing and webhook delivery
+3. Salesforce automation triggers
+4. Provisioning API call creates EnterpriseCustomer
+
+This creates opportunities for multiple users to reserve the same enterprise slug,
+leading to downstream provisioning failures in several scenarios:
+
+* Eager administrators double-clicking submit buttons
+* Frontend JavaScript errors causing duplicate API calls  
+* Malicious actors intentionally spamming checkout forms with identical slugs
```
pwnage101, line 28: I'd move "multiple users trying to reserve the same slug" into the list as an additional bullet point.
```suggestion
This creates opportunities for the same enterprise slug to be provisioned twice,
leading to downstream provisioning failures in several scenarios:

* Multiple admins attempting to submit the form using the same slug.
* Eager admins double-clicking submit buttons.
* Frontend JavaScript errors causing duplicate API calls.
* Malicious actors intentionally spamming checkout forms with identical slugs.
```

enterprise_access/apps/customer_billing/constants.py
```diff
@@ -32,3 +43,8 @@
         'DOES_NOT_EXIST': ('does_not_exist', 'This stripe_price_id has not been configured.'),
     },
 }
+
+# According to stripe's AI assistant: "When a Checkout Session is created,
+# Stripe automatically sets the expires_at timestamp to 24 hours in the future."
+# We want the slug duration to last at most as long as the checkout session expiry.
```
pwnage101, line 49: Comment Nit: Technically, I think the reservation needs to last as "least" as long, or else that opens up a window between reservation expiration and checkout session expiration when admins can complete payment after the slug gets re-reserved by somebody else.

pwnage101 (approved):

PR #757, opened 2025-06-23
docs: Add ADR for Self-Service Purchasing BFF Endpoints

pwnage101:
ENT-10532

### DIFF
--- a/docs/decisions/0029-self-service-purchasing-BFF.rst
+++ b/docs/decisions/0029-self-service-purchasing-BFF.rst
@@ -1,0 +1,496 @@
+0029 Self-Service Purchasing BFF Endpoints
+******************************************
+
+Status
+======
+**In progress** (June 2025)
+
+Context
+=======
+
+The self-service purchasing feature will be comprised of primarily the Checkout
+MFE and a set of backend endpoints (some leveraging BFF framework). This ADR
+should help to define those backend endpoints.
+
+The checkout flow is visualized via the following design mockups:
+
+1. `design mockup (old)
+<https://www.figma.com/board/DiZO3HwkQNdElBc5av5DG3/DR--Self-Service-Subscription-Flows?node-id=0-1&p=f&t=VkRFRxPmCAXyZ1bc-0>`_
+
+2. `design mockup (new)
+<https://www.figma.com/board/xWaO5vDxrxSLVB5lDhhEFm/Self-Service-designs?node-id=27-8568&t=XidhCb0byCfFzn8G-0>`_
+
+The above design mockups broadly depict these main pages:
+
+1. Create your account
+
+   * Authentication:
+
+     * JWT Authentication on initial page load.
+
+     * Otherwise, either a login modal or registration modal is triggered on page submit.
+
+   * Fields captured:
+
+     * Full Name
+
+     * Work Email
+
+     * Company Name
+
+     * Country
+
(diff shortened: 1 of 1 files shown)


pwnage101:
Added new commit to address standup review:

* "Company Name" field moved to "Create Your Account" page.
* Several fields repeated, but made inactive on repeated occurrences.
* Renamed `user_exists` response boolean to `user_exists_for_email` to make more explicit the meaning.
* Added a `self_service` boolean to the "context" BFF endpoint to support more advanced frontend logic for a "go to your self-service plan" modal.
* `create-checkout-session` POST API now also returns a `expires_at` timestamp which the frontend can use to support an expiration experience.
* `create-portal-session` POST API has been changed into a `portal-session` GET API, and removed confusing docs describing a JSON response.

pwnage101:
Note: Consider including basic validation clues to context/validation endpoint.

pwnage101:
Another bulk update to this PR in latest commit:

* Removed "stripe" name from keys in API contract (mainly the context endpoint).
* Added a `field_constraints` section to the context endpoint response in order to convey basic constraints which the frontend should be capable of imposing without API calls to the validation endpoint.
* Move any orphaned keys in BFF responses into a top-level dict. Mainly `user_exists_for_email` -> `user_authn.user_exists_for_email`. This keeps with the pattern of treating BFF as an API or domain gateway, where each top level key is the name of the API/domain.
* Various fixes to typos.

docs/decisions/0029-self-service-purchasing-BFF.rst
```diff
@@ -0,0 +1,496 @@
...
+                "error_code": "invalid_format",
+                "developer_message": "Email format validation failed"
+            },
+            "enterprise_slug": {
+                "error_code": "existing_enterprise_customer",
+                "developer_message": "The slug conflicts with an existing customer."
+            },
+            "quantity": {
+                "error_code": "range_exceeded",
+                "developer_message": "Quantity 50 exceeds allowed range [5, 30] for stripe_price_id"
```
brobro10000, line 213: [nit] Idk if there is a way to break this up into 2 seperate error codes (min/max) for instance that way the client can easily parse which error to display back to the user without further parsing.

docs/decisions/0029-self-service-purchasing-BFF.rst
```diff
@@ -0,0 +1,496 @@
...
+.. code-block:: javascript
+
+    // Frontend state management strategy
+    import { create } from 'zustand';
+
+    const useCheckoutFormStore = create<FormStore>(
+      (set) => ({
+        formData: {
+          account: {},  // page 1 state.
+          trial: {},    // page 2 state.
```
brobro10000, line 343: [inform] This is a great layout for state management data-structure but may change based on how the the frontend iterates (mostly around understanding the capabilities and nuances of zustand). This data-structure is the north star during the first iteration.

docs/decisions/0029-self-service-purchasing-BFF.rst
```diff
@@ -0,0 +1,496 @@
...
+        },
+        setFormData: (step, data) => set(
+          (store) => ({
+            formData: {
+              ...store.formData,
+              [step]: data,
```
brobro10000, line 349: [inform] Similar comment to above.

brobro10000 (approved):
Non blocking nits but one suggestion for betteer QoL downstream related to license count error response codes. Great job 🥳

docs/decisions/0029-self-service-purchasing-BFF.rst
pwnage101: I'd advocate we keep it simple because the localized error message for the user can just say "must be within x and y".

PR #758, opened 2025-06-24
feat: add enterprise group deletion event

jesperhodge:
**Description:**

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


jesperhodge:
Here's a video that shows the whole group deletion -> event bus message -> group association deletion cycle working locally on my machine. Don't try to set this up on your machine, it's too time consuming to set up.


https://github.com/user-attachments/assets/85df763b-dd47-4e09-bad4-806825956a2c

PR #763, opened 2025-06-26
chore: safely migrate duplicative fields

brobro10000:
**Description:**
Removes duplicative fields from the context and serialized response from the BFF. The `secured_algolia_api_key` field is now nested under an `algolia` object and cleaned up the context to avoid unused values polluting the context.

Test WIP

**Jira:**
ENT-10496

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/bffs/tests/test_context.py
```diff
@@ -66,7 +64,8 @@ def test_handler_context_init(
                 ],
                 'should_update_active_enterprise_customer_user': False,
             }
-
+        print(context.data)
+        print(expected_data)
```
adamstankiewicz, line 68: nit: remove!

iloveagent57 (approved):

adamstankiewicz (approved):

PR #765, opened 2025-06-30
feat: set LCR status to reversed when related content assignment is reversed

ahmad-hassan-dev:
**Description:**
We're not modifying the existing reversal mechanism for LearnerContentAssignment. We're simply updating the related LearnerCreditRequest to reversed whenever its associated assignment is reversed—ensuring the LCR reflects the correct state without altering the current flow.

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/content_assignments/signals.py
+++ b/enterprise_access/apps/content_assignments/signals.py
@@ -3,7 +3,8 @@ Signal handlers for content_assignments app.
 """
 import logging
 
-from django.db import transaction
+from django.core.exceptions import ValidationError
+from django.db import DatabaseError, IntegrityError
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.utils import timezone
@@ -12,8 +13,17 @@ from openedx_events.enterprise.signals import LEDGER_TRANSACTION_REVERSED
 from enterprise_access.apps.content_assignments.constants import LearnerContentAssignmentStateChoices
 from enterprise_access.apps.content_assignments.models import LearnerContentAssignment
 from enterprise_access.apps.core.models import User
-from enterprise_access.apps.subsidy_request.models import LearnerCreditRequestActions, SubsidyRequestStates
-from enterprise_access.apps.subsidy_request.utils import get_action_choice, get_user_message_choice
+from enterprise_access.apps.subsidy_request.models import (
+    LearnerCreditRequestActionErrorReasons,
+    LearnerCreditRequestActions,
+    SubsidyRequestStates
+)
+from enterprise_access.apps.subsidy_request.utils import (
+    get_action_choice,
+    get_error_reason_choice,
+    get_user_message_choice
+)
+from enterprise_access.utils import format_traceback
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


enterprise_access/apps/content_assignments/signals.py
```diff
@@ -54,13 +57,32 @@ def update_assignment_status_for_reversed_transaction(**kwargs):
...
+                assignment_to_update, "credit_request", None
+            )
+            if learner_credit_request:
+                learner_credit_request.state = SubsidyRequestStates.REVERSED
+                learner_credit_request.save(update_fields=["state"])
+                logger.info(
+                    f"LearnerCreditRequest {learner_credit_request.uuid} reversed due to assignment reversal."
+                )
+                # Add reversal action
+                LearnerCreditRequestActions.create_action(
```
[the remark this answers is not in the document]
sameenfatima78: Also, in case of any error, we should move back the LearnerCreditRequestActions.status back to original state and log error in LearnerCreditRequestActions.traceback

sameenfatima78 (commented):
one comment + make sure to add unit tests for this change

enterprise_access/apps/content_assignments/signals.py
```diff
@@ -52,17 +65,51 @@ def update_assignment_status_for_reversed_transaction(**kwargs):
...
+
+    if learner_credit_request:
+        action_instance = LearnerCreditRequestActions.create_action(
+            learner_credit_request=learner_credit_request,
+            recent_action=get_action_choice(SubsidyRequestStates.REVERSED),
+            status=get_user_message_choice(SubsidyRequestStates.REVERSED),
+        )
+
+    try:
+        with transaction.atomic():
```
zamanafzal, line 86: Why are we using atomic as the transaction is already reversed and here we are only changing states.

ahmad-hassan-dev: We are using `transaction.atomic()` to prevent partial updates—e.g., the assignment is reversed but the linked LearnerCreditRequest (LCR) is not, or vice versa. We don’t want these records to end up in different states, as that could lead to data inconsistencies and unexpected behavior.

This is especially crucial in scenarios where a user will try to create a new request for the same course: if the assignment and LCR are not kept in sync, the system might incorrectly allow or deny new requests, or display inaccurate information to admins and learners. By wrapping these updates in an atomic transaction, we ensure that either both state changes succeed together, or neither does, maintaining data integrity throughout the reversal process.

zamanafzal: The transaction is already reversed here, so updating the record within an atomic block isn't necessary. These are update operations with a very low risk of failure. No? 

Also, the data will not be in good shape if the transaction is reversed in the ledger, but the state is not updated.

enterprise_access/apps/content_assignments/signals.py
```diff
@@ -52,17 +65,51 @@ def update_assignment_status_for_reversed_transaction(**kwargs):
...
+
+    except (ValidationError, IntegrityError, DatabaseError) as exc:
+        error_msg = f"Failed to reverse LearnerContentAssignment {assignment_to_update.uuid}"
+        if learner_credit_request:
+            error_msg += f" and its associated LearnerCreditRequest {learner_credit_request.uuid}"
+        error_msg += f". The entire transaction was rolled back. Error: {exc}"
+        logger.error(error_msg)
+
+        if action_instance:
+            action_instance.status = get_user_message_choice(SubsidyRequestStates.APPROVED)
```
sameenfatima78, line 110: nit: `SubsidyRequestStates.APPROVED` -> `SubsidyRequestStates.ACCEPTED`

enterprise_access/apps/content_assignments/tests/test_signals.py
```diff
@@ -71,3 +84,64 @@ def test_update_assignment_lms_user_id_from_user_email_other_assignments(self):
...
+    def test_reversal_error_updates_action(self):
+        with mock.patch.object(LearnerContentAssignment, "save", side_effect=DatabaseError("Simulated DB error")):
+            update_assignment_status_for_reversed_transaction(
+                ledger_transaction=mock.Mock(uuid=self.assignment.transaction_uuid)
+            )
+        actions = LearnerCreditRequestActions.objects.filter(learner_credit_request=self.lcr)
+        assert actions.exists(), "No LearnerCreditRequestActions created"
+        action = actions.latest('created')
+        assert action.error_reason == LearnerCreditRequestActionErrorReasons.FAILED_REVERSAL
+        assert action.status == SubsidyRequestStates.APPROVED
```
sameenfatima78, line 146: nit: SubsidyRequestStates.APPROVED -> SubsidyRequestStates.ACCEPTED

enterprise_access/apps/content_assignments/signals.py
ahmad-hassan-dev: Nice Catch.

sameenfatima78 (approved):
@ahmad-hassan-dev LGTM. 💯 

But I would advise testing this change on stage before rolling it out on prod to make sure the existing assignments reversal is working as expected.

PR #768, opened 2025-06-30
feat: add SubsidyAccessPolicy model validation to prevent inactive/retired budgets

adamstankiewicz:
**Description:**
Adds model validation to `SubsidyAccessPolicy` to prevent inactive/retired budgets from saving via Django Admin.

image

**Jira:**
[ENT-9126](https://2u-internal.atlassian.net/browse/ENT-9126)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/utils.py
+++ b/enterprise_access/apps/subsidy_access_policy/utils.py
@@ -5,10 +5,18 @@ import hashlib
 
 from django.apps import apps
 from django.conf import settings
+from django.core.exceptions import ValidationError, ObjectDoesNotExist
 from edx_enterprise_subsidy_client import get_enterprise_subsidy_api_client
 from simple_history.models import HistoricalRecords, registered_models
 
 from enterprise_access.apps.subsidy_access_policy import constants
+
+from .constants import (
+    ERROR_MSG_ACTIVE_UNKNOWN_SPEND,
+    ERROR_MSG_ACTIVE_WITH_SPEND,
+    ERROR_MSG_RETIRED_UNKNOWN_SPEND,
+    ERROR_MSG_RETIRED_WITH_SPEND
+)
 
 LEDGERED_SUBSIDY_IDEMPOTENCY_KEY_PREFIX = 'ledger-for-subsidy'
 TRANSACTION_METADATA_KEYS = {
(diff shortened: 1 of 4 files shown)


enterprise_access/apps/subsidy_access_policy/utils.py
```diff
@@ -140,3 +148,56 @@ def cents_to_usd_string(cents):
...
     return "${:,.2f}".format(float(cents) / constants.CENTS_PER_DOLLAR)
+
+
+def validate_retired_budget_deactivation(policy_instance):
+    """
+    Raise a ValidationError if a policy would become both retired and inactive while having existing spend.
+    The error is attached to the field(s) being changed: 'active', 'retired', or both.
+    Uses original values from the database to detect changes.
+    """
+    if policy_instance.pk:
```
iloveagent57, line 159: nit: consider adding a comment clarifying that this conditional is to determine if we're in an update operation vs. a create operation.

enterprise_access/apps/subsidy_access_policy/utils.py
```diff
@@ -140,3 +148,56 @@ def cents_to_usd_string(cents):
...
+
+    active_will_change = policy_instance.active != original_active
+    retired_will_change = policy_instance.retired != original_retired
+
+    will_be_retired_and_inactive = (
+        (original_retired and not policy_instance.active and active_will_change and not retired_will_change) or
+        (policy_instance.retired and not policy_instance.active and retired_will_change)
+    )
+    if not will_be_retired_and_inactive:
+        return
```
iloveagent57, line 179: nit: possibly this will be clearer by adding more variables, something like
```python
would_make_retired_policy_inactive = (
    original_retired and not policy_instance.active and active_will_change and not retired_will_change
)
would_make_inactive_policy_retired = (
    policy_instance.retired and not policy_instance.active and retired_will_change
)
if not (would_make_retired_policy_inactive or would_make_inactive_policy_retired):
    return
```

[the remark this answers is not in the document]
adamstankiewicz: Do you recall if this additional case is one of the relevant use cases for the [ticket](https://2u-internal.atlassian.net/browse/ENT-9126)? For example, my understanding was that the ticket should only involve _retired_ budgets, whereas having `retired=False` doesn't prevent inactivating that budget.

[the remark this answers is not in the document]
pwnage101: huddle summary:
* Since retiring an inactive budget is a no-op, it's okay to allow it to happen.  Furthermore, this gives a pathway for ECS to switch an inactive/unretired budget to an active/retired budget in 2 saves without momentarily exposing the budget to inadvertent spend between the 2 saves.
* Closing the loophole described above could be risky because there could be a scenario where ECS really does want to hide spend.  We could discuss this further during standup.

adamstankiewicz: Updated!

pwnage101 (approved):
LGTM!

PR #792, opened 2025-07-21
feat: Update statuses to persist key instead of value and also adding

zwidekalanga:
support for filtering and sorting on nested fields

**Description:**

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/subsidy_request/migrations/0021_alter_historicallearnercreditrequestactions_error_reason_and_more.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 4.2.21 on 2025-07-22 06:15
```
jajjibhai008, line 1: I am not seeing any relevant models changes in this PR.

jajjibhai008 (commented):
Aren't we folllowing the pattern of sorting and filtering that we are using for assignments?

PR #813, opened 2025-08-11
refactor: move stuff into stripe_api.py

iloveagent57:
**Description:**
Small refactor to prep for future additional functions to fetch stripe data.

**Jira:**
ENT-10714

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/customer_billing/stripe_api.py
```diff
@@ -0,0 +1,59 @@
+"""
+Python API for interacting with Stripe (aside from functions contained in ``pricing_api.py``).
```
iloveagent57, line 2: I think eventually I'll merge these two into just `stripe_api.py`.

PR #826, opened 2025-08-27
chore: create separate celery tasks for decline and cancel email

omar-sarfraz:
**Description:**
Modified the existing email tasks for `cancel` and `decline` to use separate celery tasks instead of shared utility. Also added a base class to handle error action for email tasks.

**Jira:**
[ENT-10373](https://2u-internal.atlassian.net/browse/ENT-10737)

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/content_assignments/tasks.py
```diff
@@ -266,6 +267,10 @@ def get_learner_portal_link(self):
         slug = self.customer_data["slug"]
         return f'{settings.ENTERPRISE_LEARNER_PORTAL_URL}/{slug}'
 
+    def get_enterprise_dashboard_url(self):
```
jajjibhai008, line 270: We have the same functionality in get_learner_portal_link. We can use that

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -13,13 +13,150 @@
...
+from enterprise_access.apps.subsidy_request.models import LearnerCreditRequestActions
+from enterprise_access.apps.subsidy_request.utils import get_action_choice, get_error_reason_choice, get_user_message_choice
 from enterprise_access.tasks import LoggedTaskWithRetry
-from enterprise_access.utils import get_subsidy_model
+from enterprise_access.utils import format_traceback, get_subsidy_model
 
 logger = logging.getLogger(__name__)
 
 
+def _get_learner_credit_request_or_raise(learner_credit_request_uuid):
```
jajjibhai008, line 29: [Curious] Do we really need this utility? we can directly querry this learner_credit_request_model.objects.get(uuid=learner_credit_request_uuid) at anywhere.

jajjibhai008 (approved):

PR #831, opened 2025-09-03
feat: use Stripe price `lookup_key` instead of static price `id`

iloveagent57:
- Switch from Stripe Price IDs to lookup keys for SSP product configuration
- Update validation to check against live Stripe prices instead of static config
- Optimize validation functions to use shared pricing API calls
ENT-10749

This can be tested by hitting the `http://localhost:18270/api/v1/bffs/checkout/context/` endpoint.

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/customer_billing/pricing_api.py
```diff
@@ -187,67 +187,100 @@ def _validate_stripe_price_schema(stripe_price: stripe.Price) -> None:
     logger.debug(f'Stripe price {stripe_price.id} schema validation passed')
 
 
-def get_multiple_stripe_prices(
```
iloveagent57, line 190: Unused function (I think it was LLM fluff).

docs/decisions/0032-stripe-price-lookup-keys.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+Previously, the enterprise-access service used Stripe Price IDs for self-service purchasing (SSP) configuration and validation. Price IDs are generated by Stripe and change between environments (test, production), making configuration management difficult and error-prone.
```
pwnage101, line 12: The key concern is less so that they change between environments, more so that they change every time somebody changes the price or any metadata about the price (e.g. frequency, name, etc.)

PR #832, opened 2025-09-03
feat: [customer_billing] handle invoice.paid Stripe events

pwnage101:
ENT-10752


enterprise_access/apps/customer_billing/stripe_event_types.py
```diff
@@ -0,0 +1,257 @@
+"""
+https://github.com/stripe/stripe-python/blob/40a8d7e0/stripe/_event.py#L150
+"""
+from typing_extensions import Literal
+
+# List must exactly match the one from the stripe library source code.
+StripeEventType = Literal[
```
iloveagent57, line 7: Is there a reason to not import this from stripe?

enterprise_access/apps/customer_billing/stripe_event_handlers.py
```diff
@@ -0,0 +1,100 @@
...
+        checkout_intent_id = int(subscription.metadata['checkout_intent_id'])
+
+        logger.info(
+            f'Found checkout_intent_id="{checkout_intent_id}" '
+            f'stored on the Subscription <subscription_id="{subscription_id}"> '
+            f'related to Invoice <invoice_id="{invoice["id"]}">.'
+        )
+
+        checkout_intent = CheckoutIntent.objects.get(id=checkout_intent_id)
+        logger.info(
```
iloveagent57, line 78: nit: log that we've marked as paid _after_ calling `mark_as_paid()`.

iloveagent57 (approved):

enterprise_access/apps/customer_billing/stripe_event_types.py
pwnage101: The stripe python bindings appear to be auto-generated, and do not internally type hint the "event_type" variable using a type hint "alias", instead opting to inject this long list wherever needed.  Non-DRYness probably doesn't matter for them because their code doesn't need to be DRY if a client generator tool is the only thing writing it.

As a result of the tool not generating a type hint alias, there's none available for me to use.  Yes, technically I could introspect using `stripe.Event.__annontations__['type']`, but all the static python type checkers only do static code evaluation, and do not actually launch a runtime which would be needed to perform dictionary lookups.

pwnage101: But great question, and it should feel bad to do this.

PR #833, opened 2025-09-04
feat: [ForcedPolicyRedemption] Provide external reference ID fallback

pwnage101:
In cases where the external fulfillment already exists (Order UUID already exists for an existing GEAG allocation), allow the ID to be specified in the ForcePolicyRedemption form and annotate the transaction to give enterprise-subsidy redemption logic a fallback value to use.

ENT-10879

Related PR: https://github.com/openedx/enterprise-subsidy/pull/389

UI Screenshot
---

Screenshot 2025-09-04 at 9 17 27 PM

### DIFF
--- a/enterprise_access/apps/subsidy_access_policy/models.py
+++ b/enterprise_access/apps/subsidy_access_policy/models.py
@@ -1063,6 +1063,9 @@ class SubsidyAccessPolicy(TimeStampedModel):
                 requested_price_cents = kwargs.get('requested_price_cents')
                 if requested_price_cents is not None:
                     creation_payload['requested_price_cents'] = requested_price_cents
+                external_fulfillment_reference_id = kwargs.get('external_fulfillment_reference_id')
+                if external_fulfillment_reference_id is not None:
+                    creation_payload['external_fulfillment_reference_id'] = external_fulfillment_reference_id
                 return self.subsidy_client.create_subsidy_transaction(**creation_payload)
             except requests.exceptions.HTTPError as exc:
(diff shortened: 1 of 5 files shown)


enterprise_access/apps/subsidy_access_policy/models.py
```diff
@@ -1063,6 +1063,9 @@ def redeem(self, lms_user_id, content_key, all_transactions, metadata=None, **kw
                 requested_price_cents = kwargs.get('requested_price_cents')
                 if requested_price_cents is not None:
                     creation_payload['requested_price_cents'] = requested_price_cents
+                external_fulfillment_reference_id = kwargs.get('external_fulfillment_reference_id')
+                if external_fulfillment_reference_id is not None:
+                    creation_payload['external_fulfillment_reference_id'] = external_fulfillment_reference_id
```
pwnage101, line 1068: I think this was dead code leftover from a refactor. The approach actually used was to pass the UUID via transaction metadata, not as a net-new create_subsidy_transaction() kwarg. If you follow this kwarg into the [client function code](https://github.com/openedx/edx-enterprise-subsidy-client/blob/8a6579b96abf4e927181fc17b9b1fb8b1aa56005/edx_enterprise_subsidy_client/client.py#L231-L239), it actually doesn't have a matching kwarg, and would throw an "unexpected keyword argument" error if actually passed.

pwnage101: fix -> https://github.com/openedx/enterprise-access/pull/834

PR #837, opened 2025-09-08
feat: tighten up CheckoutIntent state transitions

pwnage101:
* model-level state transitions actually respect the allowed state transitions defined in constants.py.
* CheckoutIntent state transitions now always log when they happen.

### DIFF
--- a/enterprise_access/apps/api/v1/views/customer_billing.py
+++ b/enterprise_access/apps/api/v1/views/customer_billing.py
@@ -24,7 +24,6 @@ from enterprise_access.apps.customer_billing.api import (
     CreateCheckoutSessionValidationError,
     create_free_trial_checkout_session
 )
-from enterprise_access.apps.customer_billing.constants import ALLOWED_CHECKOUT_INTENT_STATE_TRANSITIONS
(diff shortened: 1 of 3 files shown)


enterprise_access/apps/api/v1/views/customer_billing.py
```diff
@@ -338,8 +337,8 @@ def partial_update(self, request, *args, **kwargs):
 
         # Check if state is being updated
         new_state = request.data.get('state')
-        if new_state and new_state != instance.state:
-            if not self._is_valid_state_transition(instance.state, new_state):
+        if new_state:
+            if not CheckoutIntent.is_valid_state_transition(instance.state, new_state):
```
brobro10000, line 341: [nit] Could this be a single conditional with an `and` operator? As long as new_state always gets checked first, it should always short circuit the logic before passing the value against the `CheckoutIntent`'s, `is_valid_state_transition` method.

brobro10000 (approved):
1 minor nit but looks good. 👍🏽

iloveagent57 (approved):
nice

PR #850, opened 2025-09-22
build: split up CI workflows

iloveagent57:
Try splitting lint into a discrete workflow.

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


pwnage101 (approved):
Nit: Why bucket isort_check, pii_check, and check_keywords under style instead of lint? I'd colloquially call those linters too.

PR #854, opened 2025-09-25
feat: bulk remind for lcr

ahmad-hassan-dev:
**Description:**

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance

### DIFF
--- a/enterprise_access/apps/api/v1/views/browse_and_request.py
+++ b/enterprise_access/apps/api/v1/views/browse_and_request.py
@@ -49,6 +49,7 @@ from enterprise_access.apps.core import constants
 from enterprise_access.apps.subsidy_access_policy.api import approve_learner_credit_request_via_policy
 from enterprise_access.apps.subsidy_access_policy.exceptions import SubisidyAccessPolicyRequestApprovalError
 from enterprise_access.apps.subsidy_access_policy.models import SubsidyAccessPolicy
+from enterprise_access.apps.subsidy_request import api as subsidy_request_api
 from enterprise_access.apps.subsidy_request.constants import (
     REUSABLE_REQUEST_STATES,
     LearnerCreditAdditionalActionStates,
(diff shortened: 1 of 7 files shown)


enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -1093,27 +1094,74 @@ def cancel(self, request, *args, **kwargs):
...
         """
         serializer = serializers.LearnerCreditRequestRemindSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
-        learner_credit_request = serializer.get_learner_credit_request()
-        assignment = learner_credit_request.assignment
-
-        action_instance = LearnerCreditRequestActions.create_action(
-            learner_credit_request=learner_credit_request,
-            recent_action=get_action_choice(LearnerCreditAdditionalActionStates.REMINDED),
-            status=get_user_message_choice(LearnerCreditAdditionalActionStates.REMINDED),
```
sameenfatima78, line 1106: any reason why we are moving the action in task.py?

enterprise_access/apps/api/v1/views/browse_and_request.py
```diff
@@ -1093,27 +1094,74 @@ def cancel(self, request, *args, **kwargs):
...
+        allowing admins to send bulk reminders to a specific subset of requests.
+
+        ```
+        Raises:
+            404 if no cancelable assignments were found
+            422 if any of the assignments threw an error (not found or not remindable)
+        ```
+        """
+        # A request is only remindable if it is in the 'APPROVED' state.
+        base_queryset = self.get_queryset().filter(
```
sameenfatima78, line 1144: does get_queryset() return only the requests belonging to a specific budget?

enterprise_access/apps/subsidy_request/tasks.py
```diff
@@ -396,6 +399,16 @@ def send_reminder_email_for_pending_learner_credit_request(assignment_uuid):
...
+
+    if hasattr(assignment, 'credit_request') and assignment.credit_request:
+        LearnerCreditRequestActions.create_action(
+            learner_credit_request=assignment.credit_request,
+            recent_action=get_action_choice(LearnerCreditAdditionalActionStates.REMINDED),
+            status=get_user_message_choice(LearnerCreditAdditionalActionStates.REMINDED),
+        )
+        logger.info(
+            f'Created REMINDED action for LearnerCreditRequest {assignment.credit_request.uuid}'
+        )
```
sameenfatima78, line 411: are we logging an action in case of failure with correct state?

enterprise_access/apps/api/v1/views/browse_and_request.py
ahmad-hassan-dev: The most critical reason is to ensure we only record an action after the work has been successfully completed. By creating the `LearnerCreditRequestActions` record inside the Celery task, we guarantee that a "reminded" action is only logged if the email to the user was actually sent. If we created the action in the view and the task later failed, our system's history would be incorrect.

enterprise_access/apps/subsidy_request/api.py
```diff
@@ -0,0 +1,45 @@
...
+        if req.state == SubsidyRequestStates.APPROVED and req.assignment_id is not None
+    }
+
+    non_remindable_requests = set(requests) - remindable_requests
+
+    logger.info(f'Skipping {len(non_remindable_requests)} non-remindable learner credit requests.')
+    logger.info(f'Queueing reminders for {len(remindable_requests)} learner credit requests.')
+
+    for req in remindable_requests:
+        send_reminder_email_for_pending_learner_credit_request.delay(req.assignment.uuid)
```
sameenfatima78: this statement seems like it's not indented correctly, kindly check

PR #860, opened 2025-10-01
feat: update create billing portal session api

brobro10000:
**Description:**
This pull request simplifies and refactors the `create_portal_session` API endpoint in `customer_billing.py` by removing dependencies on the LMS API client and fetching customer data directly from the local `CheckoutIntent` model. It also improves error handling and clarifies authentication requirements.

In support of this POC from the checkout MFE: https://github.com/edx/frontend-app-enterprise-checkout/pull/58

Refactoring and Dependency Removal:

* Removed the use of `LmsApiClient` and related external API calls; customer and enterprise information is now retrieved directly from the `CheckoutIntent` model. [[1]](diffhunk://#diff-2b20872f01b113277b7d353238b6c5c440b33704946f720c0a36a393c750639cL20) [[2]](diffhunk://#diff-2b20872f01b113277b7d353238b6c5c440b33704946f720c0a36a393c750639cL191-R234)
* Eliminated the import of the unused `requests` library.

API Endpoint Changes:

* Temporarily removed the role-based permission decorator and updated the function signature, with a note to revisit authentication requirements.

Error Handling Improvements:

* Added explicit error checks and logging for missing or invalid `checkout_intent_id`, missing `CheckoutIntent` records, and missing required fields.
* Improved Stripe API error handling with specific logging and appropriate HTTP status codes for different failure scenarios.

**Jira:**
ENT-XXXX

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


enterprise_access/apps/api/v1/views/customer_billing.py
```diff
@@ -181,37 +199,135 @@ def create_checkout_session(self, request, *args, **kwargs):
...
+        if not checkout_intent:
+            logger.error(f"No checkout intent for id, for requesting user {request.user.id}")
+            return Response(customer_portal_session, status=status.HTTP_404_NOT_FOUND)
+
+        stripe_customer_id = checkout_intent.stripe_customer_id
+        if not stripe_customer_id:
+            logger.error(f"No stripe customer id associated to CheckoutIntent {checkout_intent}")
+            return Response(customer_portal_session, status=status.HTTP_404_NOT_FOUND)
+
+        if not checkout_intent:
```
pwnage101, line 296: There a lot of code that is common between the two new endpoints.  I'd recommend making this DRY and consolidating as much common code into a single function, possibly an internal python API function stored in the customer_billing domain (e.g. `enterprise_access/apps/customer_billing/api.py`)

enterprise_access/apps/api/v1/views/customer_billing.py
```diff
@@ -181,37 +199,136 @@ def create_checkout_session(self, request, *args, **kwargs):
...
+        detail=False,
         methods=['get'],
-        url_path='create-portal-session',
+        url_path='create-enterprise-admin-portal-session',
+    )
+    # # UUID in path is used as the "permission object" for role-based auth.
+    @permission_required(
+        CUSTOMER_BILLING_CREATE_PORTAL_SESSION_PERMISSION,
+        fn=lambda request, **kwargs: request.GET.get('enterprise_customer_uuid') or kwargs.get(
+            'enterprise_customer_uuid')
```
brobro10000, line 213: @iloveagent57 , @pwnage101  - Sanity check: The fn in `@permission_required` now pulls `enterprise_customer_uuid` from `request.GET` (query param) instead of only `kwargs`, so RBAC actually evaluates the right enterprise UUID instead of `None` (which caused blanket 403s before).

iloveagent57: ahh, right, that makes sense.

iloveagent57:
Nice job, I think we can improve this in the future to make it align with standard DRF patterns, namely moving the `create_checkout_portal_session` action into the `CheckoutIntent` viewset. But this will definitely work for now.

PR #862, opened 2025-10-02
feat: change create_intent() to allow blank slug/name

pwnage101:
This change is needed to support the use case of creating an intent WITHOUT an enterprise slug/name, such as during submission of the PlanDetails page (first page) of the checkout stepper.

ENT-10756

Pairs with: https://github.com/edx/frontend-app-enterprise-checkout/pull/60


enterprise_access/apps/customer_billing/models.py
```diff
@@ -425,28 +429,63 @@ def create_intent(
...
+            should_reserve_new_slug = False
+            if slug:
+                if existing_intent:
+                    slug_changing = slug != existing_intent.enterprise_slug
+                    name_changing = name != existing_intent.enterprise_name
+                    should_reserve_new_slug = slug_changing or name_changing  # Cases 5, 6, 7
+                else:
+                    should_reserve_new_slug = True  # Case 4
+            else:
+                pass  # Cases 1, 2, 3
```
iloveagent57, line 466: technicall we don't need this `else`.  maybe a comment next to `should_reserve_new_slug = False` like "default to False, covers cases # 1, 2, 3

enterprise_access/apps/customer_billing/models.py
```diff
@@ -425,28 +429,63 @@ def create_intent(
...
                 # Update the existing CREATED or EXPIRED intent
                 existing_intent.state = CheckoutIntentState.CREATED
-                existing_intent.enterprise_slug = slug
-                existing_intent.enterprise_name = name
                 existing_intent.quantity = quantity
                 existing_intent.expires_at = expires_at
+
+                # Any of the following could be None since they're optional, in
+                # which case they would be unset, making this method act more
+                # like PUT than PATCH.
```
iloveagent57, line 483: We should articulate this in the method docstring, too

iloveagent57 (approved):
I think this makes sense.

PR #868, opened 2025-10-06
chore: Updating Python Requirements

iloveagent57:
Re-opening in a different PR to get CI action to run. Runs fine locally.

**Merge checklist:**

**Post merge:**
- [ ] Deploy to prod instance


brobro10000 (approved):
Approved to unblock

Do we know if this had any effect on the checkout experience beyond just the error methods being referenced directly?
Stripe customer/Stripe checkout session etc.

